PR #14, opened 2016-04-04
SLCORE-33 Mark top level modules

dbmeneses:


client-api/src/main/java/org/sonarsource/sonarlint/core/client/api/connected/RemoteModule.java
```diff
@@ -25,4 +25,6 @@
 
   String getName();
 
+  boolean isTopModule();
```
henryju: We usually name this "isRoot()" in SonarQube world so maybe we should do the same in SLCORE.

core/src/test/java/org/sonarsource/sonarlint/core/container/connected/update/ModuleListDownloaderTest.java
```diff
@@ -50,5 +50,6 @@ public void update_modules() throws Exception {
 
     ModuleList moduleList = ProtobufUtil.readFile(tempDir.toPath().resolve(StorageManager.MODULE_LIST_PB), ModuleList.parser());
     assertThat(moduleList.getModulesByKey()).hasSize(1559);
+    assertThat(moduleList.getModulesByKey().values()).extracting("qu").contains("TRK");
```
henryju: Should also contains BRC no? If everything is TRK then the feature is useless.

henryju:
Few comments then LGTM

PR #15, opened 2016-04-04
Add logging context

dbmeneses:
To avoid warnings like this:  LOGBACK: No context given for ch.qos.logback.core.ConsoleAppender


henryju:
LGTM

PR #18, opened 2016-04-04
SLCORE-34 Generate authentication token

dbmeneses:


client-api/src/main/java/org/sonarsource/sonarlint/core/client/api/connected/WsHelper.java
```diff
@@ -0,0 +1,35 @@
...
+  /**
+   * Check if it is possible to reach server with provided configuration
+   * @since 2.1
+   */
+  ValidationResult validateConnection(ServerConfiguration serverConfig);
+
+  /**
+   * Create authentication token with provided server configuration
+   * @since 2.1
+   * @param force Whether it should revoke any existing token with the same name, or fail if it exists.
```
henryju: You can maybe document the exception thrown in case token already exists

core/src/main/java/org/sonarsource/sonarlint/core/container/connected/SonarLintWsClient.java
```diff
@@ -68,6 +69,26 @@ public CloseableWsResponse get(String path) {
     return response;
   }
 
+  public CloseableWsResponse post(String path) {
+    CloseableWsResponse response = rawPost(path);
+    if (!response.isSuccessful()) {
+      throw handleError(response);
+    }
+    return new CloseableWsResponse(response);
```
henryju: Double wrap

henryju: Double wrap

henryju:
2 comments then LGTM

dbmeneses:
feedback applied

PR #19, opened 2016-04-05
SLCORE-34 Generate authentication token

dbmeneses:


henryju:
It would probably deserve an IT (check that you can generate a token in 5.3 and that it properly fail in < 5.3). Appart from that LGTM.

dbmeneses:
IT added

henryju:
LGTM

PR #31, opened 2016-04-28
SLCORE-23 Add support for Python

henryju:


core/src/main/java/org/sonarsource/sonarlint/core/container/standalone/rule/StandaloneRuleRepositoryContainer.java
```diff
@@ -50,6 +51,7 @@ private void addCoreComponents() {
     add(StandalonePluginRulesLoader.class,
       new StandaloneRulesProvider(),
       RuleFinderCompatibility.class,
+      RulesDefinitionXmlLoader.class,
```
dbmeneses: is this used?

dbmeneses:
LGTM

core/src/main/java/org/sonarsource/sonarlint/core/container/standalone/rule/StandaloneRuleRepositoryContainer.java
henryju: Yep, by Python plugin...

PR #32, opened 2016-05-17
SLCORE-48 Rules created from templates don't raise issues

dbmeneses:
Tested in SLI


henryju:
LGTM if you have time maybe try to add an IT or a medium test

PR #36, opened 2016-05-19
Fix logging

dbmeneses:


core/src/main/java/org/sonarsource/sonarlint/core/container/storage/SonarQubeActiveRulesProvider.java
```diff
@@ -73,6 +73,11 @@ public ActiveRules provide(Sonarlint.Rules storageRules, Sonarlint.QProfiles qPr
 
         org.sonarsource.sonarlint.core.proto.Sonarlint.ActiveRules activeRulesFromStorage = ProtobufUtil.readFile(storageManager.getActiveRulesPath(qProfileKey),
           Sonarlint.ActiveRules.parser());
+
+        if (!supportQualityProfilesWS) {
+          LOG.debug("  * " + language + ": " + qProfileKey + " (" + activeRulesFromStorage.getActiveRulesByKey().size() + " rules)");
```
henryju: Looks good but in this case do we need to keep the code at line 65? Can't we always use this one?

dbmeneses: For the case when activeRuleCount == 0

dbmeneses: we can do it just before the continue, though

PR #38, opened 2016-07-11
SLCORE-55 Report unparseable files

dbmeneses:


core/src/main/java/org/sonarsource/sonarlint/core/analyzer/sensor/LtsApiSensorContext.java
```diff
@@ -236,23 +210,9 @@ public Measure saveMeasure(InputFile inputFile, Measure measure) {
...
-        public String getPath() {
-          return inputPath.absolutePath();
-        }
-
-        @Override
-        public org.sonar.api.resources.Directory getParent() {
-          return Directory.create(PathUtils.sanitize(inputPath.path().getParent().toString()));
-        }
-      };
+      File f = File.create(inputPath.absolutePath());
```
henryju: This what make oldest versions of the Java plugin incompatible. I will modify the constructor to be protected so that we can override the class again.

pom.xml
```diff
@@ -32,7 +32,7 @@
   </modules>
 
   <properties>
-    <sonarqube.version>5.6-RC1</sonarqube.version>
+    <sonarqube.version>6.0-SNAPSHOT</sonarqube.version>
```
henryju: Would be better to use a QAed version so that it will build on travis.

PR #40, opened 2016-07-12
Optimize access to failed analysis files

dbmeneses:
I'm gonna assume that ClientInputFile either implements property hashcode/equals, or  there is a single instance per file (with different path). The goal is to have quick .contains() tests.

### DIFF
--- a/core/src/main/java/org/sonarsource/sonarlint/core/container/analysis/DefaultAnalysisResult.java
+++ b/core/src/main/java/org/sonarsource/sonarlint/core/container/analysis/DefaultAnalysisResult.java
@@ -19,16 +19,15 @@
  */
 package org.sonarsource.sonarlint.core.container.analysis;
 
-import java.nio.file.Path;
 import java.util.Collection;
-import java.util.HashMap;
-import java.util.Map;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/sonarsource/sonarlint/core/container/analysis/DefaultAnalysisResult.java
```diff
@@ -19,16 +19,15 @@
...
-import java.util.Map;
+import java.util.HashSet;
+import java.util.Set;
 
 import org.sonarsource.sonarlint.core.client.api.common.analysis.AnalysisResults;
 import org.sonarsource.sonarlint.core.client.api.common.analysis.ClientInputFile;
 
 public class DefaultAnalysisResult implements AnalysisResults {
-  private Map<Path, ClientInputFile> failedAnalysisFiles = new HashMap<>();
+  private Set<ClientInputFile> failedAnalysisFiles = new HashSet<>();
```
henryju: LinkedHashSet usually help in testing to have consistent result.

henryju:
LGTM

PR #52, opened 2016-10-13
SLCORE-68 Download module hierarchy

dbmeneses:


core/src/test/java/org/sonarsource/sonarlint/core/container/connected/IssueStoreFactoryTest.java
```diff
@@ -0,0 +1,40 @@
...
+public class IssueStoreFactoryTest {
+  @Rule
+  public TemporaryFolder temp = new TemporaryFolder();
+
+  @Test
+  public void testFactory() {
+    IssueStoreFactory factory = new IssueStoreFactory();
+    IssueStore store = factory.apply(temp.getRoot().toPath());
+
+    assertThat(store).isNotNull();
```
janos-ss: The `null` check is redundant if followed by an instance of check

PR #55, opened 2016-10-17
Move objectstore to API

dbmeneses:


core/src/main/java/org/sonarsource/sonarlint/core/container/connected/objectstore/SimpleObjectStore.java
```diff
@@ -25,6 +25,11 @@
 import java.nio.file.Path;
 import java.util.Optional;
 
+import org.sonarsource.sonarlint.core.client.api.connected.objectstore.ObjectStore;
+import org.sonarsource.sonarlint.core.client.api.connected.objectstore.PathMapper;
+import org.sonarsource.sonarlint.core.client.api.connected.objectstore.Reader;
+import org.sonarsource.sonarlint.core.client.api.connected.objectstore.Writer;
```
janos-ss: Unused imports?

dbmeneses: SimpleObjectStore is in the implementation while the interfaces are in the API

janos-ss: Oh I see, they are not in the same package anymore.

PR #65, opened 2016-10-27
SLCORE-71 Support reading files from VFS

dbmeneses:


core/src/test/java/org/sonarsource/sonarlint/core/TestUtils.java
```diff
@@ -45,5 +53,45 @@ public void log(String formattedMessage, Level level) {
   public static NoOpLogOutput createNoOpLogOutput() {
     return new NoOpLogOutput();
   }
+  
+  public static ClientInputFile createInputFile(final Path path, final boolean isTest) {
+    return createInputFile(path, isTest, StandardCharsets.UTF_8);
+  }
+  
+  public static ClientInputFile createInputFile(final Path path, final boolean isTest, final Charset encoding) {
```
janos-ss: Cannot use the `TestClientInputFile` here?

PR #72, opened 2016-11-17
SLCORE-78 avoid "Bad component key" log

dbmeneses:


dbmeneses:
@henryju I updated the P/R. Thanks for the feedback

janos-ss:
This change was merged.

PR #76, opened 2016-11-23
SLI-130 fix module key of nested projects in windows

janos-ss:
Also fix all unit test failures in Windows.


client-api/src/main/java/org/sonarsource/sonarlint/core/client/api/util/FileUtils.java
```diff
@@ -75,4 +79,16 @@ public static void forceMkDirs(Path path) {
     }
   }
 
+  /**
+   * Convert path to SonarQube canonical path
+   *
+   * @param path path string in the local OS
+   * @return SonarQube canonical path
+   */
+  public static String toCanonicalPath(String path) {
```
dbmeneses: Maybe rename to not confuse with File#getCanonicalPath

PR #95, opened 2017-01-17
SLCORE-98 Fix minimum supported plugin versions and ITs

janos-ss:
This is a partial fix of the ticket.

- Make ITs for LTS run with "sort of" minimum supported versions instead of latest
- Update min supported version of Cobol to 3.2


cix.sh
```diff
@@ -5,13 +5,13 @@ set -euo pipefail
...
 then
-  JAVA_VERSION=3.14
-  PHP_VERSION=2.8
-  JAVASCRIPT_VERSION=2.13
+  JAVA_VERSION=3.12
+  PHP_VERSION=2.7
+  JAVASCRIPT_VERSION=2.11
   PYTHON_VERSION=1.5
-  COBOL_VERSION=3.1.1
+  COBOL_VERSION=3.2
```
dbmeneses: maybe do if .. / else if ... / else boom ?

PR #97, opened 2017-01-24
SLCORE-94 Use pagination with WS api/components/tree

henryju:


core/src/main/java/org/sonarsource/sonarlint/core/container/connected/update/ModuleHierarchyDownloader.java
```diff
@@ -53,29 +55,37 @@ public ModuleHierarchyDownloader(SonarLintWsClient wsClient) {
...
 
-      // component -> ancestorComponent. Doesn't include root
-      Map<Component, Component> ancestors = new HashMap<>();
-      for (Component c : treeResponse.getComponentsList()) {
-        ancestors.put(c, components.get(fetchAncestorId(c.getId())));
+        modules.addAll(treeResponse.getComponentsList());
+      } catch (IOException e) {
+        throw new IllegalStateException("Failed to load module hierarchy", e);
       }
+    } while (page * 500 < treeResponse.getPaging().getTotal());
```
janos-ss: Magic number 500 is also embedded in the API query string, which might get overlooked when modifying this code in the future

core/src/test/java/org/sonarsource/sonarlint/core/WsClientTestUtils.java
```diff
@@ -71,6 +73,13 @@ public static SonarLintWsClient addResponse(SonarLintWsClient wsClient, String u
     return wsClient;
   }
 
+  public static SonarLintWsClient addResponse(SonarLintWsClient wsClient, String url, Message m) throws IOException {
+    ByteArrayOutputStream bos = new ByteArrayOutputStream();
+    m.writeTo(bos);
+    bos.close();
```
janos-ss: Why not wrap in try-with-resources?

core/src/test/java/org/sonarsource/sonarlint/core/container/connected/update/ModuleHierarchyDownloaderTest.java
```diff
@@ -69,9 +72,59 @@ public void test() {
   }
 
   @Test
+  public void testNoPagination() throws IOException {
+    TreeWsResponse.Builder responseBuilder = TreeWsResponse.newBuilder()
+      .setPaging(Paging.newBuilder().setTotal(500));
+    for (int i = 0; i < 500; i++) {
```
janos-ss: Much less than 500 iterations could equally prove a point (in `testPagination` too, further below)

core/src/test/java/org/sonarsource/sonarlint/core/container/connected/update/ModuleHierarchyDownloaderTest.java
```diff
@@ -69,9 +72,59 @@ public void test() {
...
+        .setKey("testRoot" + i));
+      WsClientTestUtils.addStreamResponse(wsClient, "api/components/show.protobuf?id=id" + i,
+        "/update/show_module1.pb");
+    }
+
+    WsClientTestUtils
+      .addResponse(wsClient, "api/components/tree.protobuf?ps=500&p=1&qualifiers=BRC&baseComponentKey=testRoot", responseBuilder.build());
+
+    Map<String, String> fetchModuleHierarchy = downloader.fetchModuleHierarchy("testRoot");
+    assertThat(fetchModuleHierarchy).hasSize(500 + 1);
```
janos-ss: What is the +1 ?

core/src/test/java/org/sonarsource/sonarlint/core/WsClientTestUtils.java
henryju: Indeed

core/src/main/java/org/sonarsource/sonarlint/core/container/connected/update/ModuleHierarchyDownloader.java
henryju: Yep, I'll extract a constant

core/src/test/java/org/sonarsource/sonarlint/core/container/connected/update/ModuleHierarchyDownloaderTest.java
henryju: Here I want to test that no extra/useless pagination is made when result count is exactly the page number. It would be unnoticed at runtime since doing a WS call with a page index too big simply return an empty result list.

PR #117, opened 2017-04-27
SLCORE-110 Allow to validate organization

henryju:


core/src/main/java/org/sonarsource/sonarlint/core/container/storage/StorageManager.java
```diff
@@ -74,11 +77,18 @@ public Path getModuleStorageRoot(String moduleKey) {
...
     try {
-      return URLEncoder.encode(name, StandardCharsets.UTF_8.name());
+      encoded = URLEncoder.encode(name, StandardCharsets.UTF_8.name());
     } catch (UnsupportedEncodingException e) {
       throw new IllegalStateException("Unable to encode name: " + name, e);
     }
+    if (encoded.length() > MAX_FOLDER_NAME_SIZE) {
+      // Most FS will not support a folder name greater than 255
+      String md5 = DigestUtils.md5Hex(encoded);
+      return encoded.substring(0, 255 - md5.length()) + md5;
```
dbmeneses: s/255/MAX_FOLDER_NAME_SIZE

dbmeneses: I'm also wondering if we should not take the md5 sum of the un-encoded name

PR #138, opened 2017-07-04
SLCORE-134 Support for Atom in LS

janos-ss:


language-server/src/main/java/org/sonarlint/languageserver/SonarLintLanguageServer.java
```diff
@@ -336,8 +336,11 @@ public void didClose(DidCloseTextDocumentParams params) {
 
   @Override
   public void didSave(DidSaveTextDocumentParams params) {
```
henryju: What about other events? didOpen, didChange? Can it receive null?

janos-ss: I haven't seen that happen yet.

PR #145, opened 2017-11-02
Add SonarLintUtils with isInternalDebugEnabled helper

janos-ss:


core/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryClient.java
```diff
@@ -53,15 +59,19 @@ void upload(TelemetryData data) {
     try {
       sendPost(httpFactory.buildClient(clientConfig), createPayload(data));
     } catch (Exception e) {
-      // fail silently
+      if (SonarLintUtils.isInternalDebugEnabled()) {
+        LOG.debug("Failed to upload telemetry data", e);
```
dbmeneses: Maybe it should be logged as an error? If the flag is enabled, means that whoever is usign it wants to see the errors. Debug might not appear at all if the SonarLint console doesn't have the debug flag activated.

core/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryClient.java
```diff
@@ -21,15 +21,21 @@
...
+import org.sonarsource.sonarlint.core.util.SonarLintUtils;
 import org.sonarsource.sonarlint.core.util.ws.DeleteRequest;
 import org.sonarsource.sonarlint.core.util.ws.HttpConnector;
 import org.sonarsource.sonarlint.core.util.ws.PostRequest;
 
 import static java.time.temporal.ChronoUnit.DAYS;
 
 public class TelemetryClient {
+
+  private static final Logger LOG = LoggerFactory.getLogger(TelemetryClient.class);
```
dbmeneses: It's better to use org.sonar.api.utils.log.Loggers.get

core/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryPathManager.java
```diff
@@ -22,10 +22,15 @@
...
 import java.nio.file.Files;
 import java.nio.file.Path;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.sonarsource.sonarlint.core.client.api.common.SonarLintPathManager;
+import org.sonarsource.sonarlint.core.util.SonarLintUtils;
 
 public class TelemetryPathManager {
 
+  private static final Logger LOG = LoggerFactory.getLogger(TelemetryPathManager.class);
```
dbmeneses: it's better to use sonar api Loggers

core/src/main/java/org/sonarsource/sonarlint/core/util/SonarLintUtils.java
```diff
@@ -0,0 +1,34 @@
...
+
+  static final String INTERNAL_DEBUG_ENV = "SONARLINT_INTERNAL_DEBUG";
+
+  private SonarLintUtils() {
+    // utility class, forbidden constructor
+  }
+
+  public static boolean isInternalDebugEnabled() {
+    String value = System.getenv(INTERNAL_DEBUG_ENV);
+    return value != null && !value.isEmpty();
```
dbmeneses: what about something like "true".equalsIgnoreCase(System.getEnv(..))

core/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryClient.java
janos-ss: One reason we silently ignored these errors was obviously to avoid any kind of noise from telemetry for users. Imagine that reason doesn't exist. Would we log such errors at any level other than debug? I'm not sure.

But anyway, I will change all these to `error` level now, to make sure we see them if they ever happen. These should probably never happen, and when they do, we want to know.

PR #148, opened 2017-11-07
Fix ITs

dbmeneses:


run-integration-tests.sh
```diff
@@ -0,0 +1,63 @@
...
+
+if [ -z "$1" ]; then
+  echo "syntax: $0 [DEV|LTS|LATEST_RELEASE]"
+  exit 1
+fi
+
+set -ex
+SQ_VERSION=$1
+
+cd its
```
janos-ss: Instead of this, probably you could add `-f its/pom.xml` to the `mvn` command. If this is needed, then move down in the script, right before it's actually needed. (It would be good to get rid of it, because `cd` inside scripts is not recommended in general.)

cix.sh
```diff
@@ -1,56 +1,6 @@
 #!/bin/bash
 set -euo pipefail
 
-plugins_min_versions_path=core/src/main/resources/plugins_min_versions.txt
```
janos-ss: I added a commit to your branch, this variable is needed by the new script where you extracted the code.

dbmeneses: thx

PR #168, opened 2018-01-09
Report telemetry errors if debug flag enabled

dbmeneses:


core/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryStorage.java
```diff
@@ -70,8 +77,15 @@ private TelemetryData load() throws IOException {
 
   TelemetryData tryLoad() {
     try {
+      if (!Files.isRegularFile(path)) {
```
janos-ss: This could be outside the try-catch

PR #181, opened 2018-02-21
SLCORE-173 Don't request more than 20 pages in paginated WS

dbmeneses:


core/src/main/java/org/sonarsource/sonarlint/core/container/connected/SonarLintWsClient.java
```diff
@@ -193,7 +194,7 @@ public String getOrganizationKey() {
         boolean isEmpty = items.isEmpty();
         Paging paging = getPaging.apply(protoBufResponse);
         // SONAR-9150 Some WS used to miss the paging information, so iterate until response is empty
-        stop = isEmpty || (paging.getTotal() > 0 && page * PAGE_SIZE >= paging.getTotal());
+        stop = isEmpty || page >= MAX_PAGES || (paging.getTotal() > 0 && page * PAGE_SIZE >= paging.getTotal());
```
henryju: Maybe log a warning in this case, because it may lead to very unexpected behavior (I'm thinking of rules, component paths, ...)

dbmeneses:
@henryju p/r updated

PR #185, opened 2018-03-19
SLCORE-177 Plugin classloader doesn't fallback to system's classloader

dbmeneses:


dbmeneses:
See details here: https://jira.sonarsource.com/browse/SLCORE-177

dbmeneses:
This doesn't work as expected: it seems that ClassRealm is trying first to use the default classloader before trying the plugins classpath.

PR #187, opened 2018-04-05
SLCORE-180 Add excludedRules to StandaloneAnalysisConfiguration

janos-ss:


client-api/pom.xml
```diff
@@ -16,6 +16,26 @@
       <version>3.0.2</version>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>org.sonarsource.sonarqube</groupId>
+      <artifactId>sonar-plugin-api</artifactId>
```
henryju: I think that adding a dependency on sonar-plugin-api just for the RuleKey class is overkill.

core/src/main/java/org/sonarsource/sonarlint/core/container/standalone/StandaloneGlobalContainer.java
```diff
@@ -170,4 +174,48 @@ public RuleDetails getRuleDetails(String ruleKeyStr) {
...
+    public ActiveRule find(RuleKey ruleKey) {
+      if (exclusions.contains(ruleKey)) {
+        return null;
+      }
+      return activeRules.find(ruleKey);
+    }
+
+    @Override
+    public Collection<ActiveRule> findAll() {
+      return activeRules.findAll();
```
henryju: Should exclude disabled rules.

core/src/main/java/org/sonarsource/sonarlint/core/container/standalone/StandaloneGlobalContainer.java
```diff
@@ -170,4 +174,48 @@ public RuleDetails getRuleDetails(String ruleKeyStr) {
...
+    }
+
+    @Override
+    public Collection<ActiveRule> findByRepository(String s) {
+      return activeRules.findByRepository(s);
```
henryju: Should exclude disabled rules.

core/src/main/java/org/sonarsource/sonarlint/core/container/standalone/StandaloneGlobalContainer.java
```diff
@@ -170,4 +174,48 @@ public RuleDetails getRuleDetails(String ruleKeyStr) {
...
+    }
+
+    @Override
+    public Collection<ActiveRule> findByLanguage(String s) {
+      return activeRules.findByLanguage(s);
```
henryju: Should exclude disabled rules.

core/src/main/java/org/sonarsource/sonarlint/core/container/standalone/StandaloneGlobalContainer.java
```diff
@@ -170,4 +174,48 @@ public RuleDetails getRuleDetails(String ruleKeyStr) {
...
+    }
+
+    @CheckForNull
+    @Override
+    public ActiveRule findByInternalKey(String s, String s1) {
+      return activeRules.findByInternalKey(s, s1);
```
henryju: Should exclude disabled rules.

core/src/main/java/org/sonarsource/sonarlint/core/container/standalone/StandaloneGlobalContainer.java
```diff
@@ -138,6 +141,7 @@ public AnalysisResults analyze(StandaloneAnalysisConfiguration configuration, Is
     analysisContainer.add(issueListener);
     analysisContainer.add(rules);
     analysisContainer.add(activeRules);
+    activeRules.setExclusions(configuration.excludedRules());
```
henryju: Is it not dangerous in case of concurrent analysis to have a mutable ActiveRulesWithExclusions? Why not simply replace the previous line by: `analysisContainer.add(new ActiveRulesWithExclusions(container.getActiveRules(), configuration.excludedRules()))`; and revert the change at line 134

henryju:
You should create a ticket in SLCORE JIRA project

PR #196, opened 2018-05-15
Improve test coverage

janos-ss:


language-server/src/main/java/org/sonarlint/languageserver/ConnectedEngineFactory.java
```diff
@@ -68,4 +69,9 @@ ConnectedSonarLintEngine create(ServerInfo serverInfo) {
   void putExtraProperty(String name, String value) {
     extraProperties.put(name, value);
   }
+
+  @VisibleForTesting
```
dbmeneses: I don't think the language server is depending on guava, it probably has access to it transitively. Either way I would rather comment it since we are trying to get rid of guava everywhere.

PR #198, opened 2018-05-21
SLCORE-188 Update SQ Plugin API to 7.1 + few other dependencies

henryju:


core/src/main/java/org/sonarsource/sonarlint/core/container/global/GlobalExtensionContainer.java
```diff
@@ -32,8 +32,6 @@ public GlobalExtensionContainer(ComponentContainer parent) {
 
   @Override
   protected void doBeforeStart() {
-    add(GlobalSettings.class);
-    add(new GlobalConfigurationProvider());
```
janos-ss: A question, just for my information. I see you moved these to `StandaloneGlobalContainer` and `StorageContainer`. 
What is change about, and how does it work?

henryju: This is because ExtensionInstaller now need to have Configuration injected, so I moved them one level up.

PR #200, opened 2018-05-23
Don't always limit WS calls to 20 pages

dbmeneses:


core/src/test/java/org/sonarsource/sonarlint/core/WsHelperImplTest.java
```diff
@@ -164,6 +169,33 @@ public void testUnsupportedServer() {
...
+    for (int i = 0; i < 21; i++) {
+      mockOrganizationsPage(i + 1, 10500);
+    }
+
+    List<RemoteOrganization> orgs = WsHelperImpl.listOrganizations(client, serverChecker, new ProgressWrapper(null));
+    assertThat(orgs).hasSize(10500);
+  }
+
+  private void mockOrganizationsPage(int page, int total) throws IOException {
+    List<Organization> orgs = IntStream.range(1, 501)
```
janos-ss: More natural with `IntStream.rangeClosed(1, 500)`

dbmeneses: definitely, thanks

PR #203, opened 2018-05-28
SLCORE-189 Make it possible to include inactive rules in standalone mode

janos-ss:


core/src/main/java/org/sonarsource/sonarlint/core/container/standalone/rule/StandaloneActiveRules.java
```diff
@@ -0,0 +1,146 @@
...
+   */
+  private static Set<org.sonar.api.rule.RuleKey> convert(Collection<RuleKey> includedRuleKeys) {
+    return includedRuleKeys.stream().map(r -> org.sonar.api.rule.RuleKey.of(r.repository(), r.rule())).collect(Collectors.toSet());
+  }
+
+  private ActiveRules filtered(Set<org.sonar.api.rule.RuleKey> excluded, Set<org.sonar.api.rule.RuleKey> included) {
+    // accept from active those that are not explicitly excluded
+    Predicate<ActiveRule> filterActive = ar -> !excluded.contains(ar.ruleKey());
+
+    // accept from ubactive those that are explicitly included
```
henryju: typo ubactive -> inactive

core/src/test/java/org/sonarsource/sonarlint/core/mediumtest/StandaloneIssueMediumTest.java
```diff
@@ -380,8 +380,9 @@ public void simpleJavaWithExcludedRules() throws Exception {
       false);
 
     final Collection<RuleKey> excludedRules = Collections.singleton(new RuleKey("squid", "S106"));
+    final Collection<RuleKey> includedRules = Collections.singleton(new RuleKey("javascript", "S2424"));
```
henryju: I'm surprised there is no assertion updated. Isn't it supposed to raise a few new issues?

henryju:
as long as the behavior is clearly decided (and tested :)) that's fine!

PR #204, opened 2018-05-29
A fix for virus scanners preventing directory moves on Windows

ivmarkov:
A fix for 
[this](https://stackoverflow.com/questions/49442860/how-to-solve-the-unable-to-update-binding-error-in-sonar-lint?rq=1) problem.

TL;DR: Some virus scanners have the nasty habit of listening the Windows filesystem for changes and opening a file for scanning immediately after the file had been created by some program (in this case, the SonarLint Eclipse plugin). 

Unfortunately this means that a move or deletion operation on the file (or directory where the file resides) might fail. We've identified this problem in our own code several years ago, and the only workaround is to try to repeat the operation a couple of times until it eventually succeeds.

Fortunately, the code changes for SonarLint are seemingly minimalistic - a very unobtrusive change in FileUtils.java is all that was necessary.

### DIFF
--- a/client-api/src/main/java/org/sonarsource/sonarlint/core/client/api/util/FileUtils.java
+++ b/client-api/src/main/java/org/sonarsource/sonarlint/core/client/api/util/FileUtils.java
@@ -21,6 +21,7 @@ package org.sonarsource.sonarlint.core.client.api.util;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.AccessDeniedException;
 import java.nio.file.AtomicMoveNotSupportedException;
 import java.nio.file.CopyOption;
 import java.nio.file.FileVisitResult;
@@ -33,9 +34,18 @@ import java.util.function.Consumer;
 import java.util.regex.Pattern;
 
 public class FileUtils {
-
+  /**
+   * A simple representation of an IO operation.
+   * An internal interface necessary for the implementation of {@link #retry()}.
+   */
+  private interface IORunnable {
+	  void run() throws IOException;
+  }
+  
   private static final String PATH_SEPARATOR_PATTERN = Pattern.quote(File.separator);
 
+  private static Boolean windows;
+  
   private FileUtils() {
     // utility class, forbidden constructor
   }
@@ -50,7 +60,7 @@ public class FileUtils {
 
   private static void moveDirPreferAtomic(Path src, Path dest) throws IOException {
     try {
-      Files.move(src, dest, StandardCopyOption.ATOMIC_MOVE);
+      retry(() -> Files.move(src, dest, StandardCopyOption.ATOMIC_MOVE));
     } catch (AtomicMoveNotSupportedException e) {
       // Fallback to non atomic move
       Files.walkFileTree(src, new CopyRecursivelyVisitor(src, dest));
@@ -98,13 +108,13 @@ public class FileUtils {
       Files.walkFileTree(path, new SimpleFileVisitor<Path>() {
         @Override
         public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException {
-          Files.delete(file);
+          retry(() -> Files.delete(file));
           return FileVisitResult.CONTINUE;
         }
 
         @Override
         public FileVisitResult postVisitDirectory(Path dir, IOException exc) throws IOException {
-          Files.delete(dir);
+          retry(() -> Files.delete(dir));
           return FileVisitResult.CONTINUE;
         }
       });
@@ -151,5 +161,48 @@ public class FileUtils {
     FileUtils.deleteRecursively(target);
     FileUtils.mkdirs(target.getParent());
     FileUtils.moveDir(work, target);
+  }
+  
+  /**
+   * On Windows, retries the provided IO operation a number of times, in an effort to make the operation succeed.
+   * 
+   * Operations that might fail on Windows are file & directory move, as well as file deletion. These failures 
+   * are typically caused by the virus scanner and/or the Windows Indexing Service. These services tend to open a file handle 
+   * on newly created files in an effort to scan their content.
+   * 
+   * @param runnable the runnable whose execution should be retried
+   */
+  private static void retry(IORunnable runnable) throws IOException {
+    for (int retries = isWindows()? 30: 1; retries > 0; retries--) {
+      if (retries > 1) {    	  
+	    try {
+	      runnable.run();
+	      break;
+	    } catch(AccessDeniedException e) {
+	      // Sleep a bit to give a chance to the virus scanner / Windows Indexing Service to release the opened file handle
+	      try {
+	        Thread.sleep(100);
+	      } catch (InterruptedException ie) {
+	    	// Nothing meaningful to be done here
+	      }
+	    }
+      } else {
+    	// Give it a one last chance, and this time do not swallow the exception
+        runnable.run();
+      }
+    }
+  }
+  
+  /**
+   * A simple check whether the underlying operating system is Windows. 
+   * @return true if the OS is Windows
+   */
+  private static boolean isWindows() {
+	  synchronized (FileUtils.class) {
+		if (windows == null)
+		  windows = System.getProperty("os.name").startsWith("Windows");
+	  }
+	  
+	  return windows;
   }
 }


client-api/src/main/java/org/sonarsource/sonarlint/core/client/api/util/FileUtils.java
```diff
@@ -152,4 +162,47 @@ public static void replaceDir(Consumer<Path> dirContentUpdater, Path target, Pat
...
+   * On Windows, retries the provided IO operation a number of times, in an effort to make the operation succeed.
+   * 
+   * Operations that might fail on Windows are file & directory move, as well as file deletion. These failures 
+   * are typically caused by the virus scanner and/or the Windows Indexing Service. These services tend to open a file handle 
+   * on newly created files in an effort to scan their content.
+   * 
+   * @param runnable the runnable whose execution should be retried
+   */
+  private static void retry(IORunnable runnable) throws IOException {
+    for (int retries = isWindows()? 30: 1; retries > 0; retries--) {
```
dbmeneses: isn't 30 a little excessive? It would take 3seconds to fail, which is a long time in SonarLint environment.

client-api/src/main/java/org/sonarsource/sonarlint/core/client/api/util/FileUtils.java
```diff
@@ -33,9 +34,18 @@
 import java.util.regex.Pattern;
 
 public class FileUtils {
-
+  /**
+   * A simple representation of an IO operation.
+   * An internal interface necessary for the implementation of {@link #retry()}.
+   */
+  private interface IORunnable {
```
dbmeneses: You could mark it with `FunctionalInterface`

client-api/src/main/java/org/sonarsource/sonarlint/core/client/api/util/FileUtils.java
```diff
@@ -152,4 +162,47 @@ public static void replaceDir(Consumer<Path> dirContentUpdater, Path target, Pat
...
+  }
+  
+  /**
+   * A simple check whether the underlying operating system is Windows. 
+   * @return true if the OS is Windows
+   */
+  private static boolean isWindows() {
+	  synchronized (FileUtils.class) {
+		if (windows == null)
+		  windows = System.getProperty("os.name").startsWith("Windows");
```
dbmeneses: This will result in a NPE if for some reason os.name is not defined.

client-api/src/main/java/org/sonarsource/sonarlint/core/client/api/util/FileUtils.java
```diff
@@ -152,4 +162,47 @@ public static void replaceDir(Consumer<Path> dirContentUpdater, Path target, Pat
...
+      }
+    }
+  }
+  
+  /**
+   * A simple check whether the underlying operating system is Windows. 
+   * @return true if the OS is Windows
+   */
+  private static boolean isWindows() {
+	  synchronized (FileUtils.class) {
```
dbmeneses: The value could be simply statically assigned to `windows`. There is no need to calculate it lazily since we know it will always be needed and it avoids to do the synchronization every time.

client-api/src/main/java/org/sonarsource/sonarlint/core/client/api/util/FileUtils.java
janos-ss: I think it would be good to extract `isWindows()? 30: 1` to a `getMaxRetries()` static helper.

client-api/src/main/java/org/sonarsource/sonarlint/core/client/api/util/FileUtils.java
```diff
@@ -152,4 +162,47 @@ public static void replaceDir(Consumer<Path> dirContentUpdater, Path target, Pat
...
+   * 
+   * Operations that might fail on Windows are file & directory move, as well as file deletion. These failures 
+   * are typically caused by the virus scanner and/or the Windows Indexing Service. These services tend to open a file handle 
+   * on newly created files in an effort to scan their content.
+   * 
+   * @param runnable the runnable whose execution should be retried
+   */
+  private static void retry(IORunnable runnable) throws IOException {
+    for (int retries = isWindows()? 30: 1; retries > 0; retries--) {
+      if (retries > 1) {
```
janos-ss: Instead of checking the current retry count in every iteration of the loop, I would use a flag, set after successful run, and then after the loop, run one last time if the flag was not set.

client-api/src/main/java/org/sonarsource/sonarlint/core/client/api/util/FileUtils.java
```diff
@@ -152,4 +162,47 @@ public static void replaceDir(Consumer<Path> dirContentUpdater, Path target, Pat
...
+	      runnable.run();
+	      break;
+	    } catch(AccessDeniedException e) {
+	      // Sleep a bit to give a chance to the virus scanner / Windows Indexing Service to release the opened file handle
+	      try {
+	        Thread.sleep(100);
+	      } catch (InterruptedException ie) {
+	    	// Nothing meaningful to be done here
+	      }
+	    }
```
janos-ss: Please use indentation consistent with the rest of the file. (2 spaces per level)

client-api/src/main/java/org/sonarsource/sonarlint/core/client/api/util/FileUtils.java
ivmarkov: We don't expect it to fail. *We expect it so succeed*. Moreover, it would succeed on the second or third retry at most which is *100ms wait time or 200ms at most*. Always, or almost always. But if - for whatever reason - once in a lifetime - the PC is under a heavy load, and McAfee is stressed to complete its virus checking on time, we don't want to report failure. The user would then wait for 3 seconds. Not a big deal IMHO.

I think if you want to continue arguing over the same lines, you should concentrate on the Thread.sleep timeout, rather than on the number of retry, as at least one Thread.sleep call will always happen on Windows, if McAfee is installed and screwing things for us.

client-api/src/main/java/org/sonarsource/sonarlint/core/client/api/util/FileUtils.java
ivmarkov: Obvious in retrospective. Will fix it accordingly, thanks!

client-api/src/main/java/org/sonarsource/sonarlint/core/client/api/util/FileUtils.java
ivmarkov: That was how the code looked like originally. A question of preference, if you ask me. But since I'm anyway fixing stuff, I can fix that as well. Anything against early exits/returns?

client-api/src/main/java/org/sonarsource/sonarlint/core/client/api/util/FileUtils.java
ivmarkov: Sorry for that - missed that one (we still use the old 4 spaces per level in our own codebase).

client-api/src/main/java/org/sonarsource/sonarlint/core/client/api/util/FileUtils.java
dbmeneses: I agree

client-api/src/main/java/org/sonarsource/sonarlint/core/client/api/util/FileUtils.java
dbmeneses: Alright then.
Just a note: here you are assuming that when an `AccessDeniedException` happens, it's a temporary situation. However, I imagine this happening due to an actual lack of permissions and in this case we want to fail as fast as possible, hence my comment. In this situation what matters is the total timeout, not the number retries or the duration of each sleep.

ivmarkov: Right... yet: what is the chance the user not to have permissions in - say - the temporary space where these files reside.

Moreover, if there is a general permissions' issue, I don't think you can recover from that. In that case, most likely the stack will be unrolled and a general error (maybe even internal error?) will appear in Eclipse, VS etc. The fact that it will appear 3 seconds later is probably not the end of the world.

ivmarkov:
Will this ever get merged into the SonarLint plugin for Eclipse? 
More than two months had passed, and the latest Eclipse plugin is still failing on file move operations.

ivmarkov:
Hello? 
Sorry for being a bit direct, but - do you guys seriously believe, that your Eclipse plugin will ever get used in a corporate environment with this issue not fixed?

janos-ss:
Hi @ivmarkov, sorry for not replying to your previous message.

This PR has been merged a while ago, but it just missed the previous release of SonarLint for Eclipse. It will be part of the next one, which should be coming soon, within a month or so.

PR #208, opened 2018-06-12
Convert workspace folder and file uris correctly to paths

janos-ss:


language-server/src/main/java/org/sonarlint/languageserver/SonarLintLanguageServer.java
```diff
@@ -560,7 +567,7 @@ private static URI parseURI(String uriStr) {
 
   // visible for testing
   void analyze(URI uri, String content, boolean shouldFetchServerIssues) {
-    if (!uri.toString().startsWith("file:/")) {
+    if (!Paths.get(uri).toFile().isFile()) {
```
henryju: This will throw FileSystemNotFoundException if the URI doesn't start with file://

janos-ss: @henryju indeed (what was I thinking). Fixed.

language-server/src/main/java/org/sonarlint/languageserver/SonarLintLanguageServer.java
```diff
@@ -378,7 +379,13 @@ private boolean updateModuleStorage(ConnectedSonarLintEngine engine, ServerInfo
...
 
   private static List<String> toList(List<WorkspaceFolder> workspaceFolders) {
-    return workspaceFolders.stream().map(f -> f.getUri().replaceAll("^file://", "")).collect(Collectors.toList());
+    return workspaceFolders.stream()
+      .map(f -> normalizeUriString(f.getUri()))
+      .collect(Collectors.toList());
+  }
+
+  static String normalizeUriString(String uriString) {
+    return Paths.get(URI.create(uriString)).toFile().toString();
```
henryju: You can have the same problem here, but I guess it's fine since VSCode and Atom only support file:// protocol as far as I know.

janos-ss: Good catch. I don't think they support either. But someday they might. I added a filter in the caller to protect against that, avoiding a potential crash.

PR #215, opened 2018-07-31
SONARWEB-72 Add SonarHTML to whitelist

pynicolas:


pynicolas:
We finally decided to not change the pluginKey for SonarHTML: that makes this PR useless.

PR #217, opened 2018-08-06
Support secondary locations

henryju:


core/src/main/java/org/sonarsource/sonarlint/core/analyzer/issue/DefaultClientIssue.java
```diff
@@ -118,11 +122,13 @@ public String getMessage() {
...
   }
 
-  private static class DefaultFlow implements Flow {
+  private class DefaultFlow implements Flow {
     private List<org.sonarsource.sonarlint.core.client.api.common.analysis.IssueLocation> locations;
 
     private DefaultFlow(List<IssueLocation> issueLocations) {
       this.locations = issueLocations.stream()
+        // Exclude cross file locations
+        .filter(l -> l.inputComponent().key().equals(fileKey))
```
dbmeneses: I wonder if this shouldn't be done before passing the flows to the `DefaultClientIssue`, which so far was just holding information (and would also avoid passing the `fileKey`)

henryju:
@dbmeneses  I have reworked the code

PR #221, opened 2018-08-08
Telemetry + minor improvements

henryju:


language-server/src/main/java/org/sonarlint/languageserver/SonarLintLanguageServer.java
```diff
@@ -345,6 +342,7 @@ private void updateBinding(@Nullable Map<?, ?> connectedModeProject) {
       logger.error(ClientLogger.ErrorType.START_CONNECTED_ENGINE_FAILED);
       return;
     }
+    telemetry.usedConnectedMode(serverInfo.serverUrl);
```
janos-ss: Why not call this after successful binding is confirmed, a few lines below? That would be a bit more accurate.

henryju: ok

PR #222, opened 2018-08-27
SLCORE-207 Support ruby and kotlin plugins

saberduck:


saberduck:
Should I use unreleased dependency for IT tests?

PR #223, opened 2018-09-19
Drop modules

dbmeneses:
ITs are passing, but I will still add ITs to test the new features


client-api/src/main/java/org/sonarsource/sonarlint/core/client/api/connected/RemoteProject.java
```diff
@@ -19,14 +19,12 @@
  */
 package org.sonarsource.sonarlint.core.client.api.connected;
 
-public interface RemoteModule {
+public interface RemoteProject {
 
   String getProjectKey();
```
henryju: This is now useless and duplicates geKey

PR #235, opened 2018-12-11
SLCORE-205 Workaround the 10k api/rules/search limit

dbmeneses:


core/src/main/java/org/sonarsource/sonarlint/core/container/connected/update/RulesDownloader.java
```diff
@@ -87,12 +97,13 @@ private void fetchRulesAndActiveRules(Rules.Builder rulesBuilder, Map<String, Ac
...
   }
 
-  private String getUrl(int page, int pageSize) {
+  private String getUrl(String severity, int page, int pageSize) {
     StringBuilder builder = new StringBuilder(1024);
     builder.append(RULES_SEARCH_URL);
     if (wsClient.getOrganizationKey() != null) {
       builder.append("&organization=").append(StringUtils.urlEncode(wsClient.getOrganizationKey()));
     }
+    builder.append("&severities=" + severity);
```
henryju: you could use append

dbmeneses: fixed

PR #239, opened 2019-01-28
Hardening

henryju:


core/src/main/resources/plugins_min_versions.txt
```diff
@@ -15,7 +16,7 @@ typescript=1.5
 web=2.6
 kotlin=1.2.1
 ruby=1.2.1
-scala=1.2.1
+sonarscala=1.2.1
```
dbmeneses: Was the key incorrect or did it change at some point? If it's the latter, how did the version number not change?

PR #244, opened 2019-03-28
Run ITs on the dogfood branch to detect regressions earlier

henryju:


.cix.yml
```diff
@@ -2,7 +2,7 @@
 
 
 SQ_VERSION:
-  - DEV
+  - DOGFOOD
```
julienlancelot: You also update https://github.com/SonarSource/sonarlint-core/blob/master/run-integration-tests.sh#L39 to fix QA

PR #246, opened 2019-06-20
SLCORE-228 Storage could be corrupted on case-insensitive filesystems

henryju:


core/src/main/java/org/sonarsource/sonarlint/core/container/storage/StoragePaths.java
```diff
@@ -118,7 +120,32 @@ public Path getQProfilesPath() {
...
+  }
+
+  /**
+   * Base64 encoding used by recent SonarQube for quality profile keys are not playing well with case insensitive FS. There is a risk
+   * of having two quality profiles keys collisions.
+   * If the key is a valid base 64 encoded string, we will reencode it as hexa instead (a bit longer, but resistant to case insensitive FS). 
+   */
+  static String tryConvertBase64UuidToString(String qProfileKey) {
+    try {
+      byte[] decoded = Base64.getUrlDecoder().decode(qProfileKey);
```
dbmeneses: I wonder if we had to decode the key instead of simply transforming to hex?

henryju: Pros of your approach: works consistently for all strings
Cons: string will be a bit longer
How to choose ?

dbmeneses: no idea :) let's leave as it is

henryju: Finally I simplified everything and use hex encoding for all strings we are using as filename, not only quality profile keys.

core/src/main/java/org/sonarsource/sonarlint/core/container/storage/StoragePaths.java
```diff
@@ -121,6 +115,16 @@ public Path getActiveRulesPath(String qProfileKey) {
     return getGlobalStorageRoot().resolve(ACTIVE_RULES_FOLDER).resolve(encodeForFs(qProfileKey) + ".pb");
   }
 
+  private static String bytesToHex(byte[] bytes) {
```
jblievremont: This looks very close to `org.sonarsource.sonarlint.core.tracking.DigestUtils#encodeHexString`, and a complete duplicate of `org.sonarsource.sonarlint.core.container.UidFactoryFast#byteArrayToHex`. Is there a particular reason for this?

henryju: I was not remembering those methods existed :) Good catch!

PR #251, opened 2019-07-05
Update orchestrator version to verify if ITs did not break

jacek-poreda-sonarsource:
- Update orchestrator version after removal of deprecated API in SonarQube


jacek-poreda-sonarsource:
Thanks! I'll merge this PR as soon as I will release new version of Orchestrator

PR #252, opened 2019-07-05
More hardening

henryju:


core/src/main/java/org/sonarsource/sonarlint/core/container/analysis/AnalysisTempFolderProvider.java
```diff
@@ -19,61 +19,47 @@
...
+    public File newDir(String name) {
+      throw throwUOEFolder();
+    }
 
-  @Override
-  public boolean componentHasLifecycle() {
-    return true;
-  }
+    private static UnsupportedOperationException throwUOEFolder() {
+      return new UnsupportedOperationException("Don't create temp folders during analysis");
```
jblievremont: I think the file/dir name should appear in the message, e.g:
> Creation of temporary file/folder '%s' is not allowed by SonarLint

henryju: The stacktrace will show the plugin trying to call this method, so I don't think it is very important, we will have other way to identify the root cause.

PR #256, opened 2019-07-17
SLCORE-237 Add support of SonarGo 1.6 (slang-based)

m-g-sonar:


m-g-sonar:
@henryju I have not idea why ITs does not run, but it fails also on master for me. I also tried branch #254 to check if it could come from this, but I get the same issue.

Note that without having at least this PR passing QA, I can not move forward with the SLang ITs checking for compatiblity with SonarGo. https://github.com/SonarSource/slang-enterprise/pull/86

core/src/main/resources/plugins_min_versions.txt
```diff
@@ -19,5 +19,6 @@ tsql=1.1.0.2177
 typescript=1.5.0.2122
 web=3.0.0.1411
 xml=2.0.1.2020
+go=1.6.0.673
```
henryju: Can you please keep alphabetic ordering?

its/tests/src/test/java/its/ConnectedModeTest.java
```diff
@@ -114,6 +115,7 @@
     .addPlugin(MavenLocation.of("org.sonarsource.slang", "sonar-kotlin-plugin", System.getProperty("kotlinVersion")))
     .addPlugin(MavenLocation.of("org.sonarsource.slang", "sonar-ruby-plugin", System.getProperty("rubyVersion")))
     .addPlugin(MavenLocation.of("org.sonarsource.slang", "sonar-scala-plugin", System.getProperty("scalaVersion")))
+    .addPlugin(MavenLocation.of("org.sonarsource.slang", "sonar-go-plugin", System.getProperty("goVersion")))
```
henryju: For the system property to exist, you have to update also https://github.com/SonarSource/sonarlint-core/blob/master/run-integration-tests.sh

core/src/main/resources/plugins_stream_support.txt
```diff
@@ -16,4 +16,5 @@ kotlin=1.1
 ruby=1.1
 sonarscala=1.4
 tsql=1.1
-xml=2.0.1
\ No newline at end of file
+xml=2.0.1
+go=1.6
```
m-g-sonar: just realized that there is two times XML in this file.

run-integration-tests.sh
```diff
@@ -37,6 +38,8 @@ case "$SQ_VERSION" in
     KOTLIN_VERSION=LATEST_RELEASE
     RUBY_VERSION=LATEST_RELEASE
     SCALA_VERSION=LATEST_RELEASE
+    # SonarLint-compatible Go plugin not yet released
+    GO_VERSION=DEV
```
m-g-sonar: to be changed once released?

m-g-sonar:
Updated according to what we discussed @henryju ! Thanks for the support!

core/src/main/resources/plugins_stream_support.txt
henryju: That's the benefit of lexicographic sorting like in the other file ;)

run-integration-tests.sh
m-g-sonar: Yep that's why I changed the version to a fixed version!

m-g-sonar:
As discussed offline, we postpone the support of SonarLint for SonarG, due to performance investigation.

PR #262, opened 2019-08-23
SLCORE-245 Improve the prefix matching algorithm

henryju:


core/src/test/java/org/sonarsource/sonarlint/core/container/storage/FileMatcherTest.java
```diff
@@ -103,6 +100,46 @@ public void should_return_most_common_prefixes() {
...
+      Paths.get("sq3/pom.xml"),
+      Paths.get("sq3/build.properties"),
+      Paths.get("sq4/src/A.java")
+
+    );
+    FileMatcher.Result match = fileMatcher.match(sqPaths, idePaths);
+    assertThat(match.idePrefix()).isEqualTo(Paths.get("local1"));
+    assertThat(match.sqPrefix()).isEqualTo(Paths.get("sq4"));
+  }
+
```
jblievremont:
```suggestion

  @Test
  public void should_match_first_similar_hierarchy() throws Exception {
    List<Path> idePaths = Arrays.asList(
      Paths.get("local/sub/index.html"),
      Paths.get("local/sub/product1/index.html"),
      Paths.get("local/sub/product2/index.html"),
      Paths.get("local/sub/product3/index.html")
    );

    List<Path> sqPaths = Arrays.asList(
      Paths.get("sq/index.html"),
      Paths.get("sq/news/index.html"),
      Paths.get("sq/news/product1/index.html"),
      Paths.get("sq/news/product2/index.html"),
      Paths.get("sq/news/product3/index.html"),
      Paths.get("sq/products/index.html"),
      Paths.get("sq/products/product1/index.html"),
      Paths.get("sq/products/product2/index.html"),
      Paths.get("sq/products/product3/index.html"),
      Paths.get("sq/company/index.html"),
      Paths.get("sq/company/jobs/index.html"),
      Paths.get("sq/company/news/index.html"),
      Paths.get("sq/company/contact/index.html")
    );

    FileMatcher.Result match = fileMatcher.match(sqPaths, idePaths);
    assertThat(match.idePrefix()).isEqualTo(Paths.get("local/sub"));
    assertThat(match.sqPrefix()).isEqualTo(Paths.get("sq/news"));
  }
```

I ran this on my machine to check the stability of the behavior in ambiguous cases, I think it might be relevant to add it for documentation purposes.

core/src/test/java/org/sonarsource/sonarlint/core/container/storage/FileMatcherTest.java
```diff
@@ -103,6 +100,46 @@ public void should_return_most_common_prefixes() {
...
+      Paths.get("local1/src/A.java"));
+
+    List<Path> sqPaths = Arrays.asList(
+      Paths.get("sq1/pom.xml"),
+      Paths.get("sq1/build.properties"),
+      Paths.get("sq2/pom.xml"),
+      Paths.get("sq2/build.properties"),
+      Paths.get("sq3/pom.xml"),
+      Paths.get("sq3/build.properties"),
+      Paths.get("sq4/src/A.java")
```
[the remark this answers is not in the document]
henryju: With your proposed change, we are back at testing that the prefix with the biggest number of file matched is the preferred one, which was already the case before. Here I want to test that sq4 will be selected even if this is not the one having the most matched files, but because `pom.xml` and `build.properties` are matched many times, their score will be lower than the single match of `src/A.java`

core/src/test/java/org/sonarsource/sonarlint/core/container/storage/FileMatcherTest.java
henryju: As discussed I don't think it is testing more cases than other existing tests, but I will add it as a "more realistic" test to have a safety net.

PR #263, opened 2019-08-26
Hardening

henryju:


core/src/main/java/org/sonarsource/sonarlint/core/container/analysis/filesystem/SonarLintInputFile.java
```diff
@@ -245,4 +249,28 @@ public boolean hasNoSonarAt(int line) {
     return this.noSonarLines.contains(line);
   }
 
+  public boolean isIgnoreAllIssues() {
+    checkMetadata();
```
henryju: @dbmeneses this `checkMetadata()` statement is missing in the scanner-engine, and I think it is a bug. Thanks to the medium test, I found that if the first issue reported by the analyzer is a file level issue (like squid:S1220), then metadata are not initialized, and so the `ignoreAllFiles` boolean has not been set.

core/src/main/java/org/sonarsource/sonarlint/core/container/analysis/filesystem/SonarLintInputFile.java
```diff
@@ -245,4 +249,28 @@ public boolean hasNoSonarAt(int line) {
...
+
+  public void addIgnoreIssuesOnLineRanges(Collection<int[]> lineRanges) {
+    if (this.ignoreIssuesOnlineRanges == null) {
+      this.ignoreIssuesOnlineRanges = new ArrayList<>();
+    }
+    this.ignoreIssuesOnlineRanges.addAll(lineRanges);
+  }
+
+  public boolean isIgnoreAllIssuesOnLine(@Nullable Integer line) {
+    checkMetadata();
```
henryju: Same here, it is missing in scanner-engine, and I think it's a bug.

core/src/main/java/org/sonarsource/sonarlint/core/container/analysis/issue/ignore/scanner/IssueExclusionsRegexpScanner.java
```diff
@@ -0,0 +1,170 @@
...
+  IssueExclusionsRegexpScanner(SonarLintInputFile inputFile, List<Pattern> allFilePatterns, List<DoubleRegexpMatcher> blockMatchers) {
+    this.allFilePatterns = allFilePatterns;
+    this.blockMatchers = blockMatchers;
+    this.inputFile = inputFile;
+    LOG.debug("Evaluate issue exclusions for '{}'", inputFile.relativePath());
+  }
+
+  @Override
+  public void handleIgnoreEoL(char c) {
+    if (ignoreAllIssues) {
```
henryju: To avoid infinite loop, I introduced a local variable to short circuit. The loop was: IgnoreIssuesFilter -> InputFile.isIgnoreAllIssues -> checkMetadata -> initMetadata -> IssueExclusionsRegexpScanner -> InputFile.isIgnoreAllIssues

core/src/main/java/org/sonarsource/sonarlint/core/container/analysis/filesystem/SonarLintInputFile.java
henryju: Don't forget to create the same ticket in SC

PR #268, opened 2019-09-11
SLCORE-247 Enable rules based on IDE settings

jblievremont:
Opening this against current 4.5 master to build PR-artifact, should be merged only with a 4.6 master.


jblievremont:
Will be merged alongside #269

PR #269, opened 2019-09-16
SLCORE-248 Add LSP request to list all available rules

jblievremont:


language-server/src/main/java/org/sonarlint/languageserver/ISonarLintLanguageServer.java
```diff
@@ -0,0 +1,33 @@
...
+package org.sonarlint.languageserver;
+
+import org.eclipse.lsp4j.jsonrpc.services.JsonRequest;
+import org.eclipse.lsp4j.services.LanguageServer;
+
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+
+public interface ISonarLintLanguageServer extends LanguageServer {
```
henryju: I don't like the "I" prefix. What about `SonarLintExtendedLanguageServer`?

language-server/src/main/java/org/sonarlint/languageserver/ISonarLintLanguageServer.java
```diff
@@ -0,0 +1,33 @@
...
+
+  @JsonRequest("SonarLint/listAllRules")
```
henryju: We were using lowercase for `sonarlint/openRuleDescription`
```suggestion
  @JsonRequest("sonarlint/listAllRules")
```

language-server/src/main/java/org/sonarlint/languageserver/RuleDescription.java
```diff
@@ -90,17 +104,27 @@ public RuleDescription setSeverity(String severity) {
...
+    return this;
+  }
+
   @Override
   public String toString() {
     return MessageJsonHandler.toString(this);
   }
 
-  public static RuleDescription of(String key, String name, String htmlDescription, String type, String severity) {
+  public static RuleDescription of(String key, String name, @Nullable String htmlDescription, @Nullable String type, String severity, boolean activeByDefault) {
```
henryju: Do we really have rules with null description/type?

language-server/src/main/java/org/sonarlint/languageserver/SonarLintLanguageServer.java
```diff
@@ -582,6 +581,21 @@ private static URI parseURI(String uriStr) {
...
+    Map<String, String> languagesNameByKey = engineCache.getOrCreateStandaloneEngine().getAllLanguagesNameByKey();
+    engineCache.getOrCreateStandaloneEngine().getAllRuleDetails()
+      .forEach(d -> {
+        String languageName = languagesNameByKey.get(d.getLanguageKey());
+        if (!result.containsKey(languageName)) {
+          result.put(languageName, new ArrayList<>());
+        }
+        result.get(languageName).add(RuleDescription.of(d));
+      });
+    return CompletableFuture.completedFuture(result);
```
henryju: Change to use the correct pattern for cancellation support. See my other PR for examples.

language-server/src/main/java/org/sonarlint/languageserver/RuleDescription.java
jblievremont: I don't think so, but they have `@CheckForNull` on `org.sonarsource.sonarlint.core.client.api.common.RuleDetails`

henryju: OK, fine then

PR #274, opened 2019-09-23
Rework logging

henryju:
* delete specific logging facade for issue tracking
* handle vscode specific $/setTraceNotification

### DIFF
--- a/language-server/src/main/java/org/sonarlint/languageserver/SonarLintLanguageServer.java
+++ b/language-server/src/main/java/org/sonarlint/languageserver/SonarLintLanguageServer.java
@@ -41,7 +41,6 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.Set;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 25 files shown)


language-server/src/main/java/org/sonarlint/languageserver/SonarLintLanguageServer.java
```diff
@@ -875,6 +866,18 @@ public void didChangeWorkspaceFolders(DidChangeWorkspaceFoldersParams params) {
...
 
+  @Override
+  public void setTraceNotification(SetTraceNotificationParams params) {
+    this.traceLevel = parseTraceLevel(params.getValue());
+  }
+
+  private static TraceValues parseTraceLevel(@Nullable String trace) {
+    return Optional.ofNullable(trace)
+      .map(String::toUpperCase)
+      .map(TraceValues::valueOf)
```
jblievremont: This has the potential to throw exceptions when `trace` is not one of the accepted trace levels. I'd be more comfortable e.g with a "safe parse" method on `TraceValues` (with default OFF).

PR #275, opened 2019-09-23
Workspace manager

henryju:
@jblievremont Only look at the second commit, the first one is already in the other PR

### DIFF
--- a/language-server/src/main/java/org/sonarlint/languageserver/SonarLintLanguageServer.java
+++ b/language-server/src/main/java/org/sonarlint/languageserver/SonarLintLanguageServer.java
@@ -25,7 +25,6 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.Socket;
 import java.net.URI;
-import java.net.URISyntaxException;
 import java.net.URL;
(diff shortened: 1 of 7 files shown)


language-server/src/main/java/org/sonarlint/languageserver/SonarLintLanguageServer.java
```diff
@@ -625,7 +599,7 @@ public AnalysisResultsWrapper analyze(Path baseDir, URI uri, String content, Iss
     @Override
     public boolean isExcludedByServerSideExclusions(URI fileUri) {
       Path baseDir = findBaseDir(fileUri);
-      ProjectBinding projectBinding = workspaceBindings.getOrDefault(fileUri.toString(), new ProjectBinding(projectKey, "", ""));
+      ProjectBinding projectBinding = workspaceBindings.getOrDefault(baseDir, new ProjectBinding(projectKey, "", ""));
```
henryju: @jblievremont I think here I spotted a bug. The wrong map key was used, so always ending using the default value. It probably means issue exclusion was unreliable when path prefixes are not empty.

PR #279, opened 2019-10-04
Set up CI with Azure Pipelines

henryju:


azure-pipelines.yml
```diff
@@ -0,0 +1,260 @@
...
+
+variables:
+  - group: sonarsource-build-variables
+
+resources:
+  repositories:
+    - repository: commonTemplates
+      type: git
+      name: pipelines-yaml-templates
+      # ref: refs/tags/v1.0
```
henryju: To be uncommented after review and having tagged the templates

its/tests/src/test/java/its/CommercialAnalyzerTest.java
```diff
@@ -57,7 +57,19 @@
   private static final String PROJECT_KEY_TSQL = "sample-tsql";
   private static final String PROJECT_KEY_APEX = "sample-apex";
 
-  private static Orchestrator ORCHESTRATOR;
+  @ClassRule
```
jblievremont: :+1:

its/tests/src/test/java/its/tools/ItUtils.java
```diff
@@ -19,9 +19,72 @@
...
+        javaVersion = props.getProperty("java");
+        pythonVersion = props.getProperty("python");
+        phpVersion = props.getProperty("php");
+        // Can't test the version 4.0.0 since there was a change of custom rule API
+        javascriptVersion = "4.2.0.6476";
+        typescriptVersion = props.getProperty("typescript");
+        kotlinVersion = props.getProperty("kotlin");
+        rubyVersion = props.getProperty("ruby");
+        scalaVersion = props.getProperty("sonarscala");
+        // Because of SONARHTML-91, it is simpler to test with 3.0.1.1444
```
jblievremont: Is this still relevant?

henryju: Yes, the minimal version we support is 3.0.0.1411. Another option would be to raise minimal version to 3.0.1.1444.

henryju:
@duncanp-sonar I reported the issue here: https://github.com/microsoft/azure-pipelines-yaml/issues/359

PR #286, opened 2019-11-18
Clean dependencies

henryju:


henryju:
@jblievremont Various changes/cleanup that I discovered "by chance".

core/pom.xml
```diff
@@ -208,10 +172,12 @@
     <dependency>
       <groupId>org.junit.jupiter</groupId>
       <artifactId>junit-jupiter-engine</artifactId>
+      <scope>test</scope>
```
henryju: Only this scope was a regression from my previous changes. Those test dependencies were wrongly packaged in the artifact...

PR #289, opened 2019-11-25
Use released version of the cache task

henryju:


henryju:
@jblievremont FYI the final version of the Cache task was released, so we should update all our builds.

PR #292, opened 2020-01-06
Support deprecated rule keys in standalone mode

henryju:


its/tests/src/test/java/its/OrganizationTest.java
```diff
@@ -193,7 +193,7 @@ public void getOrganization() {
   public void verifyExtendedDescription() {
     updateGlobal();
 
-    String ruleKey = "squid:S106";
+    String ruleKey = ItUtils.javaVersion.equals("LATEST_RELEASE") ? "java:S106" : "squid:S106";
```
jblievremont: I wonder if this would be more robust e.g with `ItUtils.javaVersion.compareTo("6") > 0`, in cases where `javaVersion` is greater than `6` but not `LATEST_RELEASE`.

henryju: I don't think we need it to be robust. This is a temporary code, that we should remove when we will change minimal version of SonarJava to be >= 6.x

PR #295, opened 2020-03-18
SLCORE-265 Enable SonarJS 6.x TypeScript sensor in standalone mode

jblievremont:


henryju:
You can use this ticket number: SLCORE-265

henryju:
Just to dump idea: we could also remove entirely the condition, and decide to enable TS in SLE in the next sprint.

client-api/src/main/java/org/sonarsource/sonarlint/core/client/api/standalone/StandaloneGlobalConfiguration.java
```diff
@@ -33,10 +39,12 @@
 public class StandaloneGlobalConfiguration extends AbstractGlobalConfiguration {
 
   private final List<URL> pluginUrls;
+  private final EnumSet<Language> enabledLanguages;
```
henryju: Why not moving that to the `AbstractGlobalConfiguration` to avoid duplication?

core/src/main/java/org/sonarsource/sonarlint/core/container/global/ExtensionInstaller.java
```diff
@@ -49,8 +50,9 @@
   /**
    * Standalone mode
    */
-  public ExtensionInstaller(SonarRuntime sonarRuntime, PluginRepository pluginRepository, Configuration bootConfiguration, PluginVersionChecker pluginVersionChecker) {
-    this(sonarRuntime, pluginRepository, bootConfiguration, pluginVersionChecker, Collections.emptySet());
+  public ExtensionInstaller(SonarRuntime sonarRuntime, PluginRepository pluginRepository, Configuration bootConfiguration, PluginVersionChecker pluginVersionChecker,
```
henryju: Instead of 2 different constructors, I think now we can inject `AbstractGlobalConfiguration`

PR #297, opened 2020-03-23
SLCORE-266 Optimize file matching algorithm

henryju:
Two optimizations:
* since in real life projects, many folders will have only one child, optimize the `Node` for the "one child" case
* use `Path` as key instead of converting back and forth `String` (minor)

Performance measures on my computer (Linux jdk 1.8.0_231), using test `FileMatcherTest#performance_test` (1M SQ files, 10k project files, very deep folders)
* before: ~20s, max heap: 3750Mb
before

* after: ~9s, max heap: 2250Mb
after


henryju:
New version, more object oriented, with two implementations of Node. Same execution time, lower memory consumption. Much more readable.
after3

PR #302, opened 2020-04-29
SLCORE-269 Expose rule parameter types in client API

jblievremont:


core/src/main/java/org/sonarsource/sonarlint/core/container/standalone/rule/StandaloneRuleParamType.java
```diff
@@ -0,0 +1,42 @@
...
+
+public enum StandaloneRuleParamType {
+
+  /**
+   * Keep in sync with constants in {@link RuleParamType}
+   */
+  STRING, TEXT, BOOLEAN, INTEGER, FLOAT, SINGLE_SELECT_LIST, MULTI_SELECT_LIST;
+
+  static StandaloneRuleParamType from(RuleParamType apiType) {
+    if (apiType.multiple()) {
```
henryju, line 32: I disagree with this mapping. Even for a multi-select list, we still need to know if elements are string, int, booleans...

henryju: IMO `multiple` should be exposed the same way as `possibleValues`, in a different field.

jblievremont: I agree. I misunderstood the range of possibilities offered by the plugin API.

PR #313, opened 2020-07-07
SLCORE-274 Update plugin API to 8.4

henryju:


core/src/main/java/org/sonarsource/sonarlint/core/container/global/MapSettings.java
```diff
@@ -75,24 +83,295 @@ public MapSettings clear() {
     return this;
   }
 
+  /**
```
jblievremont, line 86: I guess this (quite large) insert comes from moving implementation code from the plugin API to SLCORE, right?

henryju: Yep, the implementation of `Settings` in SQ was moved to the sonar-plugin-api-impl package. I copied most of it, removing the support of `Encryption` to drop the dependency on commons-codec.

PR #316, opened 2020-09-09
SLI-417 Use a single method to run a cancelable section.

damien-urruty-sonarsource:


jblievremont (approved):
LGTM.

FYI, I checked in sonarlint-eclipse and sonarlint-language-server, it seems that they (currently) don't override it on their side.

PR #318, opened 2020-09-21
Update Orchestrator to support SonarQube 8.5

henryju:
This is to fix failure on the dogfood axis of QA.


henryju:
@jblievremont FYI

PR #320, opened 2020-09-24
Rework for nodejs

henryju:
Move nodejs detection to the IDE side, and add telemetry field


jblievremont (approved):
LGTM :+1: eager to try this on VSCode :smile:

PR #322, opened 2020-10-05
SLCORE-280 Consider that TypeScript language is provided by SonarJS

henryju:


jblievremont:
The hack in `org.sonarsource.sonarlint.core.plugin.PluginInfosLoader#checkUnsatisfiedPluginDependency` needs to be updated too: the check on the "TS" plugin key is not relevant anymore, it should use the `typescript` string explicitly instead.

(It took me a bit of time to understand why suddenly the JS plugin still failed to load)

PR #323, opened 2020-10-05
SLCORE-281 Unable to locate node on Windows if multiple candidates

henryju:


core/src/main/java/org/sonarsource/sonarlint/core/NodeJsHelper.java
```diff
@@ -154,7 +154,7 @@ private String runSimpleCommand(Command command) {
       msg.append("\nstderr: ").append(String.join("\n", stdErr));
     }
     LOG.debug(msg.toString());
-    if (exitCode != 0 || stdOut.size() != 1) {
+    if (exitCode != 0) {
```
jblievremont, line 157: I'd probably keep a safeguard, to prevent an exception if the command does not return anything:

```suggestion
    if (exitCode != 0 || stdOut.isEmpty()) {
```

PR #325, opened 2020-10-12
Upgrade sonar-ws to 8.5

damien-urruty-sonarsource:


damien-urruty-sonarsource:
The goal of this PR is to upgrade sonar-ws so that we can use the hotspots API. It is a prerequisite for the MMF-2167, and It won't be merge in master for the moment but on the feature branch

jblievremont (approved):
LGTM now that ITs pass 😜

its/tests/src/test/java/its/SonarCloudTest.java
```diff
@@ -166,42 +167,41 @@ public static void prepare() throws Exception {
...
+      .setOrganization(SONARCLOUD_ORGANIZATION));
   }
 
   private static void restoreProfile(String profile) {
-    adminWsClient.qualityProfiles().restoreProfile(RestoreWsRequest.builder()
-      .setBackup(Paths.get("src/test/resources/" + profile).toFile())
-      .setOrganization(SONARCLOUD_ORGANIZATION)
-      .build());
+    File backupFile = new File("src/test/resources/" + profile);
+    // XXX can't use RestoreRequest because of a bug
```
henryju, line 185: Any link to this bug? Like ticket number?

damien-urruty-sonarsource: Unfortunately, no. This issue is a regression introduced after a change in sonar-ws. I talked to someone from the SQ team about it, and they are not willing to fix it because this endpoint will probably be dropped soon. They now discourage the use of the restore feature, and they instead recommend to create a new quality profile and activate rules in it to achieve the same result.

I would like to find some time during this sprint to do the recommended move and get rid of this comment (and I agree it is really inaccurate)

PR #330, opened 2020-10-21
SLI-458 Fetch and display a hotspot

damien-urruty-sonarsource:


client-api/src/main/java/org/sonarsource/sonarlint/core/client/api/connected/GetSecurityHotspotRequestParams.java
```diff
@@ -0,0 +1,30 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.sonarlint.core.client.api.connected;
+
+public class GetSecurityHotspotRequestParams {
+  public final String hotspotKey;
```
henryju, line 23: I think on SQ side this is named uuid. Key is reserved for functional identifiers (that the user can see and possibly update).

client-api/src/main/java/org/sonarsource/sonarlint/core/client/api/connected/GetSecurityHotspotRequestParams.java
```diff
@@ -0,0 +1,30 @@
...
+  public final String projectKey;
+
+  public GetSecurityHotspotRequestParams(String hotspotKey, String projectKey) {
```
henryju, line 26: What about branch/pr? Is it not required to get the hotspot?

core/src/main/java/org/sonarsource/sonarlint/core/container/connected/vulnerability/SecurityHotspotsService.java
```diff
@@ -0,0 +1,84 @@
...
+      hotspot.hasResolution() ? RemoteHotspot.Resolution.valueOf(hotspot.getResolution()) : null,
+      adapt(hotspot.getRule()));
+  }
+
+  private static RemoteHotspot.Rule adapt(Hotspots.Rule rule) {
+    return new RemoteHotspot.Rule(rule.getKey(), rule.getName(), rule.getSecurityCategory(), RemoteHotspot.Rule.Probability.valueOf(rule.getVulnerabilityProbability()),
+      rule.getRiskDescription(), rule.getVulnerabilityDescription(), rule.getFixRecommendations());
+  }
+
+  private static String getUrl(String hotspotId, String projectKey) {
```
henryju, line 72: s/hotspotId/hotspotUuid

henryju (approved):
I would use `uuid` as a name of the hostpot identifier for consistency with SQ. Except from that LGTM.

kirill-knize-sonarsource (approved):

client-api/src/main/java/org/sonarsource/sonarlint/core/client/api/connected/GetSecurityHotspotRequestParams.java
damien-urruty-sonarsource: In the WS I use it is named `key`, I don't find any reference to `uuid` (maybe in the internals of SQ ? could you share a pointer if you have something in mind ?). Moreover UUIDs have a specific format, and I don't think the `key` described here would be considered as a UUID (example value: `AXLMPKUw2vtYjVa-RvB5`)

client-api/src/main/java/org/sonarsource/sonarlint/core/client/api/connected/GetSecurityHotspotRequestParams.java
damien-urruty-sonarsource: It is not required, indeed. Only the hotspot "key" is needed

core/src/main/java/org/sonarsource/sonarlint/core/container/connected/vulnerability/SecurityHotspotsService.java
damien-urruty-sonarsource: Same here, I'd be happy to take any input from SQ code base/team regarding correct naming. Else I will change this to hotspotKey

client-api/src/main/java/org/sonarsource/sonarlint/core/client/api/connected/GetSecurityHotspotRequestParams.java
damien-urruty-sonarsource: I can see only references to `key` in the [APIs for hotspots](https://next.sonarqube.com/sonarqube/web_api/api/hotspots?internal=true)

PR #332, opened 2020-11-03
SLI-283 Improve coverage

damien-urruty-sonarsource:


core/src/test/java/org/sonarsource/sonarlint/core/container/connected/hotspot/SecurityHotspotsServiceTest.java
```diff
@@ -0,0 +1,163 @@
...
+  @Before
+  public void setUp() {
+    service = new SecurityHotspotsService(wsClient, parser);
+  }
+
+  @Test
+  public void it_should_call_the_expected_api_endpoint_when_fetching_hotspot_details() {
+    service.fetch(new GetSecurityHotspotRequestParams("h", "p"));
+
+    verify(wsClient).get("/api/hotspots/show.protobuf?projectKey=p&hotspot=h");
```
henryju, line 67: You should verify that URL encoding is working, especially for project key with colon

henryju (changes requested):

PR #333, opened 2020-11-04
Notifications telemetry + hardening

henryju:


core/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryLocalStorageManager.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+  }
+
+  private static Gson createGson() {
+    return new GsonBuilder()
+      .registerTypeAdapter(OffsetDateTime.class, new OffsetDateTimeAdapter())
+      .create();
+  }
+
+  private TelemetryLocalStorage load() throws IOException {
```
damien-urruty-sonarsource, line 106: I am not sure I understand the difference between read and load. When to use one or the other ? Do we need both ? If so maybe we could come up with better names

damien-urruty-sonarsource (approved):
Looks good to me, it simplifies with the need to merge that is removed, so kudos :+1: 

We still have a distinction between data that is accumulated and instant data, but I think it's ok.

I made a few comments, and I still have a doubt about the names of some concepts, but I can't suggest better names

core/src/main/java/org/sonarsource/sonarlint/core/container/model/DefaultSonarQubeNotification.java
```diff
@@ -23,10 +23,10 @@
 
 import javax.annotation.concurrent.Immutable;
 
-import org.sonarsource.sonarlint.core.client.api.notifications.SonarQubeNotification;
+import org.sonarsource.sonarlint.core.client.api.notifications.ServerNotification;
 
 @Immutable
-public class DefaultSonarQubeNotification implements SonarQubeNotification {
+public class DefaultSonarQubeNotification implements ServerNotification {
```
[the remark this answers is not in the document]
henryju: Good catch !

core/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryLocalStorageManager.java
henryju: One method (`read`) is reading from the same FileChannel used later by write (in order to ensure atomicity). 
`load` is just reading without lock (mainly used to check the isEnabled flag).

Let me try to rename to make it more explicit.

damien-urruty-sonarsource (approved):

PR #334, opened 2020-11-09
Use last released artifact for sonar-ws

jblievremont:


damien-urruty-sonarsource (approved):

henryju (approved):
Out of curiosity, why do we need this patch version of sonar-ws? Any related ticket?

jblievremont:
I don't think we need the patch version, we could probably rely on the previously released `8.5.0.37579` - I only took the patch version because it's the latest release.

henryju:
OK, my bad, I missed the fact that build was failing on master

PR #343, opened 2021-01-14
Replace batch/issues by api/issues/search

henryju:
I still have to handle better the 10k limit... but I think you can still review this.


core/src/main/java/org/sonarsource/sonarlint/core/ConnectedSonarLintEngineImpl.java
```diff
@@ -237,17 +237,17 @@ private void checkUpdateStatus() {
   }
 
   @Override
-  public List<ServerIssue> downloadServerIssues(ServerConfiguration serverConfig, ProjectBinding projectBinding, String ideFilePath) {
+  public List<ServerIssue> downloadServerIssues(ServerConfiguration serverConfig, ProjectBinding projectBinding, String ideFilePath, @Nullable ProgressMonitor monitor) {
```
damien-urruty-sonarsource, line 240: I can see that you added this monitor, am I supposed to provide it when fetching issues ? Is this action supposed to be cancelable ?

damien-urruty-sonarsource (approved):
LGTM :+1:

damien-urruty-sonarsource (approved):

PR #344, opened 2021-01-14
Use different caches per job

henryju:
We don't want to share the same cache between different jobs of the pipeline, because the content will not be the same depending on the Maven goal that is executed.


jblievremont (approved):
Looks good, let's see how it behaves!

damien-urruty-sonarsource (approved):

PR #345, opened 2021-01-14
Expose server issue locations

henryju:


damien-urruty-sonarsource (commented):
Some comments related more to design than code. Most of them are not directly related to your changes, but while we are at it might be the good time to refactor further.
A problem for me is this big hierarchy starting from `IssueRangeAndMessage` that is over-complicated:
* I think `Issue` should not extend `IssueLocation`.  To me an issue is not a location, it *has* a (primary) location. Consequently, `DefaultClientIssue` could be composed of a primary location instead of extending `TextRangeLocation`.
* There are lots of interfaces that have only one implementation, and this is a code smell to me. All the implementations are just named `Default*` which shows that there is no implementation specifics. Having data classes would be more adapted to what they represent. This is true for:
  * `Issue.Flow` only has  `DefaultFlow`. I think we should remove the interface and change `DefaultFlow` to be `Flow`.
  * same remark for `ServerIssue.Flow` and `DefaultServerFlow`
  * same remark for `Issue` and `DefaultClientIssue`
  * same remark for `ServerIssue` and `DefaultServerIssue`
  * same remark for `ServerIssueLocation` and `DefaultServerLocation`
  * same remark for `IssueLocation` and `DefaultLocation`
* I wonder if we need to keep `TextRangeLocation`. It has 2 constructors used from 2 different children, so maybe it could be removed.
* I feel the name `IssueRangeAndMessage` shows that this interface groups the range and the message (and provide default implems) but does not represent a 'concept'. 
  1. What about something like `CodeAnnotation` ? Maybe we already have a term for that, I don't know if they name it somehow e.g. in analyzers.
  2. If we can't come up with a name, maybe it's because this interface does not represent a concept on its own, but is only a part of a Location. In this case we might want to keep only `ServerIssueLocation` and `IssueLocation` without a common parent.

Sorry it was long, we can talk about it if you want!

PR #346, opened 2021-01-18
move methods to core since they are used in VSCode LS also

kirill-knize-sonarsource:
equalsIgnoringTrailingSlash() method is used in VSCode language server also now


henryju (approved):
LGTM

jblievremont (approved):
LGTM :+1:

PR #347, opened 2021-01-19
Rework HTTP client

henryju:
Let the client manage the lifecycle of the OkHTTPClient


core/src/main/java/org/sonarsource/sonarlint/core/WsHelperImpl.java
```diff
@@ -80,61 +83,35 @@ static ValidationResult validateConnection(SonarLintWsClient client, @Nullable S
...
+    ServerVersionAndStatusChecker serverChecker = new ServerVersionAndStatusChecker(wsClient);
+    return getOrganization(wsClient, serverChecker, organizationKey, new ProgressWrapper(monitor));
   }
 
   @Override
-  public Optional<RemoteProject> getProject(ServerConfiguration serverConfig, String projectKey, ProgressMonitor monitor) {
-    SonarLintWsClient client = createClient(serverConfig);
-    return fetchComponent(client, projectKey)
+  public Optional<RemoteProject> getProject(ConnectedModeEndpoint endpoint, String projectKey, ProgressMonitor monitor) {
+    SonarLintWsClient wsClient = createClient(endpoint);
```
damien-urruty-sonarsource, line 107: Each time we have to create a WsClient. I wonder if it shouldn't be a constructor parameter.

core/src/main/java/org/sonarsource/sonarlint/core/container/connected/SonarLintWsClient.java
```diff
@@ -35,171 +35,146 @@
...
+    }
+    return errorMessages.stream().collect(joining(", "));
   }
 
   public Optional<String> getOrganizationKey() {
-    return Optional.ofNullable(organizationKey);
+    if (endpoint.isSonarCloud()) {
+      return Optional.ofNullable(endpoint.getOrganization());
+    }
+    return Optional.empty();
```
damien-urruty-sonarsource, line 140: As I said, this class to me is more of an helper (pagination, errors). But in the middle we have this method about organization key. Shouldn't it belong to WsHelperImpl which is higher level ?

damien-urruty-sonarsource:
As a final comment, we don't have coherence with the way we manage web services interactions. We have WsHelperImpl that groups some calls but for other things we have dedicated classes, like `*Downloader` classes, `SecurityHotspotsService`, server notifications stuff. Maybe this is a bit outside of the scope of this PR

damien-urruty-sonarsource (approved):
LGTM, tested with IJ implementation

PR #348, opened 2021-01-21
SLCORE-294 Telemetry for taint vulnerabilities

henryju:


damien-urruty-sonarsource (approved):
Looks good! Will make the change in SLI

PR #350, opened 2021-02-01
SLCORE-198 Workaround for 10k limit

henryju:
Attempt to mitigate the 10k issues limit. I have not tested end to end, only verified that RESOLVED issues are coming first on this big SonarCloud project:
https://sonarcloud.io/api/issues/search?statuses=OPEN,CONFIRMED,REOPENED,RESOLVED&types=CODE_SMELL,BUG,VULNERABILITY&componentKeys=shipilev_jdk&organization=shipilev&s=STATUS&asc=false

Not sure it is very easy to add an integration test, because it is hard to control the order of issues.


jblievremont (approved):
LGTM :+1: hoping that this sort remains stable :sweat_smile:

PR #351, opened 2021-02-02
Fetch code snippets for taint vulnerabilities

henryju:


henryju:
Everything done in #347

PR #354, opened 2021-02-22
SLCORE-301 Return code snippet for server hotspots

henryju:


core/src/main/java/org/sonarsource/sonarlint/core/serverapi/source/SourceApi.java
```diff
@@ -51,4 +56,16 @@ public SourceApi(ServerApiHelper serverApiHelper) {
     }
   }
 
+  public static String getCodeSnippet(String[] sourceCodeLines, TextRange textRange) {
```
damien-urruty-sonarsource, line 59: I would suggest to not name the method getSomething as we are in the SourceApi, even if it's static from the outside it looks like we do a GET call. Maybe it could be something like `extractCodeSnippet`. And anyway I wonder if it shouldn't be in a separate class, to keep the SourceApi class really focused on the server apis. Also if we have a dedicated (utility) class it could have 2 methods, one taking the lines array and another taking the full string and doing the split (and calling the first one). I think it makes sense in terms of responsibility

damien-urruty-sonarsource (approved):
Just an improvement suggestion, else :+1:

core/src/main/java/org/sonarsource/sonarlint/core/serverapi/source/SourceApi.java
henryju: Done!

PR #356, opened 2021-03-05
add common method for getting taint vulnerabilities to connected engine

kirill-knize-sonarsource:


henryju:
@kirill-knize-sonarsource what is the status of this PR? Should it be closed?

henryju:
Too old topic

PR #359, opened 2021-03-16
remove eclipse specific case from the core

kirill-knize-sonarsource:


henryju (approved):
Don't forget to put the ticket number at the start of the commit message

PR #365, opened 2021-04-14
Fix C analysis in QA

henryju:


damien-urruty-sonarsource (commented):
It's green again, so the fix appears to work! But I don't understand it, could you explain ?

PR #369, opened 2021-05-04
SLCORE-312 Introduce a SonarLint Plugin API

damien-urruty-sonarsource:


plugin-api/pom.xml
```diff
@@ -0,0 +1,34 @@
...
+    <artifactId>sonarlint-core-parent</artifactId>
+    <version>5.4-SNAPSHOT</version>
+  </parent>
+  <artifactId>sonarlint-plugin-api</artifactId>
+  <packaging>jar</packaging>
+  <name>SonarLint Plugin API</name>
+  <description>API used between SonarLint and analyzers</description>
+
+  <properties>
+    <jarsigner.skip>true</jarsigner.skip>
```
jblievremont, line 15: Not sure about this, this might be required e.g for Eclipse (especially since `-core` has a dependency on this module).

jblievremont (approved):
LGTM, see my comment about signing the `sonarlint-plugin-api` JAR.

PR #374, opened 2021-05-14
SLCORE-318 Let plugins query the sonarlint-plugin-api version

damien-urruty-sonarsource:


plugin-api/src/main/java/org/sonarsource/sonarlint/plugin/api/SonarLintContext.java
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * This interface is implemented by the Plugin.Context instance provided to plugins in {link {@link org.sonar.api.Plugin#define(Plugin.Context)}}
+ *
+ * The specific instance can be retrieved in the plugin definition as follows:
+ *
+ * {@code
+ * @Override
+ *   public void define(Context context) {
+ *     if (context instanceof SonarLintContext) {
```
jblievremont, line 32: I believe this code sample wouldn't work in a scanner runtime (class `SonarLintContext` won't exist) except if the class is loaded via e.g `try { class.forName(...) } ...`.

core/src/main/java/org/sonarsource/sonarlint/core/container/global/MetadataLoader.java
```diff
@@ -28,20 +28,32 @@
...
+  private static final String SONARLINT_PLUGIN_API_VERSION_FILE_PATH = "/sonarlint-api-version.txt";
 
   private MetadataLoader() {
     // only static methods
   }
 
-  public static Version loadVersion(System2 system) {
-    URL url = system.getResource(VERSION_FILE_PATH);
+  public static Version loadSonarPluginApiVersion() {
+    URL url = System2.INSTANCE.getResource(SONAR_PLUGIN_API_VERSION_FILE_PATH);
```
jblievremont, line 39: Why this change? Generally we prefer to inject and use `System2` - this is supposed to make the code more testable since it relies on a "mockable" instance instead of static stuff (and yes, in production code it is expected to use `System2.INSTANCE` when not in a container).

core/src/main/java/org/sonarsource/sonarlint/core/container/global/MetadataLoader.java
```diff
@@ -28,20 +28,32 @@
...
     } catch (IOException e) {
-      throw new IllegalStateException("Can not load " + VERSION_FILE_PATH + " from classpath ", e);
+      throw new IllegalStateException("Can not load " + SONAR_PLUGIN_API_VERSION_FILE_PATH + " from classpath ", e);
+    }
+  }
+
+  public static Version loadSonarLintPluginApiVersion() {
+    URL url = MetadataLoader.class.getResource(SONARLINT_PLUGIN_API_VERSION_FILE_PATH);
+
+    try (Scanner scanner = new Scanner(url.openStream(), StandardCharsets.UTF_8.name())) {
```
jblievremont, line 52: Nitpick: this could probably be extracted to a utility method, only the resource path changes.

jblievremont (changes requested):
Looks nice on the principle, see my few comments.

core/src/main/java/org/sonarsource/sonarlint/core/container/global/MetadataLoader.java
damien-urruty-sonarsource: In general I tend to agree, here we didn't have any test and the argument was always `System2.INSTANCE`, that's why I inlined it.

plugin-api/src/main/java/org/sonarsource/sonarlint/plugin/api/SonarLintContext.java
damien-urruty-sonarsource: Correct. Should I just keep `This interface is implemented by the Plugin.Context instance provided to plugins in {link {@link org.sonar.api.Plugin#define(Plugin.Context)}}` ?

jblievremont (approved):

PR #383, opened 2021-06-01
SLCORE-322 Support RulesDefinitionXmlLoader

henryju:
I did not add an IT, because it will be reverted soon, when we can update sonar-plugin-api to 9.0.


damien-urruty-sonarsource (approved):
:+1:

PR #387, opened 2021-06-11
SLCORE-325 Ignore remote compatibility checks for embedded plugins

henryju:


core/src/main/java/org/sonarsource/sonarlint/core/container/connected/update/PluginReferencesDownloader.java
```diff
@@ -58,17 +56,27 @@ public PluginReferences toReferences(List<SonarAnalyzer> analyzers) {
...
+      .setKey(analyzer.key());
+    if (!configuration.getEmbeddedPluginUrlsByKey().containsKey(analyzer.key())) {
+      // For embedded plugins, ignore hash and filename
+      builder.setHash(analyzer.hash())
+        .setFilename(analyzer.filename());
+    }
+    return builder.build();
+  }
+
   private boolean analyzerFilter(SonarAnalyzer analyzer) {
```
damien-urruty-sonarsource, line 76: Suggestion: rename this method to `isCompatible`

damien-urruty-sonarsource (approved):
Suggestion: there are now 3 places that check `configuration.getEmbeddedPluginUrlsByKey().containsKey(analyzer.key())` in the class, this seems like a smell to me. I wonder if we shouldn't organize the code more explicitely. The class is called PluginReferencesDownloader, so why not pass an already filtered list of `SonarAnalyzer`s to it ?

henryju:
> Suggestion: there are now 3 places that check `configuration.getEmbeddedPluginUrlsByKey().containsKey(analyzer.key())` in the class, this seems like a smell to me. I wonder if we shouldn't organize the code more explicitely. The class is called PluginReferencesDownloader, so why not pass an already filtered list of `SonarAnalyzer`s to it ?

That's true. Let's keep that in mind for the rework of connected mode. For now I don't have the bandwidth to refactor all this code.

PR #398, opened 2021-06-28
fix bug on action in connected mode

kirill-knize-sonarsource:

### DIFF
--- a/core/src/main/java/org/sonarsource/sonarlint/core/container/storage/SonarQubeActiveRulesProvider.java
+++ b/core/src/main/java/org/sonarsource/sonarlint/core/container/storage/SonarQubeActiveRulesProvider.java
@@ -95,11 +95,7 @@ public class SonarQubeActiveRulesProvider extends ProviderAdapter {
       .stream().anyMatch(extraPluginKey -> ruleLanguage.getLanguageKey().equals(extraPluginKey));
   }
 
-  public ActiveRules getActiveRules() {
-    return activeRules;
-  }
-
-  public static org.sonar.api.batch.rule.ActiveRule createNewActiveRule(ActiveRule activeRule, Sonarlint.Rules storageRules) {
+  private static org.sonar.api.batch.rule.ActiveRule createNewActiveRule(ActiveRule activeRule, Sonarlint.Rules storageRules) {
     RuleKey ruleKey = RuleKey.of(activeRule.getRepo(), activeRule.getKey());
     Sonarlint.Rules.Rule storageRule;
     try {
--- a/core/src/main/java/org/sonarsource/sonarlint/core/container/storage/SonarQubeActiveRulesProvider.java
+++ b/core/src/main/java/org/sonarsource/sonarlint/core/container/storage/SonarQubeActiveRulesProvider.java
@@ -95,7 +95,11 @@ public class SonarQubeActiveRulesProvider extends ProviderAdapter {
       .stream().anyMatch(extraPluginKey -> ruleLanguage.getLanguageKey().equals(extraPluginKey));
   }
 
-  private static org.sonar.api.batch.rule.ActiveRule createNewActiveRule(ActiveRule activeRule, Sonarlint.Rules storageRules) {
+  public ActiveRules getActiveRules() {
+    return activeRules;
+  }
+
+  public static org.sonar.api.batch.rule.ActiveRule createNewActiveRule(ActiveRule activeRule, Sonarlint.Rules storageRules) {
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/sonarsource/sonarlint/core/container/storage/StorageAnalyzer.java
```diff
@@ -75,4 +79,9 @@ public AnalysisResults analyze(ComponentContainer parent, ConnectedAnalysisConfi
     analysisContainer.execute();
     return defaultAnalysisResult;
   }
+
+  public ActiveRule getRule(String ruleKey) {
+    ActiveRules activeRules = analysisContainer.getComponentByType(ActiveRules.class);
```
henryju, line 90: NPE if no analysis has been made before.

core/src/main/java/org/sonarsource/sonarlint/core/container/storage/StorageAnalyzer.java
```diff
@@ -61,7 +65,7 @@ private void checkStatus(@Nullable String projectKey) {
   public AnalysisResults analyze(ComponentContainer parent, ConnectedAnalysisConfiguration configuration, IssueListener issueListener, ProgressWrapper progress) {
     checkStatus(configuration.projectKey());
 
-    AnalysisContainer analysisContainer = new AnalysisContainer(parent, progress);
+    analysisContainer = new AnalysisContainer(parent, progress);
```
henryju, line 74: I don't think we should do that. The `AnalysisContainer` is intended to be created/disposed for each analysis. You should not keep a reference on it.

core/src/main/java/org/sonarsource/sonarlint/core/container/storage/StorageContainer.java
```diff
@@ -204,29 +204,41 @@ public ConnectedRuleDetails getRuleDetails(String ruleKeyStr) {
...
   }
 
   public ConnectedRuleDetails getRuleDetails(String ruleKeyStr, @Nullable String projectKey) {
     ActiveRule readActiveRuleFromStorage = getHandler().readActiveRuleFromStorage(ruleKeyStr, projectKey);
-    return getRuleDetailsWithSeverity(ruleKeyStr, readActiveRuleFromStorage.getSeverity());
+    String severity;
+    if (readActiveRuleFromStorage != null) {
+      severity = readActiveRuleFromStorage.getSeverity();
+    } else {
+      StorageAnalyzer storageAnalyzer = getComponentByType(StorageAnalyzer.class);
```
henryju, line 237: I don't understand why you need to get the severity from the active rule. When the rule come from an extra plugin, there will be no way for the user to override the severity, so you can take the default severity from the rule, using `rulesFromPlugin` like you did earlier.

core/src/main/java/org/sonarsource/sonarlint/core/container/storage/StorageContainerHandler.java
```diff
@@ -107,7 +102,7 @@ public ActiveRule readActiveRuleFromStorage(String ruleKeyStr, @Nullable String
         return activeRulesFromStorage.getActiveRulesByKeyMap().get(ruleKeyStr);
       }
     }
-    throw new IllegalArgumentException("Unable to find active rule with key " + ruleKeyStr);
+    return null;
```
henryju, line 105: Then add a `@CheckForNull` to the method

henryju (changes requested):

core/src/main/java/org/sonarsource/sonarlint/core/container/storage/SonarQubeActiveRulesProvider.java
```diff
@@ -95,7 +95,11 @@ private boolean isRuleFromExtraPlugin(Language ruleLanguage, ConnectedGlobalConf
       .stream().anyMatch(extraPluginKey -> ruleLanguage.getLanguageKey().equals(extraPluginKey));
   }
 
-  private static org.sonar.api.batch.rule.ActiveRule createNewActiveRule(ActiveRule activeRule, Sonarlint.Rules storageRules) {
+  public ActiveRules getActiveRules() {
```
henryju, line 98: To be deleted

henryju (approved):
Minor cleanup then LGTM

core/src/main/java/org/sonarsource/sonarlint/core/container/storage/SonarQubeActiveRulesProvider.java
kirill-knize-sonarsource: Done

PR #400, opened 2021-07-02
SLCORE-330 Make the connection validation request asynchronous

damien-urruty-sonarsource:


jblievremont (approved):
Nice use of futures 👍🏻

PR #404, opened 2021-07-20
SLCORE-335 Attempt to fix dead lock in ProgressReport

henryju:


core/src/main/java/org/sonarsource/sonarlint/core/util/ProgressReport.java
```diff
@@ -40,7 +41,7 @@ public ProgressReport(String threadName, long period) {
 
   @Override
   public void run() {
-    while (!Thread.interrupted()) {
+    while (!stop) {
```
henryju, line 44: My hypothesis is that the interrupt flag has been somehow lost. So I stopped relying on it for collaborative cancellation. Instead I introduced a `stop` volatile variable.

core/src/main/java/org/sonarsource/sonarlint/core/util/ProgressReport.java
```diff
@@ -64,19 +65,17 @@ public void message(Supplier<String> messageSupplier) {
 
   public void stop(@Nullable String stopMessage) {
     this.stopMessage = stopMessage;
+    this.stop = true;
     thread.interrupt();
     try {
-      thread.join();
+      thread.join(1000);
```
henryju, line 71: Just as an extra safety net, don't wait forever.

core/src/main/java/org/sonarsource/sonarlint/core/util/ProgressReport.java
```diff
@@ -64,19 +65,17 @@ public void message(Supplier<String> messageSupplier) {
...
       // Ignore
     }
   }
 
   private static void log(String message) {
-    synchronized (LOG) {
-      LOG.info(message);
-      LOG.notifyAll();
-    }
+    LOG.info(message);
```
henryju, line 78: I could not remember/find why this synchronized block was there, so I decided to remove it...

jblievremont: Probably because `LOG` can be shared between threads and `LOG.info` might not be thread-safe? E.g if 2 `ProgressReport`s try to log from separate threads.

jblievremont (approved):
I would probably keep the lock on the log, else LGTM.

core/src/main/java/org/sonarsource/sonarlint/core/util/ProgressReport.java
henryju: I would be very surprised if LOG is not thread safe. We are often doing concurrent operations (analysis), and so they may log concurrently. This is the only place I know where we are locking like this. Git history shows this was there from the start. I suggest to keep removing it, and if it creates a regression, we will add a test to document why it is there.

PR #405, opened 2021-07-21
SLCORE-336 Let analyzers contribute quick fixes for issues

jblievremont:


core/src/main/java/org/sonarsource/sonarlint/core/analyzer/issue/DefaultClientIssue.java
```diff
@@ -85,6 +88,10 @@ public ClientInputFile getInputFile() {
     return flows;
   }
 
+  public List<QuickFix> quickFixes() {
```
damien-urruty-sonarsource, line 91: As discussed we can also add the method in the `Issue` interface and mark the override here

core/src/main/java/org/sonarsource/sonarlint/core/analyzer/sensor/SonarLintSensorStorage.java
```diff
@@ -96,9 +98,11 @@ public void store(Issue issue) {
     String type = rule.type().name();
 
     List<org.sonarsource.sonarlint.core.client.api.common.analysis.Issue.Flow> flows = mapFlows(issue.flows());
+    List<QuickFix> quickFixes = issue instanceof DefaultSonarLintIssue ?
```
damien-urruty-sonarsource, line 101: The issue we get in this method should always be a `DefaultSonarLintIssue` so we could cast early

plugin-api/src/main/java/org/sonarsource/sonarlint/plugin/api/issue/NewInputFileEdit.java
```diff
@@ -0,0 +1,42 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.sonarlint.plugin.api.issue;
+
+import org.sonar.api.batch.fs.InputFile;
+
+/**
+ * Describe a file edit for a {@link NewQuickFix} as a collection of {@link NewTextEdit}s on a given {@link InputFile}.
+ * Text edits are applied in the order they are added, insofar that their ranges do not overlap.
```
damien-urruty-sonarsource, line 26: In this case, would we throw ? Ignore the addition ? If we talk about this corner case, Iwe should be more precise about how we handle it

plugin-api/src/main/java/org/sonarsource/sonarlint/plugin/api/issue/NewQuickFix.java
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.sonarlint.plugin.api.issue;
+
+/**
+ * Describe a quick fix for a {@link NewSonarLintIssue}, with a description and a collection of {@link NewInputFileEdit}.
+ * Input file edits will be applied in the order they are added, insofar that they are compatible with one another.
```
damien-urruty-sonarsource, line 24: Do we need to restate this in this class ? Or is it about order of QuickFixes ?

damien-urruty-sonarsource (approved):
Small comments, and maybe a misunderstanding: I thought the final decision was to keep add and new methods, to be more coherent with sonar-api in case one day we need to re-integrate it.

core/src/main/java/org/sonarsource/sonarlint/core/client/api/common/InputFileEdit.java
```diff
@@ -0,0 +1,30 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.sonarlint.core.client.api.common;
+
+import java.util.List;
+import org.sonar.api.batch.fs.InputFile;
+
+public interface InputFileEdit {
```
damien-urruty-sonarsource, line 25: This one could be `ClientInputFileEdit`, and we should return a `ClientInputFile` from `target()`. With the current SLI implementation it's not a problem because we don't check the file (we assume all the quick fixes target the analyzed file) but we might want to check this later

plugin-api/src/main/java/org/sonarsource/sonarlint/plugin/api/issue/NewQuickFix.java
jblievremont: Yes, I believe it deserves to be explicit: input file edits of a given quick fix are applied in the order they are declared, and text edits of a given input file edit are applied in the order they are declared.

plugin-api/src/main/java/org/sonarsource/sonarlint/plugin/api/issue/NewInputFileEdit.java
jblievremont: I believe that the check should be made only when trying to apply a fix: doing it in advance would probably hurt performance, and we wouldn't have knowledge about any change that might have happened concurrently in the files.

PR #412, opened 2021-09-14
SLCORE-343 Don't expose sonar-plugin-api TextRange to IDEs

henryju:


henryju:
@jblievremont FYI

damien-urruty-sonarsource (approved):
Nice refactoring at the same time :+1: 
Small comment, it seems a method is not covered by tests

PR #419, opened 2021-12-01
SLCORE-346 add ServerApi method to get branches for project

kirill-knize-sonarsource:


core/src/main/java/org/sonarsource/sonarlint/core/serverapi/branches/ProjectBranchesApi.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import org.sonar.api.utils.log.Logger;
+import org.sonar.api.utils.log.Loggers;
+import org.sonarsource.sonarlint.core.serverapi.ServerApiHelper;
+
+public class ProjectBranchesApi {
+  private static final Logger LOG = Loggers.get(ProjectBranchesApi.class);
+
+  private static final String LIST_ALL_PROJECT_BRANCHES_URL = "/api/project_branches/list";
```
jblievremont, line 34: Nitpick: I think the leading `/` should be removed for consistency.

core/src/test/java/org/sonarsource/sonarlint/core/container/connected/update/ProjectBranchesDownloaderTest.java
```diff
@@ -0,0 +1,91 @@
...
+        "      \"type\": \"BRANCH\",\n" +
+        "      \"status\": {\n" +
+        "        \"qualityGateStatus\": \"OK\"\n" +
+        "      },\n" +
+        "      \"analysisDate\": \"2017-04-03T13:37:00+0100\",\n" +
+        "      \"excludedFromPurge\": false\n" +
+        "    },\n" +
+        "    {\n" +
+        "      \"name\": \"master\",\n" +
+        "      \"isMain\": true,\n" +
```
jblievremont, line 61: Note: I wonder if we'll need this attribute, e.g to decide what the "fallback" should be when we cannot find the appropriate branch.

jblievremont (approved):
LGTM for a first baby step, we'll see with the actual algorithm if we need more data from the WS 😄

core/src/test/java/org/sonarsource/sonarlint/core/container/connected/update/ProjectBranchesDownloaderTest.java
kirill-knize-sonarsource: Yep. Me too. I decided not to extract all data we will possibly need to avoid redundant code and get only the name for now. We can extract more when we will know exactly what we need.

core/src/main/java/org/sonarsource/sonarlint/core/serverapi/branches/ProjectBranchesApi.java
kirill-knize-sonarsource: I used QualityProfileApi, SettingsApi and HotspotApi. All of them have leading / in paths.

PR #422, opened 2021-12-07
Simplify IT matrix

henryju:
SonarQube embeds all plugins since 7.9, so we can rely on bundled plugins.


damien-urruty-sonarsource (approved):
Nice simplification :+1:

PR #423, opened 2021-12-07
Rework part 2

henryju:


core/src/main/java/org/sonarsource/sonarlint/core/plugin/Slf4jBridgeClassLoader.java
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  @Override
+  protected Class<?> findClass(final String name) throws ClassNotFoundException {
+    if (name.startsWith("org.slf4j")) {
+      String path = name.replace('.', '/').concat(".clazz");
+      String classContentPath = "slf4j-sonar-log/" + path;
+      try (InputStream is = getParent().getResourceAsStream(classContentPath)) {
+        if (is == null) {
+          throw new IllegalStateException("Unable to find resource " + classContentPath);
```
damien-urruty-sonarsource, line 40: We could throw ClassNotFoundException, as the original method advertises.

damien-urruty-sonarsource (approved):
LGTM, minor feedback

core/src/main/java/org/sonarsource/sonarlint/core/plugin/Slf4jBridgeClassLoader.java
henryju: :+1:

PR #442, opened 2022-01-03
SLCORE-350 Support CSV encoded multi-valued properties

henryju:


analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/sonarapi/MapSettings.java
```diff
@@ -265,9 +265,7 @@ public Double getDouble(String key) {
         return ArrayUtils.EMPTY_STRING_ARRAY;
       }
 
-      return Arrays.stream(value.split(",", -1)).map(String::trim)
-        .map(s -> s.replace("%2C", ","))
```
damien-urruty-sonarsource, line 269: We were doing some kind of URL decode here, and I think we don't do it anymore, was that important ?

commons/src/test/java/org/sonarsource/sonarlint/core/commons/log/SonarLintLogTester.java
```diff
@@ -65,6 +65,15 @@
 
   private final Queue<String> logs = new ConcurrentLinkedQueue<>();
   private final Map<Level, Queue<String>> logsByLevel = new ConcurrentHashMap<>();
+  private final boolean writeToStdOut;
+
+  public SonarLintLogTester(boolean writeToStdOut) {
```
damien-urruty-sonarsource, line 70: I don't see this constructor invoked, you want to keep it to ease debugging later on local machine ?

damien-urruty-sonarsource (approved):
Some questions, and also you could replace some variable declarations with `var`s.

commons/src/test/java/org/sonarsource/sonarlint/core/commons/log/SonarLintLogTester.java
henryju: Yeah, I used it to investigate a test failure.

analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/sonarapi/MapSettings.java
henryju: AFAIK this is a remain of a very very old bug we had in the UI. I don't think we need to keep that anymore.

damien-urruty-sonarsource (approved):

PR #446, opened 2022-01-07
Clean-up old plugins at each start

damien-urruty-sonarsource:


core/src/main/java/org/sonarsource/sonarlint/core/container/storage/ProjectStoragePaths.java
```diff
@@ -31,7 +31,7 @@
   /**
    * Version of the storage. This should be incremented each time an incompatible change is made to the storage.
    */
-  public static final String STORAGE_VERSION = "3";
+  public static final String STORAGE_VERSION = "3 ";
```
henryju, line 34: Is this extra space intended?

henryju (changes requested):

core/src/main/java/org/sonarsource/sonarlint/core/container/storage/ProjectStoragePaths.java
damien-urruty-sonarsource: Of course no, I don't know how it ended up here :disappointed:

henryju (approved):

PR #451, opened 2022-01-11
SLCORE-256 Avoid crashing when a plugin has class/component loading issues

damien-urruty-sonarsource:


plugin-commons/src/main/java/org/sonarsource/sonarlint/core/plugin/commons/loading/PluginClassloaderFactory.java
```diff
@@ -90,7 +92,7 @@ private static void exportResources(PluginClassLoaderDef def, ClassloaderBuilder
     for (PluginClassLoaderDef def : defs) {
       var classloader = classloadersByBasePluginKey.get(def.getBasePluginKey());
       if (classloader == null) {
-        throw new IllegalStateException(String.format("Fail to create classloader for plugin [%s]", def.getBasePluginKey()));
+        LOG.error("Fail to create classloader for plugin '{}'", def.getBasePluginKey());
       }
       result.put(def, classloader);
```
henryju, line 97: So now the Map may contain a definition with a `null` classloader. It will fail badly later.

plugin-commons/src/test/java/org/sonarsource/sonarlint/core/plugin/commons/loading/PluginClassloaderFactoryTest.java
```diff
@@ -32,7 +31,7 @@
 import static java.util.Arrays.asList;
 import static org.assertj.core.api.Assertions.assertThat;
 
-public class PluginClassloaderFactoryTest {
+class PluginClassloaderFactoryTest {
```
henryju, line 34:
```suggestion
class PluginClassloaderFactoryTests {
```

henryju (changes requested):
I think there is a NPE

plugin-commons/src/test/java/org/sonarsource/sonarlint/core/plugin/commons/loading/PluginClassloaderFactoryTest.java
damien-urruty-sonarsource: We sometimes use plural sometimes not, what should be the rule ? Do you have any recommended naming convention in mind ?

plugin-commons/src/main/java/org/sonarsource/sonarlint/core/plugin/commons/loading/PluginClassloaderFactory.java
damien-urruty-sonarsource: Good catch! I should put in the map only if classloader is not null.
Now looking at the code, I am not even sure that classloader could be null at any time here.

plugin-commons/src/test/java/org/sonarsource/sonarlint/core/plugin/commons/loading/PluginClassloaderFactoryTest.java
henryju: I think JUnit 5 changed to use plural and personally I find it better.
So the rule is:
* JUnit4 = singular: https://github.com/junit-team/junit4/wiki/Getting-started
* JUnit 5 = plural: https://junit.org/junit5/docs/current/user-guide/#writing-tests

henryju (approved):

PR #455, opened 2022-01-17
SLCORE-353 Restore support of markdown rule description

henryju:


rule-extractor/src/main/java/org/sonar/channel/ChannelDispatcher.java
```diff
@@ -0,0 +1,142 @@
...
+ */
+package org.sonar.channel;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import org.sonarsource.sonarlint.core.commons.log.SonarLintLogger;
+
+/**
+ * Copy from sonar-channel, but replaing the slf4j logger by SonarLint one
```
damien-urruty-sonarsource, line 28:
```suggestion
 * Copy from sonar-channel, but replacing the slf4j logger by SonarLint one
```

damien-urruty-sonarsource (approved):
:+1:

PR #460, opened 2022-02-01
Do not log interrupt error when analysis engine is intentionally stopped

damien-urruty-sonarsource:


henryju (commented):
Could you add a medium test to ensure logs are clean when stopping the engine?

analysis-engine/src/test/java/org/sonarsource/sonarlint/core/analysis/mediumtests/AnalysisEngineMediumTests.java
```diff
@@ -201,6 +207,18 @@ void should_interrupt_executing_thread_when_stopping() {
     assertThat(futureLongCommand).isCompletedWithValue("INTERRUPTED");
   }
 
+  @Test
+  void should_not_log_any_error_when_stopping() {
```
henryju, line 211: Is the test failing if you remove the fix?

henryju (approved):

analysis-engine/src/test/java/org/sonarsource/sonarlint/core/analysis/mediumtests/AnalysisEngineMediumTests.java
damien-urruty-sonarsource: I checked of course :wink:

PR #461, opened 2022-02-02
SLCORE-357 add vcs integration module

kirill-knize-sonarsource:


pom.xml
```diff
@@ -36,6 +36,7 @@
     <module>analysis-engine</module>
     <module>osgi</module>
     <module>report-aggregate</module>
+    <module>sonarlint-vcs</module>
```
henryju, line 39: If you look at other modules, we don't prefix the folder name with "sonarlint-"

sonarlint-vcs/pom.xml
```diff
@@ -0,0 +1,55 @@
...
+    <artifactId>sonarlint-core-parent</artifactId>
+    <version>8.1-SNAPSHOT</version>
+  </parent>
+  <artifactId>sonarlint-vcs</artifactId>
+  <name>SonarLint Version Control System</name>
+  <description>Manage VCS integration</description>
+
+
+  <properties>
+      <maven.compiler.source>11</maven.compiler.source>
```
henryju, line 18: Should be removed, this should come from the parent pom

sonarlint-vcs/pom.xml
```diff
@@ -0,0 +1,55 @@
...
+      <maven.compiler.target>11</maven.compiler.target>
```
henryju, line 19: Same

sonarlint-vcs/src/main/java/org/sonarsource/sonarlint/core/vcs/GitUtils.java
```diff
@@ -0,0 +1,136 @@
...
+      LOG.error("Couldn't fetch commits for branch " + branchName, e);
+    }
+    return commitNames;
+  }
+
+  @CheckForNull
+  public static Git getGitForDir(URI projectDirUri) {
+    try {
+      var builder = new FileRepositoryBuilder();
+      var gitUri = new URI(projectDirUri + "/.git");
```
henryju, line 77: This look wrong. URI concatenation is very fragile. I think it doesn't make sense for this method to accept a URI for projectDirUri. It should be a Path.

sonarlint-vcs/src/main/java/org/sonarsource/sonarlint/core/vcs/GitUtils.java
```diff
@@ -0,0 +1,136 @@
...
+      var repository = builder.setGitDir(new File(gitUri)).setMustExist(true).build();
+      return new Git(repository);
+    } catch (IOException | URISyntaxException e) {
+      LOG.error("Couldn't access repository for path " + projectDirUri.getPath());
```
henryju, line 81: Preserve the original exception?

sonarlint-vcs/src/main/java/org/sonarsource/sonarlint/core/vcs/GitUtils.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+    return mainBranchName;
+  }
+
+  public static Map<String, List<String>> buildCommitsCache(Git git) {
+    List<Ref> refs;
+    try {
+      refs = git.getRepository().getRefDatabase().getRefs();
+    } catch (IOException e) {
+      LOG.error("Unable to build commits " + e);
```
henryju, line 120: Better to pass the original exception as second argument, to preserve the stacktrace

sonarlint-vcs/src/main/java/org/sonarsource/sonarlint/core/vcs/GitUtils.java
```diff
@@ -0,0 +1,136 @@
...
+      var repository = builder.setGitDir(new File(gitUri)).setMustExist(true).build();
+      return new Git(repository);
+    } catch (IOException | URISyntaxException e) {
+      LOG.error("Couldn't access repository for path " + projectDirUri.getPath());
+    }
+    return null;
+  }
+
+
+  public static Optional<String> electSQBranchForLocalBranch(String branchName, Git git, Set<ServerBranch> serverCandidates) {
```
henryju, line 87: Since the have only one branch that could be a main branch, maybe it would be simpler to pass 2 arguments:
* `Set<String> serverBranchCandidates`
* `String serverMainBranch`
This way you could remove the `ServerBranch` pojo

sonarlint-vcs/src/test/java/org/sonarsource/sonarlint/core/vcs/GitUtilsTest.java
```diff
@@ -0,0 +1,212 @@
...
+    Optional<String> branch = GitUtils.electSQBranchForLocalBranch("foo", git,
+      Set.of(new ServerBranch("foo", false),
+        new ServerBranch("bar", false),
+        new ServerBranch("master", true)));
+
+    assertThat(branch).isEmpty();
+  }
+
+  @Test
+  void shouldReturnEmptyOptionalIfNoMainBranch(@TempDir File projectDir) throws IOException, URISyntaxException {
```
henryju, line 147: Is it a real use case? Can a SQ/SC project have no main branch?

henryju (changes requested):
@kirill-knize-sonarsource May you process my feedback and then I would like to have a second look.

henryju (changes requested):

henryju (approved):

PR #471, opened 2022-03-23
SLCORE-365 Handle deprecated rule keys in connected mode

henryju:


core/src/test/java/org/sonarsource/sonarlint/core/mediumtest/fixtures/ProjectStorageFixture.java
```diff
@@ -113,18 +118,27 @@ public RuleSetBuilder(String languageKey) {
...
     }
 
     private static class ActiveRule {
       private final String ruleKey;
       private final String severity;
+      private final String templateKey;
+      private final Map<String, String> params;
 
-      private ActiveRule(String ruleKey, String severity) {
+      private ActiveRule(String ruleKey, String severity, @Nullable String templateKey, Map<String, String> params) {
```
jblievremont, line 137: Nitpick: I'd add shortcut constructors to avoid passing `null` and and empty map on 'simple' cases.

jblievremont (approved):
LGTM 👍🏻

core/src/test/java/org/sonarsource/sonarlint/core/mediumtest/fixtures/ProjectStorageFixture.java
henryju: There are builders to construct instances of this class so I don't think it is needed.

PR #475, opened 2022-04-08
Hardening

henryju:


damien-urruty-sonarsource (approved):
LGTM. Could you add the reason why we need to export Gson in the SLCORE-367 ticket ?

PR #476, opened 2022-04-19
SLCORE-369 Remove server side module support code

henryju:


core/src/main/java/org/sonarsource/sonarlint/core/container/connected/update/IssueStorePaths.java
```diff
@@ -33,61 +36,35 @@
...
-    }
-    var localPrefixLen = projectBinding.idePathPrefix().length();
-    if (localPrefixLen > 0) {
-      localPrefixLen++;
-    }
-    var sqPathPrefix = projectBinding.sqPathPrefix();
-    if (!sqPathPrefix.isEmpty()) {
-      sqPathPrefix = sqPathPrefix + "/";
+  public String idePathToSqPath(ProjectBinding projectBinding, String ideFilePathStr) {
+    var ideFilePath = Paths.get(ideFilePathStr);
```
damien-urruty-sonarsource, line 51: As you changed the implementation to use `Paths.get()`, I wondered if we could get an exception at some point. But I think all the provided parameters are valid relative paths.

damien-urruty-sonarsource (approved):
LGTM :+1: 
Lots of deleted code which is always good :smile:

core/src/main/java/org/sonarsource/sonarlint/core/container/connected/update/IssueStorePaths.java
henryju: We are already using `Path` in the prefix computation algorithm, so I think it is pretty safe.

PR #477, opened 2022-04-26
Add debug traces for server events

damien-urruty-sonarsource:


damien-urruty-sonarsource:
Integrated as part of #493

PR #478, opened 2022-04-26
SLCORE-370 Update to sonar-plugin-api 9.5

henryju:


analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/sonarapi/DefaultSensorContext.java
```diff
@@ -188,4 +190,24 @@ private static UnsupportedOperationException unsupported() {
...
+  }
+
+  @Override
+  public boolean isCacheEnabled() {
+    return false;
+  }
+
+  @Override
+  public ReadCache previousCache() {
+    throw unsupported();
```
damien-urruty-sonarsource, line 205: Will this be conditionally called on the analyzers side ? Else we could have a no-op read and write cache

pom.xml
```diff
@@ -105,7 +106,13 @@
...
         <groupId>org.mockito</groupId>
         <artifactId>mockito-core</artifactId>
-        <version>4.4.0</version>
+        <version>${mockito.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.mockito</groupId>
+        <artifactId>mockito-junit-jupiter</artifactId>
+        <version>${mockito.version}</version>
+        <scope>test</scope>
```
damien-urruty-sonarsource, line 115: Could you explain why you had to add this ?

analysis-engine/pom.xml
```diff
@@ -57,6 +57,11 @@
       <artifactId>mockito-core</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-junit-jupiter</artifactId>
+      <scope>test</scope>
+    </dependency>
```
damien-urruty-sonarsource, line 64: Same here

analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/sonarapi/DefaultSensorContext.java
henryju: Yeah, normally they should call it only when `isCacheEnabled` returns `true`

pom.xml
henryju: I will extract the change in a separate commit to make it more obvious

damien-urruty-sonarsource (approved):
The <mockito.version> property should also be moved to this new commit then.
It's clearer with this split, thanks!

PR #481, opened 2022-05-03
SLCORE-373 Collect platform and architecture with telemetry

damien-urruty-sonarsource:


henryju (approved):

jblievremont (approved):
LGTM 👍🏻 from a code point of view; IMHO we also need to ping the Data team about those changes so that they update the "lineage" document

PR #488, opened 2022-06-07
Rework taint vulnerabilities loading

henryju:


server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/storage/XodusServerIssueStore.java
```diff
@@ -82,22 +86,47 @@ public Optional<ServerIssue> getByKey(String issueKey) {
       .map(XodusServerIssueStore::adapt));
   }
 
+  public Optional<ServerTaintIssue> getTaintByKey(String issueKey) {
+    return entityStore.computeInTransaction(txn -> {
+      final Entity storedIssue = txn.find(TAINT_ISSUE_ENTITY_TYPE, KEY_PROPERTY_NAME, issueKey).getFirst();
```
damien-urruty-sonarsource, line 91: Here you should use `findUnique` because the find is case-insensitive.

damien-urruty-sonarsource (approved):
LGTM, I find that the factorization in the store between taint and issues brings a bit more complexity but it's ok

core/src/main/java/org/sonarsource/sonarlint/core/tracking/IssueTrackable.java
```diff
@@ -34,18 +34,19 @@ public class IssueTrackable implements Trackable<Issue> {
   private final Integer lineHash;
 
   public IssueTrackable(Issue issue) {
-    this(issue, null, null, null);
+    this(issue, null, null);
   }
 
-  public IssueTrackable(Issue issue, @Nullable ServerIssue.TextRange textRange, @Nullable String textRangeContent,
+  public IssueTrackable(Issue issue, @Nullable String textRangeContent,
```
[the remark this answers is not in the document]
henryju: I will rework issue tracking stuff later

server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/storage/XodusServerIssueStore.java
```diff
@@ -37,20 +38,23 @@
...
 import org.sonarsource.sonarlint.core.serverconnection.ServerIssue;
+import org.sonarsource.sonarlint.core.serverconnection.ServerTaintIssue;
 
 import static java.util.Objects.requireNonNull;
 
 public class XodusServerIssueStore implements ServerIssueStore {
   private static final String FILE_NAME = "sonarlint.db";
 
   private static final String ISSUE_ENTITY_TYPE = "Issue";
+  private static final String TAINT_ISSUE_ENTITY_TYPE = "Issue";
```
damien-urruty-sonarsource, line 49: Is it normal that we use the same entity type value for both kinds of issues ? If so we could use a single constant, e.g. `ISSUE_COMMON_ENTITY_TYPE`

server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/IssueDownloader.java
```diff
@@ -21,108 +21,147 @@
...
-   * @param branchName name of the branch. If null - issues will be downloaded only for the main branch.
+   * @param branchName name of the branch.
    * @return Iterator of issues. It can be empty but never null.
    */
-  public List<ServerIssue> download(ServerApiHelper serverApiHelper, String key, boolean fetchTaintVulnerabilities,
-    @Nullable String branchName,
-    ProgressMonitor progress) {
-    var issueApi = new ServerApi(serverApiHelper).issue();
+  public List<ServerIssue> download(ServerApiHelper serverApiHelper, String key, String branchName, boolean isSonarCloud,
+    Version serverVersion, ProgressMonitor progress) {
```
damien-urruty-sonarsource, line 63: The progress monitor is not used in this method, should we use it or remove it ?

server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/storage/XodusServerIssueStore.java
henryju: Good catch, it should be a different value

damien-urruty-sonarsource: Because we have different links so we are able to differentiate them

PR #489, opened 2022-06-08
SLCORE-372 Make server issue store branch aware

damien-urruty-sonarsource:


server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/storage/XodusServerIssueStore.java
```diff
@@ -266,13 +274,26 @@ private static Entity getOrCreateProject(String projectKey, StoreTransaction txn
       });
   }
 
-  private static Entity getOrCreateFile(Entity projectEntity, String filePath, StoreTransaction txn) {
-    var fileIterable = projectEntity.getLinks(PROJECT_TO_FILES_LINK_NAME).intersect(findAll(txn, FILE_ENTITY_TYPE, PATH_PROPERTY_NAME, filePath));
+  private static Entity getOrCreateBranch(Entity projectEntity, String branchName, StoreTransaction txn) {
+    var branchIterable = projectEntity.getLinks(PROJECT_TO_BRANCHES_LINK_NAME)
```
henryju, line 278: Might be more efficient to directly iterate over the project branches, instead of intersection with searching for all branches with given name. I expect many projects will have a "master" branch.

henryju (approved):
LGTM

server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/storage/XodusServerIssueStore.java
damien-urruty-sonarsource: I suggest we wait a bit to see if we want to optimize xodus code. I will merge it as is

PR #494, opened 2022-06-14
SLCORE-387 Pull only issues of enabled languages

damien-urruty-sonarsource:


henryju (approved):
LGTM, but I think we should add an IT to verify that the parameter is really taken into account.

PR #500, opened 2022-07-19
Synchronize taint using new WS

henryju:


server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/storage/XodusServerIssueStore.java
```diff
@@ -228,17 +226,47 @@ public void mergeIssues(String projectKey, String branchName, List<ServerIssue>
...
       .flatMap(branches -> findUnique(txn, BRANCH_ENTITY_TYPE, NAME_PROPERTY_NAME, branchName))
       .map(branch -> {
-        var lastSyncMs = (Long) branch.getProperty(LAST_SYNC_PROPERTY_NAME);
+        var lastSyncMs = (Long) branch.getProperty(LAST_ISSUE_SYNC_PROPERTY_NAME);
+        if (lastSyncMs != null) {
+          return Optional.of(Instant.ofEpochMilli(lastSyncMs));
+        } else {
+          return Optional.<Instant>empty();
+        }
+      }).orElse(Optional.<Instant>empty()));
```
damien-urruty-sonarsource, line 260: I think we could avoid having those 2 Optional.empty, maybe something like (to be tested):

```
.map(branch -> branch.getProperty(LAST_ISSUE_SYNC_PROPERTY_NAME)
.map(lastSyncMs -> Instant.ofEpochMilli((Long) lastSyncMs)
```
same for the method below

server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/storage/XodusServerIssueStore.java
```diff
@@ -457,6 +484,19 @@ private static void remove(String issueKey, @NotNull StoreTransaction txn) {
       });
   }
 
+  private static void removeTaint(String issueKey, @NotNull StoreTransaction txn) {
+    findUnique(txn, ISSUE_ENTITY_TYPE, KEY_PROPERTY_NAME, issueKey)
```
damien-urruty-sonarsource, line 488: Shouldn't it be TAINT_ISSUE_ENTITY_TYPE ?

damien-urruty-sonarsource (approved):
Small copy/paste error I think, and a suggestion, else LGTM!

server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/storage/XodusServerIssueStore.java
henryju: Good catch!

damien-urruty-sonarsource (approved):
LGTM, good work for the enums!

PR #507, opened 2022-09-14
SLCORE-401 Throw an exception when server branches storage does not exist

damien-urruty-sonarsource:


jblievremont (approved):
LGTM 👍🏻 and tested with SLLS 2.10 / SLVSCODE 3.10

PR #510, opened 2022-09-26
"Generate token" server path provider

kirill-knize-sonarsource:


server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/ServerPathProvider.java
```diff
@@ -0,0 +1,54 @@
...
+import org.sonarsource.sonarlint.core.commons.http.HttpClient;
+
+public class ServerPathProvider {
+
+  private ServerPathProvider() {
+  }
+
+  private static final String MIN_SQ_VERSION = "9.7";
+
+  public static String getServerPath(EndpointParams endpoint, HttpClient client, int port, String ideName) throws ExecutionException, InterruptedException {
```
henryju, line 35: Maybe give a better name for the method? Like `getServerUrlForTokenGeneration`?

server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/ServerPathProvider.java
```diff
@@ -0,0 +1,54 @@
...
+  static String buildServerPath(String baseUrl, String serverVersionStr, int port, String ideName) {
+    var minVersion = Version.create(MIN_SQ_VERSION);
+    var serverVersion = Version.create(serverVersionStr);
+    var path = new StringBuilder(baseUrl);
+    if (serverVersion.satisfiesMinRequirement(minVersion)) {
+      path.append("/sonarlint/auth");
+    } else {
+      path.append("/account/security");
+    }
+    path.append("?port=").append(port).append("&ideName=").append(ideName);
```
henryju, line 50: This looks strange, I don't think `/account/security` is using those parameters.

server-connection/src/test/java/org/sonarsource/sonarlint/core/serverconnection/ServerPathProviderTests.java
```diff
@@ -0,0 +1,50 @@
...
+    var serverPath = ServerPathProvider.buildServerPath("baseUrl", "9.8", 1234, "My IDE");
+
+    assertThat(serverPath).isEqualTo("baseUrl/sonarlint/auth?port=1234&ideName=My IDE");
+  }
+
+  @Test
+  void profile_token_generation_path_for_version_lower_than_9_7() {
+    var serverPath = ServerPathProvider.buildServerPath("baseUrl", "9.6", 1234, "My IDE");
+
+    assertThat(serverPath).isEqualTo("baseUrl/account/security?port=1234&ideName=My IDE");
```
henryju, line 47: This assertion looks wrong. I would expect :
```suggestion
    assertThat(serverPath).isEqualTo("baseUrl/account/security");
```

server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/ServerPathProvider.java
henryju, line 50: You should encode ideName, since it may contain invalid characters.

server-connection/src/test/java/org/sonarsource/sonarlint/core/serverconnection/ServerPathProviderTests.java
```diff
@@ -0,0 +1,50 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+class ServerPathProviderTests {
+
+
+  @Test
+  void new_auth_path_for_9_7_version() {
+    var serverPath = ServerPathProvider.buildServerPath("baseUrl", "9.7", 1234, "My IDE");
+
+    assertThat(serverPath).isEqualTo("baseUrl/sonarlint/auth?port=1234&ideName=My IDE");
```
henryju, line 33:
```suggestion
    assertThat(serverPath).isEqualTo("baseUrl/sonarlint/auth?port=1234&ideName=My%20IDE");
```

henryju (changes requested):
What about SonarCloud? The version check is not enough, since we don't really know what is the version returned by SonarCloud those days.
You have no other option that adding a new boolean parameter to the method, since only the IDE knows at this point if the user is trying to connect to SQ or SC.

server-connection/src/test/java/org/sonarsource/sonarlint/core/serverconnection/ServerPathProviderTests.java
```diff
@@ -0,0 +1,59 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class ServerPathProviderTests {
+
+
+  @Test
+  void new_auth_path_for_sonar_cloud() {
```
henryju, line 30: As discussed, I think SonarCloud will not implement the new WS soon, so you should better make it behave like an old SQ.

henryju (changes requested):

henryju (commented):
You could also add an IT in the SC test suite.

server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/ServerPathProvider.java
```diff
@@ -34,10 +34,10 @@ private ServerPathProvider() {
   private static final String MIN_SQ_VERSION = "9.7";
 
   public static String getServerUrlForTokenGeneration(EndpointParams endpoint, HttpClient client,
-    int port, String ideName, boolean isSonarCloud) throws ExecutionException, InterruptedException {
+    int port, String ideName) throws ExecutionException, InterruptedException {
     var serverApi = new ServerApi(endpoint, client);
     var systemInfo = serverApi.system().getStatus().get();
-    return buildServerPath(endpoint.getBaseUrl(), systemInfo.getVersion(), port, ideName, isSonarCloud);
+    return buildServerPath(endpoint.getBaseUrl(), systemInfo.getVersion(), port, ideName, endpoint.isSonarCloud());
```
henryju, line 40: :+1:

henryju (approved):

PR #514, opened 2022-10-12
SLCORE-410 Make port optional in token generation path provider

kirill-knize-sonarsource:


server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/ServerPathProvider.java
```diff
@@ -33,23 +34,35 @@ private ServerPathProvider() {
...
-  public static String getServerUrlForTokenGeneration(EndpointParams endpoint, HttpClient client,
-    int port, String ideName) throws ExecutionException, InterruptedException {
+  public static CompletableFuture<String> getServerUrlForTokenGeneration(EndpointParams endpoint, HttpClient client,
+    int port, String ideName) {
     var serverApi = new ServerApi(endpoint, client);
-    var systemInfo = serverApi.system().getStatus().get();
-    return buildServerPath(endpoint.getBaseUrl(), systemInfo.getVersion(), port, ideName, endpoint.isSonarCloud());
+    CompletableFuture<ServerInfo> systemInfoFuture;
+    systemInfoFuture = serverApi.system().getStatus();
+    return systemInfoFuture.thenApply(systemInfo ->
```
damien-urruty-sonarsource, line 42: I think you could inline systemInfoFuture completely here

server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/ServerPathProvider.java
```diff
@@ -33,23 +34,35 @@ private ServerPathProvider() {
...
       path.append("/account/security");
     } else {
-      path.append("/sonarlint/auth").append("?port=").append(port).append("&ideName=").append(UrlUtils.urlEncode(ideName));
+      path.append("/sonarlint/auth").append("?ideName=").append(UrlUtils.urlEncode(ideName)).append(portParameter);
     }
     return path.toString();
   }
 
+  private static String getPortParameter(int port) {
+    if (port == -1) {
```
damien-urruty-sonarsource, line 62: As I said during a discussion, I find it a bit strange to "encode" the fact that there is no port with a `-1`. By looking at the method API it's not obvious compared to a Nullable or Optional parameter (or even 2 different methods, even if the -1 is then used internally). And as it is not documented it could be hard to guess

damien-urruty-sonarsource (approved):
2 minor comments, LGTM!

server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/ServerApiHelper.java
```diff
@@ -107,9 +107,12 @@ public CompletableFuture<HttpClient.Response> rawGetAsync(String relativePath) {
...
 
   private String buildEndpointUrl(String relativePath) {
+    return concat(endpointParams.getBaseUrl(), relativePath);
+  }
+
+  public static String concat(String baseUrl, String relativePath) {
     var fullUrl = new StringBuilder();
-    var endpointUrl = endpointParams.getBaseUrl();
-    fullUrl.append(endpointUrl.endsWith("/") ? endpointUrl.substring(0, endpointUrl.length() - 1) : endpointUrl);
+    fullUrl.append(baseUrl.endsWith("/") ? baseUrl.substring(0, baseUrl.length() - 1) : baseUrl);
```
kirill-knize-sonarsource, line 115: StringUtils.removeEnd() ?

damien-urruty-sonarsource: I simplified the code. To me the real move would be to not use a String to represent a URL, and it would help with those operations

PR #517, opened 2022-10-18
Add initial version of the configuration service

henryju:


damien-urruty-sonarsource (approved):
Not a big fan of the name `ConfigurationScope`, we can still see later

PR #518, opened 2022-10-20
Basic implementation of the configuration service

henryju:


core/src/main/java/org/sonarsource/sonarlint/core/AutoBinding.java
```diff
@@ -0,0 +1,408 @@
...
+import org.sonarsource.sonarlint.core.serverapi.ServerApi;
+import org.sonarsource.sonarlint.core.serverapi.component.ServerProject;
+
+import static java.util.stream.Collectors.toList;
+import static java.util.stream.Collectors.toSet;
+import static org.apache.commons.lang.StringUtils.removeEnd;
+import static org.apache.commons.lang.StringUtils.trimToNull;
+import static org.apache.commons.lang3.StringUtils.isNotBlank;
+
+public class AutoBinding {
```
damien-urruty-sonarsource, line 66: The class is a bit big, maybe we could split it ? There are different concerns, e.g. extracting info from the `.properties` files and auto-bind

core/src/main/java/org/sonarsource/sonarlint/core/AutoBinding.java
```diff
@@ -0,0 +1,408 @@
...
+    if (bindingConfiguration.isAutoBindEnabled()) {
+      autoBindConfigScope(configScopeId);
+    }
+  }
+
+  @Subscribe
+  public void connectionAdded(ConnectionAddedEvent event) {
+    // Double check if added connection has not been removed in the meantime
+    if (connectionReferential.getConnectionById(event.getAddedConnectionId()) != null) {
+      autoBindAll();
```
damien-urruty-sonarsource, line 109: First, I was a bit surprised that the just added connection is not used at all to filter binding suggestions, which means we can suggest the user to bind to a different connection. I wonder if this wouldn't be too noisy, as the user would have probably already received a notification for that (by e.g. opening a project)

core/src/main/java/org/sonarsource/sonarlint/core/AutoBinding.java
```diff
@@ -0,0 +1,408 @@
...
+        .map(SonarCloudConnectionConfiguration.class::cast)
+        .filter(c -> organization == null || Objects.equals(organization, c.getOrganization()))
+        .map(AbstractConnectionConfiguration::getConnectionId)
+        .collect(toSet());
+    }
+    return connectionReferential.getConnectionsById().keySet();
+  }
+
+  private static boolean isSameServerUrl(String left, String right) {
+    return Objects.equals(removeEnd(left, "/"), removeEnd(right, "/"));
```
damien-urruty-sonarsource, line 272: I would expect to find this in the SonarQubeConnectionConfiguration directly.
But then I realized this class is the DTO that we use in the RPC method. So my suggestion is to introduce a separate `Connection` type, in which we could add e.g. `Connection#hasSameHost(Connection)`
I don't think we should rely on the DTO for all our operations in the backend

core/src/main/java/org/sonarsource/sonarlint/core/AutoBinding.java
```diff
@@ -0,0 +1,408 @@
...
+    if (!bindingConfiguration.isAutoBindEnabled()) {
+      LOG.debug("Configuration scope '{}' has auto-bind disabled. Skipping.", configScopeId);
+      return false;
+    }
+    return true;
+  }
+
+  private boolean isBound(BindingConfiguration bindingConfiguration) {
+    return bindingConfiguration.getConnectionId() != null
+      && bindingConfiguration.getSonarProjectKey() != null
```
damien-urruty-sonarsource, line 397: Same remark goes here, this `isBound` could be moved to a `Binding` class (not the DTO)

core/src/main/java/org/sonarsource/sonarlint/core/AutoBinding.java
```diff
@@ -0,0 +1,408 @@
...
+  public AutoBinding(ConfigurationReferential configReferential, ConnectionConfigurationReferential connectionReferential, SonarLintClient client) {
+    this.configReferential = configReferential;
+    this.connectionReferential = connectionReferential;
+    this.client = client;
+  }
+
+  @Subscribe
+  public void bindingConfigChanged(BindingConfigChangedEvent event) {
+    // Check if auto-bind is switched on
+    if (event.getNewConfig().isAutoBindEnabled() && !event.getPreviousConfig().isAutoBindEnabled()) {
```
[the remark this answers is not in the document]
henryju: Settings can always be manipulated by the users, depending on the IDE. I don't think it has to be handled on the SonarLint UI side, but I choose to handle the case in the backend.

core/src/main/java/org/sonarsource/sonarlint/core/AutoBinding.java
```diff
@@ -0,0 +1,408 @@
...
+          searchGoodMatchInConnections(candidates, configScopeName, bindingClueWithConnections.connectionIds);
+        }
+        if (cluesWithoutProjectKey.isEmpty()) {
+          searchGoodMatchInConnections(candidates, configScopeName, connectionReferential.getConnectionsById().keySet());
+        }
+      }
+    }
+
+    if (!candidates.isEmpty()) {
+      client.suggestAutoBind(new SuggestAutoBindParams(candidates));
```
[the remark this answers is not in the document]
henryju: Indeed, and as discussed, I will send only one notification for all config scopes. I might allow better UX (especially in Eclipse).

core/src/main/java/org/sonarsource/sonarlint/core/AutoBinding.java
henryju: I will only move the part checking for connectionId and projectKey being non null. To check if the connection actually exists, I would need to introduce a coupling between `BindingConfiguration` and `ConnectionConfig`. As discussed, I would like to introduce at some point a concept of "LiveConnection" and "LiveBinding", that we could use in this case. That will come later.

client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/config/scope/ConfigurationScopeWithBinding.java
```diff
@@ -0,0 +1,38 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.sonarlint.core.clientapi.config.scope;
+
+import javax.annotation.Nullable;
+import org.eclipse.lsp4j.jsonrpc.validation.NonNull;
+import org.sonarsource.sonarlint.core.clientapi.config.binding.BindingConfiguration;
+
+public class ConfigurationScopeWithBinding extends ConfigurationScope {
```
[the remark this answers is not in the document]
henryju: Removed parent class

core/src/main/java/org/sonarsource/sonarlint/core/AutoBinding.java
henryju: I will do it in another PR

core/src/main/java/org/sonarsource/sonarlint/core/AutoBinding.java
henryju: I don't think the current behavior is wrong (if the user previously received a notification, he might have marked the project as "don't auto-bind").
I will see how to do better in another PR.

core/src/main/java/org/sonarsource/sonarlint/core/AutoBinding.java
henryju: Created DTOs

core/src/main/java/org/sonarsource/sonarlint/core/ConfigurationServiceImpl.java
```diff
@@ -0,0 +1,98 @@
...
+
+  public ConfigurationServiceImpl(EventBus clientEventBus, ConfigurationReferential referential) {
+    this.clientEventBus = clientEventBus;
+    this.referential = referential;
+  }
+
+  @Override
+  public CompletableFuture<Void> initialize(InitializeParams params) {
+    params.getConfigScopes().forEach(referential::addOrReplace);
+    return CompletableFuture.completedFuture(null);
```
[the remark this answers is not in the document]
damien-urruty-sonarsource: FYI the way I implemented it in IntelliJ is that I initialize the backend before any project is opened. So only the connection service is initialized. I suspect we would implement it like this for every IDE, so maybe the question is should we have a `ConfigurationService.initialize` at all ?

damien-urruty-sonarsource (approved):

core/src/main/java/org/sonarsource/sonarlint/core/ConfigurationServiceImpl.java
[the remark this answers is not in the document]
henryju: Maybe not, indeed. There is maybe no difference of opening a project at startup and later.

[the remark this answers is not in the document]
henryju: In fact, for Eclipse, if we call `didAddConfigurationScope` once for each project, we will have auto-binding executed separately for each, which will be bad. So we need to keep the ability to register multiple scopes.

PR #519, opened 2022-10-25
Improve binding suggestion

henryju:


core/src/main/java/org/sonarsource/sonarlint/core/BindingClueProvider.java
```diff
@@ -0,0 +1,254 @@
...
+    public Set<String> getConnectionIds() {
+      return connectionIds;
+    }
+  }
+
+  private List<BindingClue> collectBindingClues(String checkedConfigScopeId) throws InterruptedException {
+    LOG.debug("Query client for binding clues...");
+    FindFileByNamesInScopeResponse response;
+    try {
+      response = client.findFileByNamesInScope(new FindFileByNamesInScopeParams(checkedConfigScopeId, List.of(SONAR_SCANNER_CONFIG_FILENAME, AUTOSCAN_CONFIG_FILENAME))).get(1,
```
damien-urruty-sonarsource, line 103: This will block the event bus thread for max 1min, not sure it's a problem or not (or maybe not now)

damien-urruty-sonarsource (approved):
LGTM! I've noticed that there are a lot of places where we don't use `var`

core/src/main/java/org/sonarsource/sonarlint/core/BindingClueProvider.java
henryju: Good catch, I will move that to a separate thread.

PR #522, opened 2022-10-27
Introduce Architecture Decision Records

damien-urruty-sonarsource:


docs/decisions/0000-move-more-responsibilities-to-the-core.md
```diff
@@ -0,0 +1,4 @@
+# Move more responsibilities from clients to the core
+
+See [this document](https://docs.google.com/document/d/1oFpgtS_cvufdzItMTTSnTTlF4ayBL_s5dDOKtf8Mgyo/edit?usp=sharing)
```
henryju, line 3: This document is not public, so I would avoid linking it from GitHub.

henryju (approved):
LGTM, even if I wonder if it should not be split in multiple ADR, as there are many decisions in the same file.

damien-urruty-sonarsource:
This stayed in a PR for too long, and we probably should have documented other decisions since these initial ones, but I'm merging in case people want to contribute other ADRs

PR #523, opened 2022-10-28
SLCORE-413 Add client method to list all taint vulnerabilities for a project

jblievremont:


henryju (commented):
I don't think this is the good way to use Xodus. It would be better to add a new link between branch and taint issues, to avoid involving the file entity. This would require a migration, and this is not something we have done yet.

jblievremont:
I'm wondering whether a migration would be necessary, or "just" invalidation + re-sync.

server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/storage/XodusServerIssueStore.java
```diff
@@ -546,4 +577,39 @@ private static Location toProtoLocation(ServerIssueLocation l) {
     }
     return location.build();
   }
+
+  static void checkCurrentSchemaVersion(StoreTransaction txn) {
+    var currentSchemaVersion = getCurrentSchemaVersion(txn);
+    if (currentSchemaVersion < CURRENT_SCHEMA_VERSION) {
+      // Migrate v0 to v1: force re-sync of taint vulnerabilities
+      txn.findWithProp(BRANCH_ENTITY_TYPE, NAME_PROPERTY_NAME).forEach(b -> b.setProperty(LAST_TAINT_SYNC_PROPERTY_NAME, Instant.EPOCH));
```
henryju, line 585: What is the point of `findWithProp` here? Why not simply `getAll`?

server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/storage/XodusServerIssueStore.java
```diff
@@ -546,4 +577,39 @@ private static Location toProtoLocation(ServerIssueLocation l) {
...
+
+      // Set schema version to current after migration(s)
+      txn.findWithProp(SCHEMA_ENTITY_TYPE, VERSION_PROPERTY_NAME).forEach(Entity::delete);
```
henryju, line 588: Same, I would use `getAll`

server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/storage/XodusServerIssueStore.java
```diff
@@ -546,4 +577,39 @@ private static Location toProtoLocation(ServerIssueLocation l) {
...
+      var newSchema = txn.newEntity(SCHEMA_ENTITY_TYPE);
+      newSchema.setProperty(VERSION_PROPERTY_NAME, CURRENT_SCHEMA_VERSION);
+      txn.saveEntity(newSchema);
+      txn.flush();
+    }
+  }
+
+  static int getCurrentSchemaVersion(StoreTransaction txn) {
+    var schemaEntities = txn.findWithProp(SCHEMA_ENTITY_TYPE, VERSION_PROPERTY_NAME);
```
henryju, line 597: Same, I would use `getAll`

henryju (changes requested):
LGTM, with a few comments.
I think a test is missing: if I'm correct the specific migration v0 -> v1 is not tested. We could maybe have a medium test. Not sure how to produce the "old" DB. I think we should avoid storing binaries as test resources, so maybe you could create the "old" DB programmatically. Or maybe: create a new DB, then revert the schema version to zero, reopen the DB, and check that all last sync timestamps have been reset to zero.

core/src/main/java/org/sonarsource/sonarlint/core/client/api/connected/ConnectedSonarLintEngine.java
```diff
@@ -73,11 +73,22 @@ public interface ConnectedSonarLintEngine extends SonarLintEngine {
...
   List<ServerTaintIssue> getServerTaintIssues(ProjectBinding projectBinding, String branchName, String filePath);
 
+  /**
+   * Gets locally stored server taint issues.
+   *
+   * @param projectBinding information about the project (must have been previously updated with {@link #updateProject(EndpointParams, HttpClient, String, boolean, String, ClientProgressMonitor)})
+   * @param branchName     branch name
+   * @param serverPathToIdePath a function to map a server path to an actual file known by the IDE, if any
+   * @return All server taint issues in the local storage for the given file. If file has no issues, an empty list is returned.
+   */
```
sophio-japharidze-sonarsource, line 89: I think this annotation needs to be updated? `serverPathToIdePath` is an extra `@param`, and the description - If I understood correctly this method gets all server taint issues for a branch (not the given _file_).

sophio-japharidze-sonarsource (approved):

henryju (approved):

PR #528, opened 2022-11-09
Deprecate quick fixes API

damien-urruty-sonarsource:


pom.xml
```diff
@@ -42,7 +42,7 @@
 
   <properties>
     <!-- Don't forget to also update SonarPluginRequirementsChecker::IMPLEMENTED_PLUGIN_API -->
```
henryju, line 44: Ping :)

plugin-api/src/main/java/org/sonarsource/sonarlint/plugin/api/issue/NewInputFileEdit.java
```diff
@@ -25,6 +25,7 @@
  * Describe a file edit for a {@link NewQuickFix} as a collection of {@link NewTextEdit}s on a given {@link InputFile}.
  * Text edits are applied in the order they are added, insofar that their ranges do not overlap.
  * @since 6.3
+ * @deprecated use org.sonar.api.batch.sensor.issue.fix.NewInputFileEdit from the sonar-plugin-api instead
```
henryju, line 28: Why not also adding the `@Deprecated` annotation on the interfaces?

henryju (changes requested):

henryju (approved):
LGTM!

PR #533, opened 2022-11-23
SLCORE-420 Allow security hotspot rules to be run in connected mode

jblievremont:


sophio-japharidze-sonarsource (approved):

henryju (approved):
LGTM. Maybe add in the hotspot IT suite a test to ensure we can get the rule description for a hotspot (even if in theory this should work the same as other issues)

PR #534, opened 2022-11-28
SLCORE-416 Resolve effective binding from parent scope when not configured

damien-urruty-sonarsource:


henryju (approved):
If I'm correct, we could now add a medium test or IT to verify the use case of a module having its parent bound, and check that active rule details are correct (ie come from the parent binding).

PR #540, opened 2022-12-02
SLCORE-426 Add telemetry support in the new backend

henryju:


client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/backend/InitializeParams.java
```diff
@@ -41,11 +42,18 @@ public class InitializeParams {
   private final boolean enableSecurityHotspots;
   private final List<SonarQubeConnectionConfigurationDto> sonarQubeConnections;
   private final List<SonarCloudConnectionConfigurationDto> sonarCloudConnections;
+  private final String sonarlintUserHome;
 
-  public InitializeParams(Path storageRoot, Set<Path> embeddedPluginPaths, Map<String, Path> connectedModeExtraPluginPathsByKey,
+  /**
+   * @param productKey        SonarLint product key (vscode, idea, eclipse, ...)
```
damien-urruty-sonarsource, line 48: It's not clear to me if this value could be any random unique identifier, or if there is a list of predefined values that are accepted.

core/src/main/java/org/sonarsource/sonarlint/core/hotspot/HotspotServiceImpl.java
```diff
@@ -56,6 +60,8 @@ public void openHotspotInBrowser(OpenHotspotInBrowserParams params) {
     var url = buildHotspotUrl(effectiveBinding.get().getSonarProjectKey(), params.getBranch(), params.getHotspotKey(), endpointParams.get());
 
     client.openUrlInBrowser(new OpenUrlInBrowserParams(url));
+
+    telemetryService.hotspotOpenedInBrowser();
```
damien-urruty-sonarsource, line 64: I think it's ok for now, maybe at some point we will need events to decouple services from telemetry

core/src/test/java/mediumtest/TelemetryMediumTests.java
```diff
@@ -0,0 +1,117 @@
...
+
+    // Emulate another process has disabled telemetry
+    var telemetryLocalStorageManager = new TelemetryLocalStorageManager(TelemetryPathManager.getPath(sonarlintUserHome, SonarLintBackendFixture.MEDIUM_TESTS_PRODUCT_KEY));
+    telemetryLocalStorageManager.tryUpdateAtomically(data -> {
+      data.setEnabled(false);
+    });
+
+    assertThat(backend.getTelemetryService().getStatus().get().isEnabled()).isFalse();
+  }
+
```
damien-urruty-sonarsource, line 115: as it is mentioned in the Javadoc somewhere, we should maybe also test the default `~/.sonarlint` location is used if no SL home is provided

core/src/test/java/mediumtest/TelemetryMediumTests.java
```diff
@@ -0,0 +1,117 @@
...
+      System.clearProperty("sonarlint.telemetry.disabled");
+    } else {
+      System.setProperty("sonarlint.telemetry.disabled", oldValue);
+    }
+  }
+
+  @Test
+  void it_should_not_create_telemetry_file_if_telemetry_disabled_by_system_property(@TempDir Path sonarlintUserHome) throws ExecutionException, InterruptedException {
+    System.setProperty("sonarlint.telemetry.disabled", "true");
+    var fakeClient = newFakeClient().build();
```
damien-urruty-sonarsource, line 62: If you don't use the fakeClient for anything, there is a `build` method without parameter. That simplifies a bit

damien-urruty-sonarsource (approved):
LGTM, only a few minor comments

client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/backend/InitializeParams.java
henryju: At the moment, this value is only used in telemetry. Maybe I should rename it `telemetryProductKey`?

core/src/main/java/org/sonarsource/sonarlint/core/hotspot/HotspotServiceImpl.java
henryju: yes, possibly

core/src/test/java/mediumtest/TelemetryMediumTests.java
henryju: I don't want to take the risk to mess up my local `~/.sonarlint`, but maybe I can add a unit test.

PR #541, opened 2022-12-06
SLCORE-415 Use released version of plugin-api 9.13

damien-urruty-sonarsource:


henryju (approved):
I would have used a NOOP implementation for `NewMessageFormatting`, but this is minor.

PR #544, opened 2022-12-13
SLCORE-432 Extract description sections for security hotspots

jblievremont:


its/tests/src/test/java/its/ConnectedModeBackendTest.java
```diff
@@ -181,6 +187,174 @@ public void returnDescriptionSectionsForTaintRules() throws ExecutionException,
...
+    assumeThat(ORCHESTRATOR.getServer().version()).isGreaterThanOrEqualTo(Version.create("9.7"));
+
+    backend.getConfigurationService().didAddConfigurationScopes(new DidAddConfigurationScopesParams(
+      List.of(new ConfigurationScopeDto("project", null, true, "Project", new BindingConfigurationDto("ORCHESTRATOR", PROJECT_KEY_JAVA_HOTSPOT, false)))));
+
+    var activeRuleDetailsResponse = backend.getActiveRulesService().getActiveRuleDetails(new GetActiveRuleDetailsParams("project", javaRuleKey(ORCHESTRATOR, "S4792"))).get();
+
+    var description = activeRuleDetailsResponse.details().getDescription();
+    assertThat(description)
+      .extracting("right.introductionHtmlContent")
```
henryju, line 201: I think using `extracting` here is not required. Would be cleaner with `assertThat(description.getRight().getIntroductionHtmlContent())`.

its/tests/src/test/java/its/ConnectedModeBackendTest.java
```diff
@@ -181,6 +187,174 @@ public void returnDescriptionSectionsForTaintRules() throws ExecutionException,
...
+      .isNull();
+    assertThat(description)
+      .extracting("right.tabs", as(list(ActiveRuleDescriptionTabDto.class)))
```
henryju, line 204: Same, I would not use `extracting` here.

damien-urruty-sonarsource (approved):
LGTM, my only comment is that I recently removed asserting rule description sections content in another test. The description can change relatively often so it is a bit fragile to check that

henryju (approved):
Minor feedback then LGTM

PR #548, opened 2023-01-04
SLCORE-434 Expose education principles in the More Info tab

damien-urruty-sonarsource:


core/src/main/java/org/sonarsource/sonarlint/core/rules/EducationPrinciples.java
```diff
@@ -0,0 +1,94 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.sonarlint.core.rules;
+
+import java.util.HashMap;
+import java.util.Map;
+import javax.annotation.CheckForNull;
+import org.sonarsource.sonarlint.core.commons.log.SonarLintLogger;
+
+class EducationPrinciples {
```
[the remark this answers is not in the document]
damien-urruty-sonarsource: bien vu !

core/pom.xml
```diff
@@ -138,22 +138,6 @@
   </dependencies>
 
   <build>
-    <resources>
```
henryju, line 141: It was dead code right? Maybe should have been in a separate commit.

core/src/main/java/org/sonarsource/sonarlint/core/rules/CleanCodePrinciples.java
```diff
@@ -0,0 +1,48 @@
...
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import javax.annotation.CheckForNull;
+import org.apache.commons.io.IOUtils;
+import org.sonarsource.sonarlint.core.commons.log.SonarLintLogger;
+
+class CleanCodePrinciples {
+
+  @CheckForNull
+  public static String getContent(String key) {
```
henryju, line 31: @mithil-dobarkar-sonarsource in the end Damien moved content to resources. Maybe you can discuss with him the rational to see if you should do the same on your side for the "other" context content.

henryju (approved):
LGTM

core/pom.xml
damien-urruty-sonarsource: Yep, split in 2 commits

core/src/main/java/org/sonarsource/sonarlint/core/rules/CleanCodePrinciples.java
damien-urruty-sonarsource: My rationale is that it is static content, it was a bit ugly in the code (we don't have text blocks) and I had a lot of issues from Sonar :)
In this case I think it is more logical to have them as resources

PR #550, opened 2023-01-09
Migrate ITs to JUnit5

henryju:


its/tests/src/test/java/com/sonar/orchestrator/OrchestratorBuilder.java
```diff
@@ -0,0 +1,166 @@
...
+import com.sonar.orchestrator.server.StartupLogWatcher;
+import com.sonar.orchestrator.util.OrchestratorUtils;
+import com.sonar.orchestrator.util.System2;
+import java.io.File;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import javax.annotation.Nullable;
+
+public class OrchestratorBuilder {
```
henryju, line 36: @damien-urruty-sonarsource I am in the process of fixing Orchestrator, so this class can be removed later.

its/tests/src/test/java/com/sonar/orchestrator/OrchestratorExtension.java
```diff
@@ -0,0 +1,46 @@
...
+import com.sonar.orchestrator.Orchestrator;
+import com.sonar.orchestrator.config.Configuration;
+import com.sonar.orchestrator.container.SonarDistribution;
+import com.sonar.orchestrator.server.StartupLogWatcher;
+import org.jetbrains.annotations.Nullable;
+import org.junit.jupiter.api.extension.AfterAllCallback;
+import org.junit.jupiter.api.extension.BeforeAllCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+public class OrchestratorExtension extends Orchestrator implements BeforeAllCallback, AfterAllCallback {
```
henryju, line 31: This one should also been removed after I update Orchestrator

its/tests/src/test/java/its/ConnectedModeExcludeByVersionTest.java
```diff
@@ -1,119 +0,0 @@
...
-
-import static its.tools.ItUtils.SONAR_VERSION;
-import static java.util.Collections.emptySet;
-import static org.assertj.core.api.Assertions.assertThat;
-import static org.junit.Assume.assumeTrue;
-
-/**
- * This test verifies that SonarPython 1.13 gets excluded due to being below minimum supported version.
- */
-public class ConnectedModeExcludeByVersionTest extends AbstractConnectedTest {
```
damien-urruty-sonarsource, line 55: This IT was removed in the process, is it normal ?

damien-urruty-sonarsource (approved):
One question, else LGTM

its/tests/src/test/java/its/ConnectedModeExcludeByVersionTest.java
henryju: Yes, I think this test is not really interesting as an IT. We already have unit tests for that anyway. It is very costly to start a SQ server just for that old case IMO. My goal (see next PR) is to share instances of SQ as much as possible, and this might require dropping a few tests.

PR #551, opened 2023-01-09
Speed up ITs

henryju:


damien-urruty-sonarsource (approved):
I really like this new hierarchy with nested test classes, kudos for the big rework!

PR #553, opened 2023-01-11
SLCORE-437 Provide a cli to extract rules metadata from plugins

henryju:


duncanp-sonar:
@henryju as discussed, could you add the internal key property please?

Also, I think the manifest is missing the `Main-Class` setting.

henryju:
@duncanp-sonar I have added the internalKey.

> Also, I think the manifest is missing the Main-Class setting.
The manifest entry is only in the "fat JAR" (the one with name xxx-jar-with-dependencies.jar)

damien-urruty-sonarsource (approved):
I have one concern. The `RulesDefinitionExtractor.extractRules()` method returns a List<SonarLintRuleDefinition> that is used to generate the json output. It means that SonarLintRuleDefinition becomes a contract, and we could easily forget that, change the name of a field, add or remove one and it would break on the reader side. 
I would suggest to introduce another data type (yes, again) in this new module. It would also help clearly identify what fields are needed by consumers

henryju:
> It means that SonarLintRuleDefinition becomes a contract

We discussed that with @duncanp-sonar and we agreed this is good enough as a first start. It will only be used at build time by the SLVS bubble, so a breaking change in the output is not a big deal. We will see if we need to introduce a more specific contract later.

duncanp-sonar:
> It will only be used at build time

I'll make sure it causes a build failure in SLVS if the file doesn't contain the expected fields so we can't miss it on our side.

PR #554, opened 2023-01-12
SLCORE-438 map securityStandard to vulnerability probability

sophio-japharidze-sonarsource:


core/src/test/java/org/sonarsource/sonarlint/core/tracking/ServerHotspotTrackableTests.java
```diff
@@ -24,6 +24,7 @@
 import org.sonarsource.sonarlint.core.commons.RuleType;
 import org.sonarsource.sonarlint.core.commons.TextRange;
 import org.sonarsource.sonarlint.core.serverapi.hotspot.ServerHotspot;
+import org.sonarsource.sonarlint.core.commons.VulnerabilityProbability;
```
henryju, line 27: Useless import?

rule-extractor/pom.xml
```diff
@@ -59,6 +60,10 @@
       <artifactId>mockito-core</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
henryju, line 63: Why do you need this dependency? If you really need it, please put it with other production dependencies, not with test dependencies.

henryju (changes requested):
A few comments, but overall LGTM!

core/src/test/java/org/sonarsource/sonarlint/core/tracking/ServerHotspotTrackableTests.java
sophio-japharidze-sonarsource: Needed for line 36

rule-extractor/pom.xml
sophio-japharidze-sonarsource: It was needed for the new `SecurityStandards` class, to use `ImmutableMap` and `Ordering.` Moved it under regular dependencies

henryju: Normally, with Java 11+, you can avoid using Guava, it will allow to keep SLCORE distribution smaller

henryju (approved):

PR #558, opened 2023-01-27
Update leftover copyright notices

jblievremont:


core/src/main/proto/sonarlint.proto
```diff
@@ -1,5 +1,5 @@
 // SonarLint, open source software quality management tool.
-// Copyright (C) 2015-2022 SonarSource
+// Copyright (C) 2015-2023 SonarSource
```
henryju, line 2: Should be a way to configure the license plugin to handle proto files, but I didn't had the time to look at it.

henryju (approved):
Good catch anyway, thanks!

PR #560, opened 2023-01-31
SLCORE-445 Honor server settings for secrets in connected mode

kirill-knize-sonarsource:


core/src/main/java/org/sonarsource/sonarlint/core/ConnectedSonarLintEngineImpl.java
```diff
@@ -321,7 +319,7 @@ public CompletableFuture<ConnectedRuleDetails> getActiveRuleDetails(EndpointPara
     var ruleDefFromPluginOpt = analysisContext.get().findRule(ruleKey);
     if (ruleDefFromPluginOpt.isPresent()) {
       var ruleDefFromPlugin = ruleDefFromPluginOpt.get();
-      if (globalConfig.getExtraPluginsPathsByKey().containsKey(ruleDefFromPlugin.getLanguage().getPluginKey()) || projectKey == null) {
```
damien-urruty-sonarsource, line 324: Removing this condition will cause an error when connected to an old SQ. Without the condition we will try to get the rule from the server while it does not exist. I suggest to keep loading the active rule details from the rule definition with the condition: rule language is secrets and serverConnection.supportSecretAnalysis() is false

core/src/test/java/mediumtest/ActiveRulesMediumTests.java
```diff
@@ -113,83 +113,60 @@ void it_should_return_rule_loaded_from_server_plugin_when_project_is_bound_and_p
...
 
-  @Test
-  void it_should_return_embedded_rule_when_project_is_bound_and_rule_comes_from_extra_plugin() throws ExecutionException, InterruptedException {
-    StorageFixture.newStorage("connectionId")
-      .withProject("projectKey")
-      .create(storageDir);
-    backend = newBackend()
-      .withBoundConfigScope("scopeId", "connectionId", "projectKey")
-      .withStorageRoot(storageDir.resolve("storage"))
-      .withExtraPlugin(TestPlugin.PYTHON)
```
damien-urruty-sonarsource, line 124: This `withExtraPlugin` method is probably not required anymore

damien-urruty-sonarsource (commented):
You

damien-urruty-sonarsource (approved):
:+1:

PR #567, opened 2023-02-17
SLCORE-452 Adapt tab title for security hotspots rule description

damien-urruty-sonarsource:


core/src/main/java/org/sonarsource/sonarlint/core/rules/ActiveRuleDetailsAdapter.java
```diff
@@ -143,12 +139,25 @@ private static Collection<ActiveRuleDescriptionTabDto> transformSectionsButIntro
...
+    if (ROOT_CAUSE_SECTION_KEY.equals(sectionKey)) {
+      return RuleType.SECURITY_HOTSPOT.equals(ruleDetails.getType()) ? "What's the risk?" : "Why is this an issue?";
+    }
+    if (ASSESS_THE_PROBLEM_SECTION_KEY.equals(sectionKey)) {
+      return "Assess the risk";
+    }
+    if (HOW_TO_FIX_SECTION_KEY.equals(sectionKey)) {
+      return "How can I fix it?";
+    }
+    return "More Info";
```
henryju, line 158: Shouldn't we explicitely test if sectionKey == RESOURCES_SECTION_KEY?
In case of unknown section key, we could log a warning and simply return the key.

henryju (approved):
Minor comment then LGTM

core/src/main/java/org/sonarsource/sonarlint/core/rules/ActiveRuleDetailsAdapter.java
damien-urruty-sonarsource: At this point it's not possible, the sectionKey parameter has been validated earlier

henryju: ok, I see

PR #568, opened 2023-02-18
VulnerabilityProbability: get enum by string

giufus:
Hi there, 

this is just a tiny contribution that could help to fix the red sign in [current quality gate](https://next.sonarqube.com/sonarqube/project/issues?resolved=false&inNewCodePeriod=true&types=BUG&id=org.sonarsource.sonarlint.intellij%3Asonarlint-intellij&open=AYY8HvK4ujquqN6aARrd) for sonarlint-intellij project.

I don't think there is an issue about that in the jira project (core or intellij).

### DIFF
--- a/commons/src/main/java/org/sonarsource/sonarlint/core/commons/VulnerabilityProbability.java
+++ b/commons/src/main/java/org/sonarsource/sonarlint/core/commons/VulnerabilityProbability.java
@@ -46,4 +46,12 @@ public enum VulnerabilityProbability {
       .filter(t -> t.score == score)
       .findFirst();
   }
+  public static Optional<VulnerabilityProbability> byName(@Nullable String name) {
+    if (name == null) {
+      return Optional.empty();
+    }
+    return Arrays.stream(values())
+      .filter(t -> t.name() == name)
+      .findFirst();
+  }
 }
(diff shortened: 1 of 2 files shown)


commons/src/main/java/org/sonarsource/sonarlint/core/commons/VulnerabilityProbability.java
```diff
@@ -46,4 +46,12 @@ public static Optional<VulnerabilityProbability> byScore(@Nullable Integer score
       .filter(t -> t.score == score)
       .findFirst();
   }
+  public static Optional<VulnerabilityProbability> byName(@Nullable String name) {
+    if (name == null) {
+      return Optional.empty();
+    }
+    return Arrays.stream(values())
+      .filter(t -> t.name() == name)
```
nquinquenel, line 54: When comparing Strings you should use equals()

commons/src/test/java/org/sonarsource/sonarlint/core/commons/VulnerabilityProbabilityTest.java
```diff
@@ -0,0 +1,27 @@
+package org.sonarsource.sonarlint.core.commons;
+
+import org.junit.jupiter.api.Test;
+
+class VulnerabilityProbabilityTest {
+
+    @Test
+    public void byName_when_name_is_ok() {
+        assert VulnerabilityProbability.byName("HIGH").isPresent();
```
nquinquenel, line 9: For assertion we use Assertj library

nquinquenel (changes requested):
Hello giufus, thank you for you contribution. However, I do no understand what is the issue you are trying to resolve, the link you provided doesn't display anything in particular. 

Could you please improve the PR description and make it clear for us, currently we cannot accept your contribution as it is.

PR #580, opened 2023-03-23
Update the plugin API

henryju:


plugin-commons/src/main/java/org/sonarsource/sonarlint/core/plugin/commons/sonarapi/SonarLintLoggerImpl.java
```diff
@@ -17,36 +17,40 @@
...
+import javax.annotation.Nullable;
+import org.sonar.api.utils.log.Logger;
+import org.sonar.api.utils.log.LoggerLevel;
 import org.sonarsource.sonarlint.core.commons.log.SonarLintLogger;
 
-/**
- * This class can't be moved to another package because {@link BaseLogger} is not public. And we have to extend BaseLogger to please {@link DefaultProfiler}.
- */
-public class SonarApiLogger extends BaseLogger {
+public class SonarLintLoggerImpl implements Logger {
```
damien-urruty-sonarsource, line 27: This class is called `SonarLintLoggerImpl` and has a field with type `SonarLintLogger` but I don't see the relation between the 2 (I was expecting e.g. the 1st one implementing the 2nd one). Maybe one of the names could be improved ?

plugin-commons/src/main/java/org/sonarsource/sonarlint/core/plugin/commons/loading/Slf4jBridgeClassLoader.java
```diff
@@ -20,28 +20,26 @@
...
-    // Use Platform ClassLoader as parent in order to avoid finding slf4j in the Application classloader (some IDEs can provide slf4j to plugins)
+    // Use Platform ClassLoader as parent in order to avoid finding SLF4J in the Application classloader (some IDEs can provide SLF4J to plugins)
     super(ClassLoader.getPlatformClassLoader());
     this.sonarLintClassLoader = sonarLintClassLoader;
   }
 
   @Override
   protected Class<?> findClass(final String name) throws ClassNotFoundException {
-    if (name.startsWith("org.sonar.api.utils.log")) {
+    if (name.startsWith("org.sonarsource.sonarlint.core.commons.log")) {
```
damien-urruty-sonarsource, line 42: Shouldn't we keep the old value for backward compatibility ?

damien-urruty-sonarsource (approved):
A few questions, else LGTM

plugin-commons/src/main/java/org/sonarsource/sonarlint/core/plugin/commons/loading/Slf4jBridgeClassLoader.java
henryju: I don't think so

plugin-commons/src/main/java/org/sonarsource/sonarlint/core/plugin/commons/sonarapi/SonarLintLoggerImpl.java
henryju: Indeed, I will try something

PR #581, opened 2023-03-24
SLCORE-461 Move smart notifications to core

nquinquenel:


core/src/main/java/org/sonarsource/sonarlint/core/smartnotifications/LastEventPolling.java
```diff
@@ -0,0 +1,55 @@
...
+  private final SmartNotificationsStorage storage;
+
+  public LastEventPolling(Path storageRoot) {
+    this.storage = new SmartNotificationsStorage(storageRoot);
+  }
+
+  public ZonedDateTime getLastEventPolling(String projectKey, String connectionId) {
+    var lastEventPollingEpoch = storage.getLastEventPolling(projectKey, connectionId);
+    ZonedDateTime lastEventPolling;
+    lastEventPolling =
```
damien-urruty-sonarsource, line 39: Maybe join lines and use var ?

core/src/main/java/org/sonarsource/sonarlint/core/smartnotifications/SmartNotifications.java
```diff
@@ -0,0 +1,120 @@
...
+        n.message(), n.link(), scopeIds.get(n.projectKey()));
+      client.showSmartNotification(smartNotification);
+    }
+  }
+
+  public void shutdown() {
+    smartNotificationsPolling.shutdownNow();
+  }
+
+  private static ZonedDateTime getLastNotificationTime(ZonedDateTime lastTime) {
```
damien-urruty-sonarsource, line 115: I don't think this is used, should it be ?

core/src/test/java/mediumtest/SmartNotificationsMediumTests.java
```diff
@@ -0,0 +1,125 @@
...
+      .withBoundConfigScope("scopeId", CONNECTION_ID, PROJECT_KEY)
+      .withBoundConfigScope("scopeId2", CONNECTION_ID, PROJECT_KEY)
+      .withBoundConfigScope("scopeId3", CONNECTION_ID_2, PROJECT_KEY)
+      .withBoundConfigScope("scopeId4", CONNECTION_ID_2, PROJECT_KEY_2)
+      .withStorageRoot(tmpDir)
+      .build(fakeClient);
+
+    await().atMost(3, TimeUnit.SECONDS).until(() -> !fakeClient.getSmartNotificationsToShow().isEmpty());
+
+    assertThat(fakeClient.getSmartNotificationsToShow()).hasSize(3);
```
damien-urruty-sonarsource, line 114: If possible I would assert the content of the notifications

core/src/main/java/org/sonarsource/sonarlint/core/smartnotifications/LastEventPolling.java
```diff
@@ -0,0 +1,55 @@
...
+      lastEventPollingEpoch.map(aLong -> ZonedDateTime.ofInstant(Instant.ofEpochMilli(aLong), ZoneId.systemDefault())).orElseGet(ZonedDateTime::now);
+    setLastEventPolling(lastEventPolling, projectKey, connectionId);
+    return lastEventPolling;
+  }
+
+  public void setLastEventPolling(ZonedDateTime dateTime, String projectKey, String connectionId) {
```
damien-urruty-sonarsource, line 45: I don't see this method used outside of this class, is it expected ?

core/src/main/java/org/sonarsource/sonarlint/core/smartnotifications/SmartNotifications.java
nquinquenel: Yep it should be used, (based on current implementation) we shouldn't retrieve more than a day old of notifications in case the last polling event is older than one day. I forgot to add this :smile:

core/src/test/java/mediumtest/SmartNotificationsMediumTests.java
nquinquenel: Yes! Tests in progress :keyboard:

core/src/main/java/org/sonarsource/sonarlint/core/smartnotifications/SmartNotifications.java
```diff
@@ -0,0 +1,134 @@
...
+    var oneDayAgo = ZonedDateTime.now().minusDays(1);
+    return lastTime.isAfter(oneDayAgo) ? lastTime : oneDayAgo;
+  }
+
+  private Map<String, Map<String, Set<String>>> getProjectKeysAndConfigScopeIdsPerConnectionIds() {
```
damien-urruty-sonarsource, line 113: I suggest moving this method to ConfigurationRepository. Also it's a bit complex, what about something like:

```
public Map<String, Map<String, Set<String>>> getScopeIdsPerProjectKeyPerConnectionId() {
  Map<String, Map<String, Set<String>>> scopeIdsPerProjectKeyPerConnectionId = new HashMap<>();
  bindingPerConfigScopeId
    .forEach((scopeId, bindingConfiguration) -> scopeIdsPerProjectKeyPerConnectionId.computeIfAbsent(bindingConfiguration.getConnectionId(), k -> new HashMap<>())
      .computeIfAbsent(bindingConfiguration.getSonarProjectKey(), k -> new HashSet<>())
      .add(scopeId));
  return scopeIdsPerProjectKeyPerConnectionId;
}
```
Also the Map<String, Map<String, Set<String>>> looks primitive, there is probably a concept to introduce, like `ConnectionBindings`? To have a Map<String, ConnectionBindings>

core/src/main/java/org/sonarsource/sonarlint/core/smartnotifications/SmartNotifications.java
```diff
@@ -0,0 +1,134 @@
...
+    this.client = client;
+  }
+
+  public void initialize(Path storageRoot) {
+    lastEventPollingService = new LastEventPolling(storageRoot);
+    smartNotificationsPolling = Executors.newSingleThreadScheduledExecutor(r -> new Thread(r, "Smart Notifications Polling"));
+    smartNotificationsPolling.scheduleAtFixedRate(this::poll, 1, 60, TimeUnit.SECONDS);
+  }
+
+  private void poll() {
```
damien-urruty-sonarsource, line 64: I would suggest to split this method into smaller ones, e.g. a separate poll() method that you would call per connection ?

damien-urruty-sonarsource (commented):
Another round of comments, mostly about cosmetics

core/src/main/java/org/sonarsource/sonarlint/core/smartnotifications/SmartNotifications.java
nquinquenel: That's better thanks, it was just missing the check `isBound()` in your example

`ConnectionBindings` -> it would be more like `ProjectScopes` in our case, I'll try to come up with something convenient

core/src/main/java/org/sonarsource/sonarlint/core/smartnotifications/SmartNotifications.java
```diff
@@ -0,0 +1,133 @@
...
+
+  private void manageNotificationsForConnection(ServerApi serverApi, Map<String, Set<String>> scopeIdsPerProjectKey,
+    String connectionId) {
+    var developersApi = serverApi.developers();
+
+    isConnectionIdSupported.putIfAbsent(connectionId, developersApi.isSupported());
+    if (Boolean.TRUE.equals(isConnectionIdSupported.get(connectionId))) {
+      var projectKeysByLastEventPolling =
+        scopeIdsPerProjectKey.keySet().stream()
+          .distinct()
```
damien-urruty-sonarsource, line 93: I think distinct is not useful here as you already have a Set ?

damien-urruty-sonarsource (approved):
A few minor comments, else LGTM! :+1:

core/src/main/java/org/sonarsource/sonarlint/core/repository/connection/AbstractConnectionConfiguration.java
```diff
@@ -27,26 +27,33 @@ public abstract class AbstractConnectionConfiguration {
...
     var that = (AbstractConnectionConfiguration) o;
-    return Objects.equals(connectionId, that.connectionId);
+    return Objects.equals(connectionId, that.connectionId)
+      && Objects.equals(disableNotifications, that.disableNotifications);
   }
 
   @Override
   public int hashCode() {
-    return Objects.hash(connectionId);
+    return Objects.hash(connectionId, disableNotifications);
```
henryju, line 57: Doesn't look very useful to have the disableNotifications part of the hash, since connectionId should be unique anyway.

core/src/main/java/org/sonarsource/sonarlint/core/smartnotifications/LastEventPolling.java
```diff
@@ -0,0 +1,52 @@
...
+
+public class LastEventPolling {
+
+  private final SmartNotificationsStorage storage;
+
+  public LastEventPolling(Path storageRoot) {
+    this.storage = new SmartNotificationsStorage(storageRoot);
+  }
+
+  public ZonedDateTime getLastEventPolling(String projectKey, String connectionId) {
```
henryju, line 36: This is minor, but I tend to order parameters by having connectionId before projectKey, as this is a kind of hierarchical relationship.

spec/glossary.adoc
```diff
@@ -20,4 +20,5 @@
 [[sensor]]Sensor:: A Sensor is the Java interface implemented by <<sonar_analyzer,Sonar analyzers>> to analyze source files and report <<finding,findings>>.
 [[sonar_analyzer]]Sonar Analyzer:: A component responsible to analyze user's code, for one or multiple languages.
 [[sonar_project]]Sonar Project:: A project that exists on SonarQube or SonarCloud. Sonar Projects are identified by a <<project_key,projectKey>>.
+[[smart_notification]]Smart Notifications:: A feature that enables the IDE to receive notifications, only available in <<connected_mode,connected mode>> with SonarQube or SonarCloud. You receive a smart notification when the quality gate status of an opened project in your IDE changes, or when an analysis raises new issues that you've introduced in an opened project.
```
henryju, line 23: Only available with SQ *commercial editions*

henryju (commented):
A few comments already. It would be great to test that adding/removing a connection/binding after the backend has been started will lead to the correct behavior.

spec/glossary.adoc
damien-urruty-sonarsource: I think they had moved it to CE as well. I don't see any mention about a commercial edition on the [docs](https://docs.sonarqube.org/latest/user-guide/sonarlint-connected-mode/)

henryju: Indeed: https://sonarsource.atlassian.net/browse/SONAR-14306
Good to know, maybe we should change ITs to test with the community edition.

PR #583, opened 2023-03-31
 SLCORE-465 Synchronize Sonar projects from the backend 

damien-urruty-sonarsource:


client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/SonarLintClient.java
```diff
@@ -89,4 +91,18 @@ public interface SonarLintClient {
    */
   @JsonRequest
   CompletableFuture<AssistBindingResponse> assistBinding(AssistBindingParams params);
+
+  /**
+   * Requests the client to start showing progress to users.
+   * If there is an error while creating the corresponding UI, clients can fail the returned future.
+   * Tasks requesting the start of the progress should wait for the client to answer before continuing.
+   */
+  @JsonNotification
```
henryju, line 101: Should be `@JsonRequest`

client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/SonarLintClient.java
```diff
@@ -105,4 +106,6 @@ public interface SonarLintClient {
    */
   @JsonNotification
   void reportProgress(ReportProgressParams params);
+
+  void didSynchronizeConfigurationScopes(DidSynchronizeConfigurationScopeParams params);
```
henryju, line 110: Missing `@JsonNotification`

core/src/main/java/org/sonarsource/sonarlint/core/sync/SynchronizationServiceImpl.java
```diff
@@ -0,0 +1,159 @@
...
+    this.serverApiProvider = serverApiProvider;
+    this.taskManager = new TaskManager(client);
+  }
+
+  public void initialize(Path storageRoot, Path workDir, Set<Language> enabledLanguagesInConnectedMode, boolean taintVulnerabilitiesEnabled) {
+    this.storageRoot = storageRoot;
+    this.workDir = workDir;
+    this.enabledLanguagesInConnectedMode = enabledLanguagesInConnectedMode;
+    this.taintVulnerabilitiesEnabled = taintVulnerabilitiesEnabled;
+    scheduledSynchronizer = Executors.newSingleThreadScheduledExecutor(r -> new Thread(r, "Sonar Project Synchronizer"));
```
henryju, line 78: The thread name is a bit limiting IMO. What about "SonarLint Local Storage Synchronizer"?

henryju (approved):
A few minor comments, but overall LGTM!

nquinquenel (approved):
LGTM

client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/client/progress/StartProgressParams.java
```diff
@@ -0,0 +1,78 @@
...
+  /**
+   * The task ID is a unique identifier, generated by the backend that identifies a long-running task.
+   * The same ID needs to be re-used when reporting progress to the client for a given task.
+   */
+  private final String taskId;
+  /**
+   * Configuration scope on which to report the progress. Could be null for a global task that does not relate to a configuration scope in particular
+   */
+  @CheckForNull
+  private final String configurationScopeId;
```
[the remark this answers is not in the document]
damien-urruty-sonarsource: You're right, and I am not sure in terms of UX how that would work on the client side. 

On SLI when starting a background task, if I specify a project the progress will be shown in the project window. If I provide `null`, IntelliJ displays the progress on the window in foreground when starting the task. I think returning the list of config scopes wouldn't be useful, as we wouldn't know on the client side how to split
The other option from the backend with the current API is to trigger several progress reports, one per config scope for the sync for example.

I suggest leaving it like this and we can still revisit later

PR #588, opened 2023-04-17
SLCORE-472 remove empty lines from clean code principles page

sophio-japharidze-sonarsource:
Before:
image (6)
After:
Screenshot 2023-04-17 at 16 56 51

### DIFF
--- a/core/src/main/java/org/sonarsource/sonarlint/core/rules/RuleDetailsAdapter.java
+++ b/core/src/main/java/org/sonarsource/sonarlint/core/rules/RuleDetailsAdapter.java
@@ -160,7 +160,7 @@ class RuleDetailsAdapter {
   private static String concat(Collection<String> htmlSnippets) {
     return htmlSnippets.stream()
       .filter(StringUtils::isNotBlank)
-      .collect(Collectors.joining("<br/><br/>"));
+      .collect(Collectors.joining());
   }
 
   private static String getTabContent(RuleDetails.DescriptionSection section, @Nullable String extendedDescription, Set<String> educationPrincipleKeys) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/sonarsource/sonarlint/core/rules/RuleDetailsAdapter.java
```diff
@@ -160,7 +160,7 @@ private static String getTabTitle(RuleDetails ruleDetails, String sectionKey) {
   private static String concat(Collection<String> htmlSnippets) {
     return htmlSnippets.stream()
       .filter(StringUtils::isNotBlank)
-      .collect(Collectors.joining("<br/><br/>"));
+      .collect(Collectors.joining());
```
henryju, line 163: @damien-urruty-sonarsource do you remember why you added those extra BR? I think it makes sense to remove them and deal with IDE specific L&F on IDE side (with CSS or something else), but maybe this was required for SLI?

sophio-japharidze-sonarsource: FYI this is current display in IntelliJ
Screenshot 2023-04-18 at 09 34 40

henryju (approved):

core/src/main/java/org/sonarsource/sonarlint/core/rules/RuleDetailsAdapter.java
damien-urruty-sonarsource: No, I don't remember why I added those extra line breaks... Probably not a problem for SLI

PR #591, opened 2023-04-19
Manage httpclient in the backend

henryju:


server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/ServerApiHelper.java
```diff
@@ -106,7 +106,7 @@ public CompletableFuture<HttpClient.Response> rawGetAsync(String relativePath) {
     return client.getAsync(url)
       .whenComplete((response, error) -> {
         var duration = Duration.between(startTime, Instant.now());
-        LOG.debug("{} {} {} | response time={}ms", "GET", response.code(), url, duration.toMillis());
+        LOG.debug("{} {} {} | response time={}ms", "GET", response != null ? response.code() : null, url, duration.toMillis());
```
jblievremont, line 109: FYI, a similar issue exists a few lines above in `rawGet`

PR #594, opened 2023-05-04
SLCORE-477 Implement ability to query binding suggestions

vojtech-suchy-sonarsource:


core/src/main/java/org/sonarsource/sonarlint/core/SonarLintBackendImpl.java
```diff
@@ -193,6 +193,11 @@ public RulesServiceImpl getRulesService() {
     return rulesService;
   }
 
+  @Override
+  public BindingSuggestionProviderImpl getBindingSuggestionService(){
```
henryju, line 197:
```suggestion
  public BindingSuggestionProviderImpl getBindingSuggestionService() {
```

core/src/main/java/org/sonarsource/sonarlint/core/BindingSuggestionProviderImpl.java
```diff
@@ -117,22 +123,33 @@ public void connectionAdded(ConnectionConfigurationAddedEvent event) {
...
+
   private void queueBindingSuggestionComputation(Set<String> configScopeIds, Set<String> candidateConnectionIds) {
     executorService.submit(() -> {
-      try {
-        if (enabled.get()) {
-          bindingSuggestionComputation(configScopeIds, candidateConnectionIds);
-        } else {
-          LOG.debug("Skipping binding suggestion computation as it is disabled");
-        }
-      } catch (InterruptedException e) {
```
henryju, line 128: Out of curiosity, why have you removed this?

core/src/main/java/org/sonarsource/sonarlint/core/BindingSuggestionProviderImpl.java
```diff
@@ -117,22 +123,33 @@ public void connectionAdded(ConnectionConfigurationAddedEvent event) {
     }
   }
 
+  @Override
+  public CompletableFuture<SuggestBindingResponse> suggestBinding(GetBindingSuggestionParams params) {
+    return CompletableFuture.supplyAsync(() -> {
+      var suggestions = computeBindingSuggestions(Set.of(params.getConfigScopeId()), Set.of(params.getConnectionId()));
+      return new SuggestBindingResponse(suggestions);
+    }, executorService);
```
henryju, line 131: Nice! I think it is good to share the same thread pool.

client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/client/SuggestBindingParent.java
```diff
@@ -0,0 +1,39 @@
...
+package org.sonarsource.sonarlint.core.clientapi.client;
+
+import java.util.List;
+import java.util.Map;
+import org.eclipse.lsp4j.jsonrpc.validation.NonNull;
+import org.sonarsource.sonarlint.core.clientapi.backend.config.binding.BindingSuggestionDto;
+
+public class SuggestBindingParent {
+
+  Map<String, List<BindingSuggestionDto>> suggestions;
```
henryju, line 29: Should be private

client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/client/SuggestBindingParent.java
```diff
@@ -0,0 +1,39 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.sonarlint.core.clientapi.client;
+
+import java.util.List;
+import java.util.Map;
+import org.eclipse.lsp4j.jsonrpc.validation.NonNull;
+import org.sonarsource.sonarlint.core.clientapi.backend.config.binding.BindingSuggestionDto;
+
+public class SuggestBindingParent {
```
henryju, line 27: Even if this will create a bit of duplication, I have a doubt about having a common ancestor here. From an OOP point of view it doesn't make sense in my opinion.
Also we want to keep this client API stable, so we should avoid sharing request/params class between different requests, as a change on one side can lead to an unexpected change on the other side.

client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/backend/binding/BindingSuggestionService.java
```diff
@@ -0,0 +1,34 @@
...
+import org.sonarsource.sonarlint.core.clientapi.client.SuggestBindingResponse;
+
+public interface BindingSuggestionService {
+
+  /**
+   * Calculates a suggested binding for a 'configScopeId' and 'connectionId' specified in the {@link GetBindingSuggestionParams}
+   * @return {@link SuggestBindingResponse} containing binding suggestions
+   */
+  @JsonRequest
+  CompletableFuture<SuggestBindingResponse> suggestBinding(GetBindingSuggestionParams params);
```
henryju, line 33: There is a risk of confusion with the notification on the client interface. Maybe we should find a different name?
```suggestion
  CompletableFuture<GetBindingSuggestionsResponse> getBindingSuggestions(GetBindingSuggestionsParams params);
```

client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/SonarLintBackend.java
```diff
@@ -53,6 +53,9 @@ public interface SonarLintBackend {
   @JsonDelegate
   RulesService getRulesService();
 
+  @JsonDelegate
+  BindingSuggestionService getBindingSuggestionService();
```
henryju, line 57: I wonder if this is not too granular as a service. We can always change later, but maybe we could already name it simply `BindingService`?

henryju (changes requested):

core/src/main/java/org/sonarsource/sonarlint/core/BindingSuggestionProviderImpl.java
vojtech-suchy-sonarsource: The log is in the [queueBindingSuggestionComputation](https://github.com/SonarSource/sonarlint-core/blob/feature/vs/SLCORE-477_implement_option_to_query_binding_suggestions/core/src/main/java/org/sonarsource/sonarlint/core/BindingSuggestionProviderImpl.java#L139) method now.

client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/SonarLintBackend.java
vojtech-suchy-sonarsource: Good point; Address in [078e014](https://github.com/SonarSource/sonarlint-core/pull/594/commits/078e01419b054088432516cdda1ff62c8418bd96)

core/src/main/java/org/sonarsource/sonarlint/core/BindingSuggestionProviderImpl.java
henryju: I was more asking for the try catch

client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/backend/binding/BindingSuggestionService.java
vojtech-suchy-sonarsource: Good point; Addressed in [078e014](https://github.com/SonarSource/sonarlint-core/pull/594/commits/078e01419b054088432516cdda1ff62c8418bd96)

client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/client/SuggestBindingParent.java
vojtech-suchy-sonarsource: Good point; Addressed in [078e014](https://github.com/SonarSource/sonarlint-core/pull/594/commits/078e01419b054088432516cdda1ff62c8418bd96)

henryju (approved):

PR #595, opened 2023-05-04
Fix null pointer access when HTTP host is unreachable

jblievremont:


henryju (commented):
Is it something you can reproduce in a test? Since we are gonna refactor the HTTP client, I would like to avoid regressions later.

nquinquenel:
This is something I also noticed, and I made a modification on the HTTP client [PR](https://github.com/SonarSource/sonarlint-core/pull/591/files#diff-07e823031c1ccf4c915bdd663fa0df7c3a27cf04b8697a6b4712b03917946a81L109), LGTM :+1:

damien-urruty-sonarsource:
This PR is opened for a long time, what should we do with it? I never encountered this error myself, is it fixed with what you mentioned Nicolas? If so, then we can close this one

jblievremont:
Closing this old PR that has been fixed in a different way.

PR #602, opened 2023-05-17
SLCORE-487 Deprecate explicit option to sync taint vulnerabilities

damien-urruty-sonarsource:


core/src/main/java/org/sonarsource/sonarlint/core/languages/LanguageSupportRepository.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+package org.sonarsource.sonarlint.core.languages;
+
+import java.util.Collection;
+import java.util.EnumSet;
+import java.util.Set;
+import org.sonarsource.sonarlint.core.commons.Language;
+
+public class LanguageSupportRepository {
+  private static final EnumSet<Language> LANGUAGES_RAISING_TAINT_VULNERABILITIES = EnumSet.of(Language.CS, Language.JAVA, Language.JS, Language.TS, Language.PHP, Language.PYTHON);
```
nquinquenel, line 28: Quick question, how do you make sure this list is up-to-date?

nquinquenel (approved):
LGTM!

core/src/main/java/org/sonarsource/sonarlint/core/languages/LanguageSupportRepository.java
damien-urruty-sonarsource: That's a good question! It's pretty rare that a new language is supported in the security analyzer, so I don't expect this list to change frequently

PR #603, opened 2023-05-17
 SLCORE-485 Expose if changing hotspot status is permitted 

damien-urruty-sonarsource:


server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/hotspot/HotspotApi.java
```diff
@@ -192,10 +204,9 @@ private static HotspotReviewStatus getStatus(Hotspots.SearchWsResponse.Hotspot h
     return HotspotReviewStatus.TO_REVIEW;
   }
 
-  private static String getShowUrl(String hotspotKey, String projectKey) {
+  private static String getShowUrl(String hotspotKey) {
     return HOTSPOTS_SHOW_API_URL
-      + "?projectKey=" + urlEncode(projectKey)
-      + "&hotspot=" + urlEncode(hotspotKey);
+      + "?hotspot=" + urlEncode(hotspotKey);
```
nquinquenel, line 209: Why were we using `projectKey` as a parameter? API documentation doesn't show anything for this param

nquinquenel (approved):
LGTM

server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/hotspot/HotspotApi.java
damien-urruty-sonarsource: I have no idea.. I thought it was required by some old versions of SonarQube but I was not able to find it

PR #607, opened 2023-05-23
Update to Orchestrator 4.0

henryju:


its/tests/src/test/java/its/utils/OrchestratorUtils.java
```diff
@@ -19,15 +19,16 @@
...
 
 import static its.utils.ItUtils.SONAR_VERSION;
 
 public class OrchestratorUtils {
-  public static OrchestratorBuilder defaultEnvBuilder() {
-    return Orchestrator.builderEnv()
+  public static OrchestratorExtensionBuilder defaultEnvBuilder() {
+    return OrchestratorExtension.builderEnv()
       .defaultForceAuthentication()
+      .keepBundledPlugins()
```
nquinquenel, line 31: No big deal to set `keepBundledPlugins` by default whereas it was previously only enabled for Security Hotspot and Taint vulnerabilities?

henryju: If I'm not mistaken, `keepBundledPlugins` was always used in all test suites. I simply factorized it there. Anyway I think for us this is good to use a SQ distributions as close as possible to the "real life".

PR #608, opened 2023-05-23
SLCORE-488 SSE for hotspots

sophio-japharidze-sonarsource:


server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/events/hotspot/UpdateStorageOnSecurityHotspotRaised.java
```diff
@@ -0,0 +1,50 @@
...
+      event.getRuleKey(),
+      event.getMainLocation().getMessage(),
+      event.getMainLocation().getFilePath(),
+      adapt(event.getMainLocation().getTextRange()),
+      event.getCreationDate(),
+      event.getStatus(),
+      event.getVulnerabilityProbability(),
+      null
+    );
+    storage.project(event.getProjectKey()).findings().insert(event.getBranch(), hotspot);
```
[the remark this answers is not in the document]
sophio-japharidze-sonarsource: This is being checked in the `insert()` [here](https://github.com/SonarSource/sonarlint-core/blob/cf944c455c06634c77b487083e8ba3dbc165a9bb/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/storage/XodusServerIssueStore.java#L655) . Or do you mean to add another check here before calling `insert`?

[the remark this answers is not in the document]
jblievremont: 🙈 missed that one. Then LGTM 👍🏻!

server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/hotspot/HotspotApi.java
```diff
@@ -182,14 +182,13 @@ private static ServerHotspot adapt(Hotspots.SearchWsResponse.Hotspot hotspot, St
...
-      VulnerabilityProbability.valueOf(hotspot.getVulnerabilityProbability()));
+      getStatus(hotspot.getStatus(), hotspot.getResolution()),
+      VulnerabilityProbability.valueOf(hotspot.getVulnerabilityProbability()),
+      hotspot.getAssignee());
   }
 
-  private static HotspotReviewStatus getStatus(Hotspots.SearchWsResponse.Hotspot hotspot) {
-    var status = hotspot.getStatus();
-    if ("REVIEWED".equals(status) && hotspot.hasResolution()) {
-      var resolution = hotspot.getResolution();
```
damien-urruty-sonarsource, line 192: There is a specificity with protobuf, generated getters never return null for Strings. The default value is the empty string `""`. The presence or absence of a field is normally checked with `hasXXX` methods. So here the second part of your new condition will always evaluate as `true`. I don't think it's an issue, if the payload is formed as expected, if the status is REVIEWED there should always be a resolution. The previous check was a kind of guard just in case

server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/storage/XodusServerIssueStore.java
```diff
@@ -234,8 +235,10 @@ private static ServerHotspot adaptHotspot(Entity storedHotspot) {
     var startLineOffset = (Integer) storedHotspot.getProperty(START_LINE_OFFSET_PROPERTY_NAME);
     var endLine = (Integer) storedHotspot.getProperty(END_LINE_PROPERTY_NAME);
     var endLineOffset = (Integer) storedHotspot.getProperty(END_LINE_OFFSET_PROPERTY_NAME);
-    var textRange = new org.sonarsource.sonarlint.core.commons.TextRange(startLine, startLineOffset, endLine, endLineOffset);
+    var hash = (String) storedHotspot.getProperty(RANGE_HASH_PROPERTY_NAME);
+    var textRangeWithHash = new TextRangeWithHash(startLine, startLineOffset, endLine, endLineOffset, hash);
```
damien-urruty-sonarsource, line 239: The hash might be null for hotspots that were already in the store, could this cause a problem later?

damien-urruty-sonarsource: Today the hash won't be used for tracking, as in `ServerHotspotTrackable` we return `null` from `getTextRange()`. As not all hotspots will have a hash (see previous comment) you might need to return something from `ServerHotspotTrackable.getTextRange()` only if a hash is present

damien-urruty-sonarsource (approved):
LGTM, great work!

PR #611, opened 2023-05-26
Simplify services initialization

henryju:
I was tired of this "two steps" initialization, where we have to deal with many objects having non-final fields, since they are only properly initialized after the `initialize` request has been called.
I had this idea of having 2 "facades", the first one being only there to deal with the initialization request, and then delegating everything to the second facade, which would be fully initialized.


damien-urruty-sonarsource (commented):
While I agree about the non-final fields being annoying, I am not sure the solution really simplifies things. Now when we add a new service we have 2 places to modify. 

One thing I tried recently to avoid passing a lot of things in `initialize` methods is to introduce repositories, see `LanguageSupportRepository`. I think it makes more sense to wrap the data around a repository (also for example because language support might be dynamically changed one day through a new service). If we continue that way we could imagine removing all the initialize methods.

Happy to discuss it further

PR #614, opened 2023-06-01
First steps for SSL support

henryju:
Sorry, I think I messed up with my previous PR, you might have already reviewed the first 2 commits.

### DIFF
--- a/core/src/main/java/org/sonarsource/sonarlint/core/http/HttpClientManager.java
+++ b/core/src/main/java/org/sonarsource/sonarlint/core/http/HttpClientManager.java
@@ -19,6 +19,8 @@
  */
 package org.sonarsource.sonarlint.core.http;
 
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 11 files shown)


core/src/main/java/org/sonarsource/sonarlint/core/http/HttpClientManager.java
```diff
@@ -63,48 +60,55 @@ public class HttpClientManager {
...
-      .setUserAgent(userAgent)
-      // SLI-629 - Force HTTP/1
-      .setVersionPolicy(HttpVersionPolicy.FORCE_HTTP_1)
-      // proxy settings
-      .setRoutePlanner(new SystemDefaultRoutePlanner(new ProxySelector() {
+    var sslFactoryBuilder = SSLFactory.builder()
+      .withDefaultTrustMaterial()
+      .withSystemTrustMaterial()
+
+      .withTrustMaterial(new X509TrustManager() {
```
nquinquenel, line 71: Minor remark since it will be likely reworked when doing the real implementation, but I think this should be moved to its own class as it will probably get more complex

nquinquenel (approved):
LGTM for a first step!

One remark as a side note for potential future tests. When testing against a trusted self-signed certificate from WireMock, with the current implementation it will fail with a `SSLPeerUnverifiedException` due to WireMock's certificate lacking the information about server identity, as needed for the hostname [check](https://datatracker.ietf.org/doc/html/rfc2818#section-3.1).

It's possible to make it pass by having the hostname verifier trust everything by default for the tests, but I think a more suitable solution will be to generate our own certificate and pass it to WireMock.

PR #615, opened 2023-06-01
Fix test files generated in the middle of sources

damien-urruty-sonarsource:


damien-urruty-sonarsource:
While making sure that no test files were generated in the sources, I made the following changes:
* have reasonable default folders in the backend builder (storage root, user home). This simplifies tests as they don't have to care about the temp dirs
* improve the builder to let tests declare the storage when they declare a connection
* augment the backend impl with a test implementation to let tests access extra information
* have the telemetry test check the contents on disk (not using the production code to load it)

core/src/test/java/mediumtest/fixtures/SonarLintBackendFixture.java
```diff
@@ -242,7 +236,15 @@ public SonarLintBackendImpl build(FakeSonarLintClient client) {
       return sonarLintBackend;
     }
 
-    public SonarLintBackendImpl build() {
+    private static Path tempDirectory(String prefix) {
+      try {
+        return Files.createTempDirectory(prefix);
```
henryju, line 264: Will this folder be deleted after tests are run?

henryju (approved):
I have a doubt regarding the manual creation of the temp folder. Maybe we should change the SonarLintBackendFixture to a JUnit 5 Extension, so that it can use the tempfolder managed by JUnit?
Apart from that, LGTM.

core/src/test/java/mediumtest/fixtures/SonarLintBackendFixture.java
damien-urruty-sonarsource: yes, it's in the shutdown method of the `SonarLintTestBackend`.

Regarding using a JUnit 5 extension, I thought about it and there are 2 reasons why I did not move forward:
* this [issue](https://github.com/junit-team/junit5/issues/1786) suggests that an extension cannot rely on `@TempDir`. We would need to manually manage those files as described in the comments
* The backend usually has different parameters in each test, so the extension would have to be a wrapper around the creation of the backend builder, which would make things a bit more complicated

PR #617, opened 2023-06-02
Add a MutableTrustManager to save accepted certificates

henryju:


henryju:
@nquinquenel I will work on code quality next week, but I think you can already have a look at this part.

core/src/main/java/org/sonarsource/sonarlint/core/http/HttpClientManager.java
```diff
@@ -64,34 +60,14 @@ public class HttpClientManager {
...
-
-    var sonarlintTruststore = sonarlintUserHome.resolve("/ssl/truststore.jks");
-    if (Files.exists(sonarlintTruststore)) {
-      sslFactoryBuilder.withTrustMaterial(sonarlintTruststore, "changeit".toCharArray());
-    }
+      .withTrustMaterial(confirmingTrustManager)
+      .build();
     var asyncConnectionManager = PoolingAsyncClientConnectionManagerBuilder.create()
-      .setTlsStrategy(Apache5SslUtils.toTlsStrategy(sslFactoryBuilder.build()))
+      .setTlsStrategy(new DefaultClientTlsStrategy(sslFactory.getSslContext()))
```
nquinquenel, line 70: I'd suggest using
`new DefaultClientTlsStrategy(
                sslFactory.getSslContext(),
                sslFactory.getSslParameters().getProtocols(),
                sslFactory.getSslParameters().getCipherSuites(),
                SSLBufferMode.STATIC,
                sslFactory.getHostnameVerifier()
        )`
to fully use our sslFactory object parameters, and also to move from the default hostname verifier to the OkHttpClient one (copied and used by the library)

core/src/main/java/org/sonarsource/sonarlint/core/http/MutableTrustManager.java
```diff
@@ -0,0 +1,214 @@
...
+ * https://github.com/JetBrains/intellij-community/blob/f4fc17b0c44d38d65fb7ad47b968bed55c889609/platform/platform-api/src/com/intellij/util/net/ssl/ConfirmingTrustManager.java#L313
+ * 
+ */
+public final class MutableTrustManager implements X509TrustManager {
+
+  private static final X509Certificate[] NO_CERTIFICATES = new X509Certificate[0];
+
+  private static final SonarLintLogger LOG = SonarLintLogger.get();
+  private final Path myPath;
+  private final String myPassword;
```
nquinquenel, line 56: I think it's not necessary to store the password as an attribute, can be passed in the constructor once to create the key store, and then we can pass the password when adding a certificate. It would avoid having it in the memory, and it shouldn't be a big performance issue to read the system property only when adding a certificate. Also, the password should be stored as char[] for security purposes IMO

WDYT?

henryju: I don't see how we can avoid having the password in memory at some point.
I am not aware that `char[]` is more secure than a String. Do you have any pointers to that?

henryju: Also I think the goal is to not create an empty keystore as long as a custom certificate has not been approved

henryju: OK, interesting to know, but I don't think this is a concern for us, as the password is a constant anyway.

core/src/main/java/org/sonarsource/sonarlint/core/http/HttpClientManager.java
Hakky54: Just noticed this comment. Yes the default hostname verifier of SSLFactory is a copy of the one from OkHttp

PR #619, opened 2023-06-06
Extract HTTP related code in a separate module, plus cleanup

henryju:


core/src/test/java/mediumtest/AuthenticationMediumTests.java
```diff
@@ -49,7 +49,7 @@
 class AuthenticationMediumTests {
 
   @RegisterExtension
-  WireMockExtension sonarqubeMock = WireMockExtension.newInstance()
+  static WireMockExtension sonarqubeMock = WireMockExtension.newInstance()
```
nquinquenel, line 52: Regarding such changes, it should probably be done also in `SslMediumTests` and `ProxyMediumTests`

nquinquenel (approved):
Minor comments but LGTM

core/src/main/java/org/sonarsource/sonarlint/core/http/ClientProxyCredentialsProvider.java
```diff
@@ -0,0 +1,59 @@
...
+
+  @Override
+  public Credentials getCredentials(AuthScope authScope, HttpContext context) {
+    try {
+      var response = client.getProxyPasswordAuthentication(
+        new GetProxyPasswordAuthenticationParams(authScope.getHost(), authScope.getPort(), authScope.getProtocol(),
+          authScope.getRealm(), authScope.getSchemeName()))
+        .get();
+      if (response.getProxyUser() != null || response.getProxyPassword() != null) {
+        return new UsernamePasswordCredentials(response.getProxyUser(), response.getProxyPassword().toCharArray());
```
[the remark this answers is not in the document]
henryju: Good catch!

core/src/test/java/mediumtest/AuthenticationMediumTests.java
henryju: Indeed. FYI I am fixing it on-the-fly since this is a very minor performance detail (the local HTTP server will be restarted between each tests)

PR #620, opened 2023-06-07
Add method to get HotspotStatus value from its title

kirill-knize-sonarsource:
Since we pass only titles to the frontend we need a way to map them back to enum values.

### DIFF
--- a/core/src/test/java/mediumtest/hotspots/HotspotStatusTests.java
+++ b/core/src/test/java/mediumtest/hotspots/HotspotStatusTests.java
@@ -21,6 +21,7 @@ package mediumtest.hotspots;
 
 import org.junit.jupiter.api.Test;
 import org.sonarsource.sonarlint.core.clientapi.backend.hotspot.HotspotStatus;
+import org.sonarsource.sonarlint.core.commons.HotspotReviewStatus;
 
 import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
 import static org.junit.jupiter.api.Assertions.assertThrows;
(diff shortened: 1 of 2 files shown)


core/src/test/java/mediumtest/hotspots/HotspotStatusTests.java
```diff
@@ -0,0 +1,48 @@
...
+    assertThat(HotspotStatus.valueOfTitle("Safe")).isEqualTo(HotspotStatus.SAFE);
+    assertThat(HotspotStatus.valueOfTitle("Fixed")).isEqualTo(HotspotStatus.FIXED);
+    assertThat(HotspotStatus.valueOfTitle("Acknowledged")).isEqualTo(HotspotStatus.ACKNOWLEDGED);
+    var thrown = assertThrows(IllegalArgumentException.class, ()-> HotspotStatus.valueOfTitle("Unknown"));
+    assertThat(thrown).hasMessage("There is no such title of the hotspot status: Unknown");
+  }
+
+  @Test
+  void valueOfHotspotReviewStatusTest() {
+    for (HotspotReviewStatus value : HotspotReviewStatus.values()) {
```
kirill-knize-sonarsource, line 43: If new value will be added to HotspotReviewStatus, the test will fail if corresponding value will not be added to HotspotStatus.

damien-urruty-sonarsource (commented):
I added a few comments. I am guessing a few things so my understanding might be wrong, but I would not be in favor of putting these changes in the core. Of course, we can discuss it further

PR #628, opened 2023-06-15
SLCORE-500 Move smart notifications check to backend

nquinquenel:


core/src/test/java/mediumtest/ConnectionSetupMediumTests.java
```diff
@@ -202,6 +204,44 @@ void it_should_fails_to_validate_connection_if_host_not_found() throws IOExcepti
...
+
+    var connectionResponse = backend.getConnectionService()
+      .checkSmartNotificationsSupported(new CheckSmartNotificationsSupportedParams(
+        new TransientSonarQubeConnectionDto(server.baseUrl(), Either.forLeft(new TokenDto("foo"))))).get();
+
+    assertThat(connectionResponse).isTrue();
+  }
+
+  @Test
+  void it_should_not_support_notifications_if_sonarqube_7_9() throws ExecutionException, InterruptedException {
```
damien-urruty-sonarsource, line 233: To make it more clear what the limit is, you could have a test with a version just before they were supported, and another test with the version that brings support.

its/tests/src/test/java/its/SonarCloudTests.java
```diff
@@ -447,6 +448,16 @@ void testConnection() throws ExecutionException, InterruptedException {
     assertThat(failIfWrongCredentials.getMessage()).isEqualTo("Authentication failed");
   }
 
+  @Test
+  void testSmartNotifications() throws ExecutionException, InterruptedException {
```
damien-urruty-sonarsource, line 452: I'm not sure this test is necessary, as it is covered by a medium test already

damien-urruty-sonarsource (approved):
LGTM, might need a bit of improvement on the Javadocs

client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/backend/connection/ConnectionService.java
```diff
@@ -61,4 +62,10 @@ public interface ConnectionService {
   @JsonRequest
   CompletableFuture<ValidateConnectionResponse> validateConnection(ValidateConnectionParams params);
 
+  /**
+   * Check that smart notifications are supported by the server
+   */
+  @JsonRequest
+  CompletableFuture<Boolean> checkSmartNotificationsSupported(CheckSmartNotificationsSupportedParams params);
```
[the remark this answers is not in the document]
nquinquenel: I originally started with `CheckSmartNotificationsSupportedResponse` but then switched to simple Boolean as the class was just consisting of this attribute, but I see your point, it makes sense in the end

core/src/test/java/mediumtest/ConnectionSetupMediumTests.java
nquinquenel: I decided to go for the manual `withSmartNotificationsSupported(boolean)` as there are a lot of cases to take into account if we were to test between edition and version

PR #633, opened 2023-06-20
Bumped sslcontext kickstart and switched to TrustManagerParameters for predicate

Hakky54:
The withInflatableTrustMaterial has a better/alternative parameter now. Before it was needed to pass the chain and auth type, however this is marked now as deprecated and will be removed in the future.

The [TrustManagerParameters](https://github.com/Hakky54/sslcontext-kickstart/blob/master/sslcontext-kickstart/src/main/java/nl/altindag/ssl/model/TrustManagerParameters.java) contains for now the chain, authtype, socket and sslengine and additional validations can be done with those variables if reeded. The TrustManagerParameters can also grow over time and have more variable in the future. With the existing setup it was not possible to add these options.

### DIFF
--- a/core/src/main/java/org/sonarsource/sonarlint/core/http/AskClientCertificatePredicate.java
+++ b/core/src/main/java/org/sonarsource/sonarlint/core/http/AskClientCertificatePredicate.java
@@ -19,17 +19,17 @@
  */
 package org.sonarsource.sonarlint.core.http;
 
-import java.security.cert.X509Certificate;
 import java.util.Arrays;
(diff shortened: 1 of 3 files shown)


Hakky54:
Hi @henryju I created this small PR, however the build is failing. Any idea what the issue is? I am not able to see the details

henryju:
Our pipeline is not able to run on external PRs (to not leak credentials) so this is expected. I will create an internal PR to validate your changes.

henryju:
I merged your change in #635. Thanks!

PR #636, opened 2023-06-22
New additions to simplify HTTPClient migration

henryju:


henryju:
@kirill-knize-sonarsource @vojtech-suchy-sonarsource when migrating SLE to use the HTTPClient from the backend, I noticed there were a missing feature: fetching organization(s) for an unsaved connection. This is maybe not needed in VSCode, but I let you check.
@nquinquenel I think this is something that has been missed in SLI.

core/src/main/java/org/sonarsource/sonarlint/core/embedded/server/ShowHotspotRequestHandler.java
```diff
@@ -46,24 +47,27 @@
...
 import org.sonarsource.sonarlint.core.commons.TextRange;
+import org.sonarsource.sonarlint.core.repository.connection.ConnectionConfigurationRepository;
 import org.sonarsource.sonarlint.core.serverapi.hotspot.ServerHotspotDetails;
 import org.sonarsource.sonarlint.core.telemetry.TelemetryServiceImpl;
 
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 
-class ShowHotspotRequestHandler implements HttpRequestHandler {
+@Named
+@Singleton
```
nquinquenel, line 57: If we are using Spring, why not use `@Component` instead of `@Named` and `@Singleton` from Java CDI?

nquinquenel (approved):
LGTM, few questions

core/src/main/java/org/sonarsource/sonarlint/core/embedded/server/ShowHotspotRequestHandler.java
henryju: My idea is to limit Spring coupling as much as possible, in the (very unlikely, but who knows) case we would like to move to another CDI framework.

PR #642, opened 2023-07-03
SLCORE-506 Allow to configure HTTPClient timeouts

henryju:


http/src/main/java/org/sonarsource/sonarlint/core/http/HttpClientProvider.java
```diff
@@ -61,25 +67,47 @@ public HttpClientProvider(String userAgent, @Nullable Path sonarlintUserHome,
...
       .build();
 
     sharedClient.start();
   }
 
+  private static Optional<Timeout> getTimeoutFromSystemProp(String key) {
+    return Optional.ofNullable(System.getProperty(key))
+      .map(s -> {
+        try {
+          return Duration.ofMinutes(Integer.parseInt(s));
```
nquinquenel, line 103: Does not look very clear to me what's expected. Is this supposed to be documented or simply explained on case-by-case troubleshooting?

So you can provide either a number for minutes or a string of pattern `PnDTnHnMn.nS` for the duration parsing?

nquinquenel (approved):
LGTM, just one question!

http/src/main/java/org/sonarsource/sonarlint/core/http/HttpClientProvider.java
henryju: Yeah, for most users it will be minutes, but for tests, I needed to allow smaller durations. I considered to only accept `PnDTnHnMn.nS` but I think this is not very user friendly. WDYT?

nquinquenel: I'm not strongly opinionated on this one, personally, I'd have gone for seconds only. In case it might be beneficial to have smaller granularity (milliseconds), then why not keep it as it is now.

PR #645, opened 2023-07-04
QA should not wait for unit tests to start

henryju:
The QA should not wait for other test jobs to complete.


jblievremont (approved):
LGTM 👍🏻 good catch!

PR #647, opened 2023-07-06
Rework initialize params

henryju:


client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/backend/initialize/FeatureFlagsDto.java
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Optional features toggles of the backend. To ease transition or to accommodate different client needs.
+ */
+public class FeatureFlagsDto {
+
+  private final boolean shouldManageSmartNotifications;
+  @Deprecated
+  // not controllable anymore, it is the backend's responsibility to decide
+  private final boolean taintVulnerabilitiesEnabled;
```
damien-urruty-sonarsource, line 30: Instead of moving this, I think we could remove it

client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/backend/initialize/FeatureFlagsDto.java
```diff
@@ -0,0 +1,69 @@
...
+  private final boolean shouldSynchronizeProjects;
+  private final boolean shouldManageLocalServer;
+  private final boolean enableSecurityHotspots;
+
+  public FeatureFlagsDto(boolean shouldManageSmartNotifications, boolean taintVulnerabilitiesEnabled, boolean shouldSynchronizeProjects, boolean shouldManageLocalServer,
```
damien-urruty-sonarsource, line 35: It's a bit more work but wouldn't it be a good time to move to a capabilities system? Here the clients would still have to set each flag, and we could instead propose a list of opt-ins, with constants or enum values. See https://sonarsource.atlassian.net/browse/SLCORE-474

damien-urruty-sonarsource (approved):
Improvement ideas, else LGTM

client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/backend/initialize/FeatureFlagsDto.java
henryju: You mean replacing booleans with enums? I think that even if adding new flags creates binary incompatibilities, this is not a big deal and clearly "document" new features that each IDE can decide to enable or not.
But we can do that later I guess.

client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/backend/initialize/FeatureFlagsDto.java
henryju: OK, I will merge as is, and we can remove this at the next opportunity.

PR #648, opened 2023-07-07
SLCORE-508 Store resolution for local-only issues

nquinquenel:


core/src/test/java/mediumtest/hotspots/HotspotStatusChangeMediumTests.java
```diff
@@ -69,7 +69,6 @@ void tearDown() throws ExecutionException, InterruptedException {
     }
   }
 
-  @Test
```
damien-urruty-sonarsource, line 72: Any reason why this annotation has been removed? Also below in the file

nquinquenel: It's a mistake, because it has the same test name as the one on `IssueStatusChangeMediumTests`, and this one was failing, so I removed the wrong one (during our tests). Obviously, I'll revert it 😄

nquinquenel:
cc @kirill-knize-sonarsource

core/src/main/java/org/sonarsource/sonarlint/core/local/only/XodusLocalOnlyIssueStore.java
```diff
@@ -0,0 +1,260 @@
...
+      .map(branch -> branch.getLinks(CONFIGURATION_SCOPE_ID_TO_FILES_LINK_NAME))
+      .flatMap(files -> findUniqueAmong(files, PATH_PROPERTY_NAME, filePath))
+      .map(fileToLoad -> fileToLoad.getLinks(XodusLocalOnlyIssueStore.FILE_TO_ISSUES_LINK_NAME))
+      .map(issueEntities -> StreamSupport.stream(issueEntities.spliterator(), false)
+        .map(adapter)
+        .collect(Collectors.toList()))
+      .orElseGet(Collections::emptyList));
+  }
+
+  public void createLocalOnlyIssue(String configurationScopeId, LocalOnlyIssue issue, IssueStatus newStatus) {
```
damien-urruty-sonarsource, line 117: I would suggest `store` or `save` or something else than `create` which seems strange as we don't create it (it is a parameter)

core/src/test/java/org/sonarsource/sonarlint/core/local/only/XodusLocalOnlyIssueStoreTests.java
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  @Test
+  void should_return_empty_when_file_path_unknown() {
+    var issues = store.load("configScopeId", "path");
+
+    assertThat(issues).isEmpty();
+  }
+
+  private void compare_issue_to_saved_one(LocalOnlyIssue savedIssue, LocalOnlyIssue localOnlyIssue, IssueStatus resolution, String comment) {
```
damien-urruty-sonarsource, line 68: Another suggestion, maybe move this method to the bottom and have the tests first ?

core/src/test/java/org/sonarsource/sonarlint/core/local/only/XodusLocalOnlyIssueStoreTests.java
```diff
@@ -0,0 +1,148 @@
...
+    assertThat(savedIssue.getMessage()).isEqualTo(localOnlyIssue.getMessage());
```
damien-urruty-sonarsource, line 69: One thing I use in those kind of situations is `assertThat(savedIssue).usingRecursiveComparison().isEqualTo(localOnlyIssue)`. Even if there is no equals, it will compare field by field

nquinquenel: Thanks, I didn't know this! I added `ignoringFieldsOfTypes(LocalOnlyIssueResolution.class)` because it is added when storing the issue

core/src/test/java/org/sonarsource/sonarlint/core/local/only/XodusLocalOnlyIssueStoreTests.java
nquinquenel: Removed following other comment below

core/src/main/java/org/sonarsource/sonarlint/core/local/only/LocalOnlyIssueStorageService.java
```diff
@@ -0,0 +1,54 @@
...
+import java.nio.file.Path;
+
+public class LocalOnlyIssueStorageService {
+
+  private final Path projectsStorageBaseDir;
+  private final Path workDir;
+  private XodusLocalOnlyIssueStore localOnlyIssueStore;
+
+  public LocalOnlyIssueStorageService(Path backupDir, Path workDir) {
+    projectsStorageBaseDir = backupDir;
```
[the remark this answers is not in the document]
nquinquenel: Naming is not super correct, but we use the `storageRoot` for `projectsStorageBaseDir`, and then it is used for storing the backup file later.

Might be worth to move it to a better location?

core/src/main/java/org/sonarsource/sonarlint/core/issue/IssueServiceImpl.java
```diff
@@ -70,16 +77,26 @@ public CompletableFuture<Void> changeStatus(ChangeIssueStatusParams params) {
...
+            .exceptionally(throwable -> {
+              throw new IssueStatusChangeException(throwable);
+            });
+        }
+        // TODO: Submit to new SQ API endpoint
+        var optionalIssue = localOnlyIssueRepository.findByKey(params.getIssueKey());
+        if (optionalIssue.isPresent()) {
+          var issue = optionalIssue.get();
+          localOnlyIssueStorageService.get().createLocalOnlyIssue(params.getConfigurationScopeId(), issue, params.getNewStatus());
+          return null;
```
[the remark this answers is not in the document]
nquinquenel: This is inside the `.map()` 🤷🏻

PR #653, opened 2023-07-11
Update parent pom

henryju:


damien-urruty-sonarsource:
I think this PR could be closed @henryju? Or updated with the latest parent POM

henryju:
Outdated

PR #661, opened 2023-07-19
SLCORE-512 Synchronize resolved local-only issues with SonarQube

damien-urruty-sonarsource:


kirill-knize-sonarsource (approved):
A bit unfortunate that we send two requests in case of muting with comment. A possible candidate for hardening. Else LGTM.

PR #664, opened 2023-07-20
SLCORE-507 Allow changing status of local-only issue only for SQ 10.2+

damien-urruty-sonarsource:


kirill-knize-sonarsource (approved):
LGTM, thanks!

PR #666, opened 2023-07-21
SLCORE-511 Periodically clean local-only issues store

damien-urruty-sonarsource:


core/src/main/java/org/sonarsource/sonarlint/core/local/only/LocalOnlyIssueStorageService.java
```diff
@@ -33,6 +36,11 @@ public LocalOnlyIssueStorageService(Path storageRoot, Path workDir) {
     this.workDir = workDir;
   }
 
+  @PostConstruct
+  public void purgeOldIssues() {
+    get().purgeIssuesOlderThan(Instant.now().minus(7, ChronoUnit.DAYS));
```
nquinquenel, line 41: Seems like a very short period to me

sophio-japharidze-sonarsource (approved):
I guess we can increase the number of days later

PR #669, opened 2023-07-28
Fix pulling logic for new enabled languages

eray-felek-sonarsource:


thahnen (approved):
LGTM! Thanks for the tests, makes sense as I didn't notice the edge case in the old PR either.

PR #675, opened 2023-08-08
SLCORE-533 Use sonar-plugin-api 10.1 with new taxonomy

jblievremont:


thahnen:
@jblievremont the `server-connection/src/main/proto/sonarlint.proto` is not yet final or is it? I was assuming only the "ws-*.proto" files will be changed when specified alongside SQ / SC?
But the file only now having impacts seems a bit off.

jblievremont:
This is not final yet, hence the "draft" status on the PR.

rule-extractor/src/main/java/org/sonarsource/sonarlint/core/rule/extractor/SonarLintRuleDefinition.java
```diff
@@ -62,6 +67,15 @@ public SonarLintRuleDefinition(RulesDefinition.Rule rule) {
     this.name = rule.name();
     this.defaultSeverity = IssueSeverity.valueOf(rule.severity());
     this.type = RuleType.valueOf(rule.type().name());
+    var cleanCodeAttributeFromRule = rule.cleanCodeAttribute();
+    this.cleanCodeAttribute = cleanCodeAttributeFromRule == null ?
+      // TODO Should we keep this as a transition feature? When do we drop this default value?
```
nquinquenel, line 72: What is the goal of this default value? On the client side, I would expect a null value in the case of old CCT

jblievremont: IIRC this was to get non-null values returned while we wait for plugins to be updated. Probably we can remove this now.

thahnen: Hi @jblievremont, can you remove the default values maybe already tomorrow? Would help me (at least) a lot with testing ^^

thahnen: Merci 🙇‍♂️

server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/storage/XodusServerIssueStore.java
```diff
@@ -634,6 +652,9 @@ private static void updateTaintIssueEntity(ServerTaintIssue issue, Entity issueE
     if (ruleDescriptionContextKey != null) {
       issueEntity.setProperty(RULE_DESCRIPTION_CONTEXT_KEY_PROPERTY_NAME, ruleDescriptionContextKey);
     }
+    issueEntity.setBlob(FLOWS_BLOB_NAME, toProtoFlow(issue.getFlows()));
```
nquinquenel, line 655: This is already set few lines above, to be removed

nquinquenel (commented):
Few remarks but LGTM (and waiting for the TODO to be done 😄)

thahnen (commented):
LGTM! I'm awaiting the last changes with the TODOs, then I'll add a final review ^^

commons/src/main/java/org/sonarsource/sonarlint/core/commons/ImpactSeverity.java
```diff
@@ -0,0 +1,36 @@
...
+  MEDIUM("Medium"),
+  HIGH("High");
+
+  private final String displayLabel;
+
+  ImpactSeverity(String displayLabel) {
+    this.displayLabel = displayLabel;
+  }
+
+  public String getDisplayLabel() {
```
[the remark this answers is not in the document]
jblievremont: I decided to keep it, for consistency and accessibility - e.g to set the alt-text (and maybe on-hover text) for the icons.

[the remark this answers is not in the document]
thahnen: Yeah, that's also fine with me 😄

nquinquenel (approved):

PR #679, opened 2023-08-10
SLCORE-537 Add tests for ConfigurationScopeAddedEvent; Fix code smells

sophio-japharidze-sonarsource:


core/src/main/java/org/sonarsource/sonarlint/core/websocket/WebSocketService.java
```diff
@@ -119,10 +117,21 @@ public void handleEvent(ConnectionConfigurationAddedEvent connectionConfiguratio
...
+      connectionIdsInterestedInNotifications.remove(updatedConnectionId);
+      if (!connection.isDisableNotifications()){
+        // Notifications are enabled, reopen the connection
+        createConnectionIfNeeded(updatedConnectionId);
+        resubscribeAll();
+      } else if(!connectionIdsInterestedInNotifications.isEmpty()) {
+        var configurationScopesToUnsubscribe = configurationRepository.getConfigScopesWithBindingConfiguredTo(updatedConnectionId);
+        for(var configScope : configurationScopesToUnsubscribe) {
+          subscribedProjectKeysByConfigScopes.remove(configScope.getId());
+        }
```
damien-urruty-sonarsource, line 132: I wonder if we shouldn't do this at another moment too. If there is only a single SC connection, and if it was updated to disable notifications, we wouldn't reach this branch, and the configurations copes would still be in the `subscribedProjectKeysByConfigScopes`, which might cause problems later when adding a new SC connection

damien-urruty-sonarsource (approved):

PR #683, opened 2023-08-15
SLCORE-537 recreate connection every 2 hours

sophio-japharidze-sonarsource:
Current implementation will simply perform a check every 2 hours and

- if there is already a connection AND, there is a connection that is still interested in receiving notifications, close it, recreate it, and resubscribe for all projects

Subsequently, if a connection was created at 118th minute for example, in one minute it will be closed and reopened. another approach could be to store connection creation time somewhere and refresh only after 2 hours since connection was created, but this adds complexity without too much value gain IMO.


core/src/main/java/org/sonarsource/sonarlint/core/websocket/WebSocketService.java
```diff
@@ -61,6 +66,22 @@ public WebSocketService(SonarLintClient client, ConnectionConfigurationRepositor
       .dispatch(QualityGateChangedEvent.class, new ShowSmartNotificationOnQualityGateChangedEvent(client, configurationRepository));
   }
 
+  @PostConstruct
+  public void initialize() {
+    ScheduledExecutorService webSocketConnectionRefresher = Executors.newSingleThreadScheduledExecutor(r -> new Thread(r, "WebSocket connection refresher"));
+    webSocketConnectionRefresher.scheduleAtFixedRate(this::refreshConnectionIfNeeded, 119, 119, TimeUnit.MINUTES);
```
damien-urruty-sonarsource, line 72: Raising another option as it was not described in the PR summary: we could schedule the task in the `createConnectionIfNeeded` method, so we know the task runs exactly 119min after a connection was opened. But it means you might need to cancel the previous task. So a bit more complicated than this implementation

core/src/main/java/org/sonarsource/sonarlint/core/websocket/WebSocketService.java
```diff
@@ -61,6 +66,22 @@ public WebSocketService(SonarLintClient client, ConnectionConfigurationRepositor
       .dispatch(QualityGateChangedEvent.class, new ShowSmartNotificationOnQualityGateChangedEvent(client, configurationRepository));
   }
 
+  @PostConstruct
+  public void initialize() {
+    ScheduledExecutorService webSocketConnectionRefresher = Executors.newSingleThreadScheduledExecutor(r -> new Thread(r, "WebSocket connection refresher"));
```
damien-urruty-sonarsource, line 71: You might need to store this in a field and properly shut it down when the service is shutdown.

core/src/main/java/org/sonarsource/sonarlint/core/websocket/SonarCloudWebSocket.java
```diff
@@ -46,12 +47,17 @@ public class SonarCloudWebSocket {
...
-  public static SonarCloudWebSocket create(HttpClient httpClient, Consumer<ServerEvent> serverEventConsumer) {
+  public static SonarCloudWebSocket create(HttpClient httpClient, Consumer<ServerEvent> serverEventConsumer, Runnable connectionRefresher) {
     var webSocket = new SonarCloudWebSocket();
     webSocket.ws = httpClient.createWebSocketConnection(WEBSOCKET_DEV_URL, rawEvent -> webSocket.handleRawMessage(rawEvent, serverEventConsumer));
-    webSocket.messageHistoryCleaner.scheduleAtFixedRate(webSocket::cleanUpMessageHistory, 0, 1, TimeUnit.MINUTES);
+    webSocket.sonarCloudWebSocketScheduler.scheduleAtFixedRate(webSocket::cleanUpMessageHistory, 0, 1, TimeUnit.MINUTES);
+    if(scheduledConnectionRefresh != null) {
+      scheduledConnectionRefresh.cancel(true);
+    }
+    scheduledConnectionRefresh = webSocket.sonarCloudWebSocketScheduler.scheduleAtFixedRate(connectionRefresher, 119, 119, TimeUnit.MINUTES);
```
damien-urruty-sonarsource, line 60: What about scheduling it for a single run instead of fixed rate? Anyway when the task is run, the socket will be closed and the executor shutdown. This way there is no need to cancel the recurring task

sophio-japharidze-sonarsource: Yes, I can schedule it for one time in 119 minutes, but I guess cancellation is still needed? In case connection creation is called during the 119 minutes, we would reschedule to start 119 minutes after the newly created time

damien-urruty-sonarsource (approved):
LGTM!

PR #684, opened 2023-08-15
SLCORE-538 do not show polled notification for SonarCloud QG_CHANGED event

sophio-japharidze-sonarsource:


core/src/main/java/org/sonarsource/sonarlint/core/smartnotifications/SmartNotifications.java
```diff
@@ -110,10 +114,12 @@ private void manageNotificationsForConnection(ServerApi serverApi, Map<String, S
       var notifications = retrieveServerNotifications(developersApi, projectKeysByLastEventPolling);
 
       for (var n : notifications) {
-        var smartNotification = new ShowSmartNotificationParams(n.message(), n.link(), scopeIdsPerProjectKey.get(n.projectKey()),
-          n.category(), connectionId);
-        client.showSmartNotification(smartNotification);
-        telemetryService.smartNotificationsReceived(n.category());
+        if (connection != null && connection.getKind() == ConnectionKind.SONARQUBE && !sonarCloudEventsToIgnore.contains(n.category())) {
```
damien-urruty-sonarsource, line 117: Not sure about this condition, shouldn't the last part be a OR?

damien-urruty-sonarsource: Also the last check wouldn't work, the category is different from the event type

damien-urruty-sonarsource (approved):

PR #685, opened 2023-08-16
SLCORE-542 websocket keepalive

sophio-japharidze-sonarsource:


damien-urruty-sonarsource (approved):
LGTM, it will be easier to add a test when we switch to medium tests

PR #686, opened 2023-08-16
Skip downloading the license plugin

damien-urruty-sonarsource:


damien-urruty-sonarsource:
This was just a test to check if the SC license plugin is really mandatory, seems it's not. Closing

PR #687, opened 2023-08-17
Re-connect web socket when closed by server or on error

damien-urruty-sonarsource:


core/src/main/java/org/sonarsource/sonarlint/core/websocket/SonarCloudWebSocket.java
```diff
@@ -50,7 +50,7 @@ public class SonarCloudWebSocket {
 
   public static SonarCloudWebSocket create(HttpClient httpClient, Consumer<ServerEvent> serverEventConsumer, Runnable connectionRefresher) {
     var webSocket = new SonarCloudWebSocket();
-    webSocket.ws = httpClient.createWebSocketConnection(WEBSOCKET_DEV_URL, rawEvent -> webSocket.handleRawMessage(rawEvent, serverEventConsumer));
+    webSocket.ws = httpClient.createWebSocketConnection(WEBSOCKET_DEV_URL, rawEvent -> webSocket.handleRawMessage(rawEvent, serverEventConsumer), connectionRefresher);
```
sophio-japharidze-sonarsource, line 53: The `connectionRefresher` will first try to close the websocket connection and then reopen it. So at some point it might try to `sendClose()` to an already closed websocket. Maybe we need to adjust the `close()` method in this class

core/src/main/java/org/sonarsource/sonarlint/core/websocket/SonarCloudWebSocket.java
```diff
@@ -115,7 +115,10 @@ private static Optional<? extends ServerEvent> parse(WebSocketEvent event) {
 
   public void close() {
     if (this.ws != null) {
-      this.ws.sendClose(WebSocket.NORMAL_CLOSURE, "");
+      // output could already be closed if an error occurred
+      if (!this.ws.isOutputClosed()) {
```
sophio-japharidze-sonarsource, line 119: Could these two if's be grouped together with an && ?

sophio-japharidze-sonarsource: OK, let's keep it this way

PR #688, opened 2023-08-17
SLCORE-541 QualityGateChangedEvent parsing

sophio-japharidze-sonarsource:


core/src/main/java/org/sonarsource/sonarlint/core/websocket/SonarCloudWebSocket.java
```diff
@@ -100,8 +100,10 @@ private void handleRawMessage(String message, Consumer<ServerEvent> serverEventC
   }
 
   private static Optional<? extends ServerEvent> parse(WebSocketEvent event) {
-    var eventType = event.eventType;
-    if (!parsersByType.containsKey(eventType)) {
+    var eventType = event.event;
+    if (eventType == null){
+      return Optional.empty();
```
sophio-japharidze-sonarsource, line 105: I separated these two checks because if the received message is not an event we want to simply ignore and not log an error.

core/src/main/java/org/sonarsource/sonarlint/core/websocket/SonarCloudWebSocket.java
```diff
@@ -76,7 +76,7 @@ public void unsubscribe(String projectKey) {
   }
 
   private void send(String messageType, String projectKey) {
-    var unsubscribePayload = new WebSocketEventSubscribePayload(messageType, "QualityGateChanged", projectKey);
+    var unsubscribePayload = new WebSocketEventSubscribePayload(messageType, new String[]{"QualityGateChanged"}, projectKey);
```
damien-urruty-sonarsource, line 79: I notice now that there is a map in this class from event type to parsers. Maybe we could use the map keys instead of duplicating the event type literal?

damien-urruty-sonarsource (approved):
One minor comment, else LGTM

PR #689, opened 2023-08-17
Show smart notification from poll only if web socket is closed

damien-urruty-sonarsource:


core/src/main/java/org/sonarsource/sonarlint/core/smartnotifications/SmartNotifications.java
```diff
@@ -126,6 +129,11 @@ private void manageNotificationsForConnection(ServerApi serverApi, Map<String, S
     }
   }
 
+  private boolean shouldShowNotification(AbstractConnectionConfiguration connection, ServerNotification n) {
```
sophio-japharidze-sonarsource, line 132: Could we reverse this to make it `intentional | clear` 😂 
method `shouldIgnoreNotification()` and condition would be `connection.getKind() == ConnectionKind.SONARCLOUD ||
      webSocketService.hasOpenConnection() || Objects.equals(n.category(), "QUALITY_GATE")`

sophio-japharidze-sonarsource (approved):

PR #690, opened 2023-08-18
Use medium tests to verify the web sockets functionality

damien-urruty-sonarsource:


core/src/main/java/org/sonarsource/sonarlint/core/websocket/SonarCloudWebSocket.java
```diff
@@ -38,8 +38,10 @@
 import org.sonarsource.sonarlint.core.websocket.parsing.QualityGateChangedEventParser;
 
 public class SonarCloudWebSocket {
-  public static final String WEBSOCKET_DEV_URL = "wss://squad-5-events-api.sc-dev.io/";
-  public static final String WEBSOCKET_URL = "wss://events-api.sonarcloud.io/";
+
+  public static String getUrl() {
+    return System.getProperty("sonarlint.internal.sonarcloud.websocket.url", "wss://events-api.sonarcloud.io/");
```
sophio-japharidze-sonarsource, line 43: cool! from where do we usually set this kind of properties?

core/src/test/java/org/sonarsource/sonarlint/core/websocket/WebSocketServiceTest.java
```diff
@@ -59,9 +59,9 @@
 import static org.mockito.Mockito.verifyNoInteractions;
 import static org.mockito.Mockito.verifyNoMoreInteractions;
 import static org.mockito.Mockito.when;
-import static org.sonarsource.sonarlint.core.websocket.SonarCloudWebSocket.WEBSOCKET_DEV_URL;
 
 class WebSocketServiceTest {
+  private static final String WEBSOCKET_DEV_URL = "wss://events-api.sonarcloud.io/";
```
sophio-japharidze-sonarsource, line 64: this is prod url. is it intentional?

damien-urruty-sonarsource: This class will eventually disappear when tests are migrated to medium ones in this PR

core/src/main/java/org/sonarsource/sonarlint/core/websocket/SonarCloudWebSocket.java
damien-urruty-sonarsource: In your case you can insert `-Dsonarlint.internal.sonarcloud.websocket.url=blah` in the `java` command starting the LS

core/src/main/java/org/sonarsource/sonarlint/core/websocket/WebSocketService.java
```diff
@@ -45,135 +48,217 @@
...
-    var connection = connectionId != null ? connectionConfigurationRepository.getConnectionById(connectionId) : null;
-    var isSonarCloudConnection = connection != null && connection.getKind().equals(ConnectionKind.SONARCLOUD);
-    var projectKey = subscribedProjectKeysByConfigScopes.remove(bindingConfigChangedEvent.getNewConfig().getConfigScopeId());
-    if (projectKey != null && !subscribedProjectKeysByConfigScopes.containsValue(projectKey)) {
-      unsubscribe(projectKey);
-    }
-    if (newProjectKey != null && isSonarCloudConnection && !connection.isDisableNotifications()) {
-      createConnectionIfNeeded(connectionId);
-      subscribe(configScopeId, newProjectKey);
+    if (!shouldEnableWebSockets) {
```
sophio-japharidze-sonarsource, line 81: this check looks a bit annoying in every method. Is it only for limiting which tests initialize websocket connection?

core/src/test/java/mediumtest/websockets/WebSocketMediumTests.java
```diff
@@ -0,0 +1,726 @@
...
+
+      backend.getConnectionService().didUpdateConnections(new DidUpdateConnectionsParams(emptyList(), emptyList()));
+
+      await().atMost(Duration.ofSeconds(2)).untilAsserted(() -> assertThat(webSocketServer.getConnections())
+        .extracting(WebSocketConnection::isOpened, WebSocketConnection::getReceivedMessages)
+        .containsExactly(tuple(false, List.of("{\"action\":\"subscribe\",\"events\":[\"QualityGateChanged\"],\"project\":\"projectKey\"}"))));
+    }
+
+    @Test
+    void should_not_close_connection_if_another_sonarcloud_connection_is_active() {
```
sophio-japharidze-sonarsource, line 423: Does this test belong to the following nested class - `WhenConnectionUpdated`?

sophio-japharidze-sonarsource (approved):

core/src/main/java/org/sonarsource/sonarlint/core/websocket/WebSocketService.java
damien-urruty-sonarsource: I don't like it either. The service works exclusively in reaction to events, so the only way I found to respect the flag is to check at the beginning of each `handle` method. The event bus implementation only allows us to use annotations to subscribe. Maybe we should find another implementation that lets us subscribe programmatically, this way we could have the check in a single place

core/src/test/java/mediumtest/websockets/WebSocketMediumTests.java
damien-urruty-sonarsource: No because we remove a connection in the test, so WhenConnectionRemoved is right here

PR #691, opened 2023-08-22
SLCORE-536 Synchronize Clean Code fields for taint vulnerabilities

jblievremont:


server-api/src/main/proto/sonarqube/ws-issues.proto
```diff
@@ -56,6 +56,11 @@ message Issue {
   optional string creationDate = 23;
   optional sonarqube.ws.commons.RuleType type = 27;
   optional string ruleDescriptionContextKey = 37;
+  // skipping unused fields 38-39
+  optional sonarqube.ws.commons.CleanCodeAttribute cleanCodeAttribute = 40;
+  optional sonarqube.ws.commons.CleanCodeAttributeCategory cleanCodeAttributeCategory = 41;
```
damien-urruty-sonarsource, line 61: Is this used at all? I don't see it used when handling API result. Same for the `TaintVulnerabilityLite` message

server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/TaintIssueDownloader.java
```diff
@@ -124,13 +124,13 @@ private static ServerTaintIssue convertTaintVulnerability(SourceApi sourceApi, I
...
     }
     var ruleDescriptionContextKey = taintVulnerabilityFromWs.hasRuleDescriptionContextKey() ? taintVulnerabilityFromWs.getRuleDescriptionContextKey() : null;
-    // TODO Parse actual values from stream
-    var cleanCodeAttribute = CleanCodeAttribute.TRUSTWORTHY;
-    var impacts = Map.of(
-      SoftwareQuality.RELIABILITY, ImpactSeverity.LOW,
-      SoftwareQuality.MAINTAINABILITY, ImpactSeverity.MEDIUM,
-      SoftwareQuality.SECURITY, ImpactSeverity.HIGH
-    );
+    var cleanCodeAttribute = taintVulnerabilityFromWs.hasCleanCodeAttribute() ? CleanCodeAttribute.valueOf(taintVulnerabilityFromWs.getCleanCodeAttribute().name()) : null;
```
damien-urruty-sonarsource, line 127: Not sure when the WS could return `UNKNOWN_ATTRIBUTE` (0) but I think `valueOf` would throw here. The same goes for the other 5 `valueOf` added calls below

server-api/src/main/proto/sonarqube/ws-issues.proto
jblievremont: Indeed, this field is not used by SLCORE, I believe we can safely skip it.

server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/TaintIssueDownloader.java
jblievremont: Following our discussion, I will make sure that we treat `UNKNOWN_XXX` enum values as `null` where appropriate.

server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/TaintIssueDownloader.java
```diff
@@ -179,13 +203,13 @@ private static ServerTaintIssue convertLiteTaintIssue(TaintVulnerabilityLite lit
...
-      SoftwareQuality.MAINTAINABILITY, ImpactSeverity.MEDIUM,
-      SoftwareQuality.SECURITY, ImpactSeverity.HIGH
-    );
+    var cleanCodeAttribute = liteTaintIssueFromWs.hasCleanCodeAttribute() ? CleanCodeAttribute.valueOf(liteTaintIssueFromWs.getCleanCodeAttribute().name()) : null;
+    var impacts = liteTaintIssueFromWs.getImpactsList().stream()
+      .map(i -> Map.entry(
+        SoftwareQuality.valueOf(i.getSoftwareQuality().name()),
+        ImpactSeverity.valueOf(i.getSeverity().name())
+      ))
+      .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
```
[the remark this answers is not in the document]
jblievremont: It would have helped to flag each instance separately.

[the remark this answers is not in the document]
damien-urruty-sonarsource: I mentioned the 5 `valueOf` calls in the previous comment but yes next time I will add separate comments

damien-urruty-sonarsource (approved):

PR #694, opened 2023-08-24
SLCORE-546 Let users reopen issues existing on the server

nquinquenel:


client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/backend/issue/ResolutionStatus.java
```diff
@@ -19,15 +19,15 @@
  */
 package org.sonarsource.sonarlint.core.clientapi.backend.issue;
 
-public enum IssueStatus {
+public enum ResolutionStatus {
```
damien-urruty-sonarsource, line 22: Be aware that this will break the clients' update of the CORE. I think it's ok, not a big change, but it might be good to communicate the changes to all parties

core/src/main/java/org/sonarsource/sonarlint/core/issue/IssueServiceImpl.java
```diff
@@ -178,17 +176,36 @@ public CompletableFuture<Void> addComment(AddIssueCommentParams params) {
...
+    var optionalBinding = configurationRepository.getEffectiveBinding(configurationScopeId);
+    return optionalBinding
+      .flatMap(effectiveBinding -> serverApiProvider.getServerApi(effectiveBinding.getConnectionId()))
+      .map(connection -> {
+        var binding = optionalBinding.get();
+        var projectServerIssueStore = storageService.binding(binding).findings();
+        var issueId = params.getIssueId();
+        boolean isServerIssue = projectServerIssueStore.containsIssue(issueId, false);
+        if (isServerIssue) {
+          return connection.issue().changeStatusAsync(issueId, Transition.REOPEN)
```
damien-urruty-sonarsource, line 189: Nitpick: The method starts to become big. I would suggest extracting this branch and the else to two new methods, something like `reopenLocalIssue` and `reopenServerIssue`

damien-urruty-sonarsource (commented):
Small comments, else LGTM

damien-urruty-sonarsource (approved):

core/src/main/java/org/sonarsource/sonarlint/core/issue/IssueServiceImpl.java
nquinquenel: Good point!

PR #701, opened 2023-08-30
SLCORE-520 Suppress "skipping section" debug log

nquinquenel:


commons/src/test/java/org/sonarsource/sonarlint/core/commons/log/LogOutputDelegatorTests.java
```diff
@@ -34,7 +35,7 @@ class LogOutputDelegatorTests {
 
   @Test
   void should_not_throw_exception_when_not_set() {
-    delegator.log("asd", Level.DEBUG);
+    assertDoesNotThrow(() -> delegator.log("asd", Level.DEBUG));
```
henryju, line 38: :+1:

PR #702, opened 2023-08-30
SLCORE-449 Warn users connected to a soon unsupported SQ version

nquinquenel:


core/src/main/java/org/sonarsource/sonarlint/core/VersionSoonUnsupportedHelper.java
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  @Subscribe
+  public void configurationScopesAdded(ConfigurationScopesAddedEvent event) {
+    var configScopeIds = event.getAddedConfigurationScopeIds();
+    notifyIfSoonUnsupported(configScopeIds);
+  }
+
+  @Subscribe
+  public void bindingConfigChanged(BindingConfigChangedEvent event) {
```
henryju, line 67: When changing the binding of multiple projects in a row (quite common in SLE), the client will receive many notifications for the same connection...

core/src/main/java/org/sonarsource/sonarlint/core/VersionSoonUnsupportedHelper.java
```diff
@@ -0,0 +1,113 @@
...
+    this.client = client;
+    this.configRepository = configRepository;
+    this.connectionRepository = connectionRepository;
+    this.serverApiProvider = serverApiProvider;
+  }
+
+  @Subscribe
+  public void configurationScopesAdded(ConfigurationScopesAddedEvent event) {
+    var configScopeIds = event.getAddedConfigurationScopeIds();
+    notifyIfSoonUnsupported(configScopeIds);
```
henryju, line 63: When adding many config scopes in a row (for example when opening multiple projects in SLE workspace), the client will receive many notifications (one per project), while I would expect to receive only one for the connection.

core/src/main/java/org/sonarsource/sonarlint/core/VersionSoonUnsupportedHelper.java
```diff
@@ -0,0 +1,113 @@
...
+    configScopeIds.forEach(configScopeId -> {
+      var effectiveBinding = configRepository.getEffectiveBinding(configScopeId);
+      if (effectiveBinding.isPresent()) {
+        var connectionId = effectiveBinding.get().getConnectionId();
+        notifyIfSoonUnsupported(configScopeId, connectionId);
+      }
+    });
+  }
+
+  private void notifyIfSoonUnsupported(String configScopeId, String connectionId) {
```
henryju, line 85: At this point, we don't know if a notification will be sent, so I would rename the method to something like: `checkIfSoonUnsupported`

core/src/test/java/mediumtest/http/ProxyMediumTests.java
```diff
@@ -73,19 +73,19 @@ class ProxyMediumTests {
   @BeforeEach
   void configureProxy(TestInfo info) {
     if (info.getTags().contains(PROXY_AUTH_ENABLED)) {
-      proxyMock.stubFor(get(urlMatching(".*"))
+      proxyMock.stubFor(get(urlMatching("/api/rules/.*"))
```
henryju, line 76: Unrelated change?

henryju (changes requested):
I would ensure to only check and send a notification once for a given connection. Waiting for a true connection state component, maybe add a cache in the VersionSoonUnsupportedHelper to remember that the check has already been done for this connection id.

core/src/test/java/mediumtest/http/ProxyMediumTests.java
nquinquenel: By allowing everything, it sends a request to `/api/system/status` due to the changes I made. This means the tests fail when verifying if `/api/rules/show` got requested.
The easy and quick fix was to allow only `/api/rules` because we do not need anything else in this test

core/src/main/java/org/sonarsource/sonarlint/core/VersionSoonUnsupportedHelper.java
nquinquenel: If the user mutes the first notification when doing the first update, he won't receive any after that

IMO this is expected behaviour, this event is not in a 'batch' so I cannot know right away if the events are linked

core/src/main/java/org/sonarsource/sonarlint/core/VersionSoonUnsupportedHelper.java
nquinquenel: Good point, I will gather the same connections together

core/src/test/java/mediumtest/http/ProxyMediumTests.java
nquinquenel: Following the main comment that I didn't see, caching could be done but then, in case I switch from `Connection_A` -> `Connection_B` -> `Connection_A`, I would expect to see the message again for `Connection_A`?

core/src/main/java/org/sonarsource/sonarlint/core/VersionSoonUnsupportedHelper.java
nquinquenel: Following the main comment that I didn't see, caching could be done but then, in case I switch from `Connection_A` -> `Connection_B` -> `Connection_A`, I would expect to see the message again for `Connection_A`, WDYT?

henryju: What about keeping a Set of connection id for which a notification was already sent? This won't survive a restart of the backend, but at least this would avoid checking many time the same connection?

core/src/main/java/org/sonarsource/sonarlint/core/VersionSoonUnsupportedHelper.java
```diff
@@ -0,0 +1,119 @@
...
+    if (connection != null && connection.getKind() == ConnectionKind.SONARQUBE) {
+      serverApiProvider.getServerApi(connectionId).ifPresent(serverApi -> serverApi.system().getStatus()
+          .thenApply(ServerInfo::getVersion)
+          .thenApply(Version::create)
+          .thenApply(VersionUtils::isVersionSupportedDuringGracePeriod)
+          .thenAccept(isSupported -> {
+            if (Boolean.TRUE.equals(isSupported)) {
+              client.showSoonUnsupportedMessage(
+                new ShowSoonUnsupportedMessageParams(
+                  String.format(UNSUPPORTED_NOTIFICATION_ID, connectionId),
```
damien-urruty-sonarsource, line 102: If I understand the ID would be `sonarlint.unsupported.My Connection.id`. There is nothing related to the version itself in the ID. So the following scenario could happen:
* users use a soon-to-be-dropped version, so we send them a notification
* users decide to not show it again
* their SQ version is updated to X.Y
* on a future SL version we mark the X.Y version to be soon dropped
* users won't receive the notification because the ID would be the same and they already dismissed it

I am not sure a SL connection would live long enough for this to happen though

core/src/main/java/org/sonarsource/sonarlint/core/VersionSoonUnsupportedHelper.java
nquinquenel: But if the user switches to the same connection multiple times during his session, and if he didn't mute the notification on the IDE side, maybe he would like to see it multiple times?

If the responsibility of 'one-time' is moved to the core, then we should the notification one time on client side and mute it by default after that, but then he maybe not the best UX because user might miss it

server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/VersionUtils.java
```diff
@@ -19,16 +19,37 @@
...
+  private static final Instant CURRENT_LTS_RELEASE_DATE = ZonedDateTime.of(2023, 2, 1, 0, 0, 0, 0, ZoneId.systemDefault()).toInstant();
   private static final String VERSION_REGEX = ".*?(\\d+\\.\\d+(?:\\.\\d+)*).*";
   private static final Pattern JAR_VERSION_PATTERN = Pattern.compile(VERSION_REGEX);
 
   private VersionUtils() {
   }
 
+  public static boolean isVersionSupportedDuringGracePeriod(Version currentVersion) {
+    return currentVersion.compareTo(CURRENT_LTS) < 0 && currentVersion.compareToIgnoreQualifier(PREVIOUS_LTS) >= 0
+      && ZonedDateTime.now().minusYears(1).toInstant().compareTo(CURRENT_LTS_RELEASE_DATE) < 0;
```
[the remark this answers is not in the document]
nquinquenel: mmh good point, following [documentation](https://docs.sonarsource.com/sonarlint/intellij/team-features/connected-mode/#sonarlint-sonarqube-version-support-policy), _we still enable connecting SonarLint to the previous LTS version for a certain period of time (currently 12 months after the latest LTS release)_

It's a bit contradictory? 🤔

core/src/main/java/org/sonarsource/sonarlint/core/VersionSoonUnsupportedHelper.java
nquinquenel: Indeed, I knew something was missing 😅

server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/VersionUtils.java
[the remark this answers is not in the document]
damien-urruty-sonarsource: This is the general policy but we have not been very responsive in implementing it (this [ticket](https://sonarsource.atlassian.net/browse/SLCORE-446) is opened for 6 months and I think at that time the 7.9 was already eligible for removal)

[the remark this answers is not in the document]
damien-urruty-sonarsource: So maybe the first soon-to-be-dropped version should be 7.9 and the next minimal supported version 8.9? To not drop a version out of the blue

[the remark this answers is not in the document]
nquinquenel: I made the switch to 7.9 👍🏻

core/src/main/java/org/sonarsource/sonarlint/core/VersionSoonUnsupportedHelper.java
henryju: In this case the user will receive the notification the next time SonarLint starts, I don't think it is a big deal here. But do as you want.

nquinquenel: I switched to use the storage to retrieve the version, so it avoids making a lot of unnecessary calls to the server. The downside is that it might be unsynchronized if the server is updated but that's a very rare case and only last for an hour at most.

Then I also switch to an in-memory cache for the connection/version already seen, which is cleared at each restart.

client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/client/message/ShowSoonUnsupportedMessageParams.java
```diff
@@ -0,0 +1,72 @@
...
+ * </ul>
+ */
+public class ShowSoonUnsupportedMessageParams {
+
+  @NonNull
+  private final String notificationId;
+  @NonNull
+  private final String configurationScopeId;
+  @NonNull
+  private final MessageType type;
```
henryju, line 46: Since we don't have a generic notification, but a specific one, what is the point of this `type`? It will always be WARNING at the moment, right?

client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/client/message/ShowSoonUnsupportedMessageParams.java
```diff
@@ -0,0 +1,72 @@
...
+ * <ul>
+ *   <li>"ERROR"</li>
+ *   <li>"WARNING"</li>
+ *   <li>"INFO"</li>
+ * </ul>
+ */
+public class ShowSoonUnsupportedMessageParams {
+
+  @NonNull
+  private final String notificationId;
```
henryju, line 42: I wonder if we could not find a more informative name than `notificationId`? What about `doNotShowAgainId`?

henryju (approved):
Minor comments then LGTM

client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/client/message/ShowSoonUnsupportedMessageParams.java
nquinquenel: Right, removed + added explanation in java doc

PR #703, opened 2023-08-31
SLCORE-497 Move SSE management to the new backend

damien-urruty-sonarsource:


client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/SonarLintClient.java
```diff
@@ -157,4 +158,12 @@ default CompletableFuture<GetProxyPasswordAuthenticationResponse> getProxyPasswo
   default CompletableFuture<CheckServerTrustedResponse> checkServerTrusted(CheckServerTrustedParams params) {
     return CompletableFuture.completedFuture(new CheckServerTrustedResponse(false));
   }
+  /**
+   * @deprecated More precise notifications will eventually be sent to clients, like didUpdateTaintVulnerabilities to be used directly for display
```
henryju, line 162: Maybe wait for the new methods to be available before marking it as deprecated?

core/src/test/java/mediumtest/server/events/ServerSentEventsMediumTests.java
```diff
@@ -0,0 +1,556 @@
...
+        .withEnabledLanguageInStandaloneMode(Language.JS)
+        .withExtraEnabledLanguagesInConnectedMode(Language.JAVA)
+        .withServerSentEventsEnabled()
+        .withSonarCloudConnection("connectionId")
+        .withUnboundConfigScope("configScope")
+        .build();
+
+      bind("configScope", "connectionId", "projectKey");
+
+      await().pollDelay(Duration.ofMillis(200)).atMost(Duration.ofSeconds(2))
```
henryju, line 116: I am not sure about this assertion. Waiting for something to not happen is always tricky. Maybe something like:
```
Thread.sleep(200); // Give time for the SSE service to start
assertThat(requestedPaths()).isEmpty();
```

core/src/test/java/mediumtest/server/events/ServerSentEventsMediumTests.java
```diff
@@ -0,0 +1,556 @@
...
+        .withExtraEnabledLanguagesInConnectedMode(Language.JAVA)
+        .withServerSentEventsEnabled()
+        .withSonarQubeConnection("connectionId", sonarServerMock.baseUrl())
+        .withBoundConfigScope("configScope", "connectionId", "projectKey")
+        .build();
+      await().atMost(Duration.ofSeconds(2)).untilAsserted(() -> assertThat(requestedPaths()).hasSize(1));
+
+      bind("configScope", "connectionId", "projectKey");
+
+      await().pollDelay(Duration.ofMillis(200)).atMost(Duration.ofSeconds(2)).untilAsserted(() -> assertThat(requestedPaths()).hasSize(1));
```
henryju, line 137: Same

core/src/test/java/mediumtest/server/events/ServerSentEventsMediumTests.java
```diff
@@ -0,0 +1,556 @@
...
+        .withExtraEnabledLanguagesInConnectedMode(Language.JAVA)
+        .withServerSentEventsEnabled()
+        .withSonarQubeConnection("connectionId", sonarServerMock.baseUrl())
+        .withBoundConfigScope("configScope", "connectionId", "projectKey")
+        .build();
+      await().atMost(Duration.ofSeconds(2)).untilAsserted(() -> assertThat(requestedPaths()).hasSize(1));
+
+      unbind("configScope");
+
+      await().pollDelay(Duration.ofMillis(200)).atMost(Duration.ofSeconds(2)).untilAsserted(() -> assertThat(requestedPaths()).hasSize(1));
```
henryju, line 156: Same, maybe we need a way to assert that all background processes/events have be executed...

henryju (changes requested):
Some comments on the medium test, but overall LGTM

core/src/test/java/mediumtest/server/events/ServerSentEventsMediumTests.java
damien-urruty-sonarsource: I found something better suited for this check: 
> await().during(Duration.ofMillis(300)).until(() -> requestedPaths().isEmpty());

Awaitility will check that the `requestedPaths()` remains empty during 300ms, and will fail if not the case.

core/src/test/java/mediumtest/server/events/ServerSentEventsMediumTests.java
damien-urruty-sonarsource: Yes this could be useful, but for that we would need a way to flush the event bus, and I don't think that's possible with the current implementation. Another reason to switch to something else?

henryju (approved):

PR #710, opened 2023-09-11
MMF-3244 SonarLint users in connected mode have an option to focus on New Code

kirill-knize-sonarsource:


damien-urruty-sonarsource (commented):
Some late comments as I just started testing a bit more and noticed a few things. Happy to discuss further if needed

PR #712, opened 2023-09-12
SLCORE-556 Gather telemetry about the use of "focus on new code"

damien-urruty-sonarsource:


henryju (commented):
Shouldn't we also transmit the initial focus mode to the backend at startup in the InitializeParams? Or do you expect IDEs to call `didFocusChanged` once just after starting the backend (which would be misleading IMO)?

jblievremont (approved):
LGTM 👍🏻 and checked with the Data Team 📜

PR #713, opened 2023-09-13
SLCORE-558 Expose a way for clients to know whether a security hotspot is 'on new code'

damien-urruty-sonarsource:


damien-urruty-sonarsource:
cc FYI @kirill-knize-sonarsource @jblievremont

client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/backend/tracking/ClientTrackedFindingDto.java
```diff
@@ -23,7 +23,7 @@
 import javax.annotation.CheckForNull;
 import javax.annotation.Nullable;
 
-public class ClientTrackedIssueDto {
```
henryju, line 26: FYI, with the new CCT, the only concept left will be "issues". "findings" was considered but not retained. We can keep "finding" as an internal concept for now, as long as we still have hotspots.

core/src/main/java/org/sonarsource/sonarlint/core/tracking/IssueTrackingServiceImpl.java
```diff
@@ -215,7 +215,7 @@ private static Collection<Trackable> toLocalOnlyIssueTrackables(List<LocalOnlyIs
   @PreDestroy
   public void shutdown() {
     if (!MoreExecutors.shutdownAndAwaitTermination(executorService, 1, TimeUnit.SECONDS)) {
-      LOG.warn("Unable to stop binding suggestions executor service in a timely manner");
+      LOG.warn("Unable to stop issue updater executor service in a timely manner");
```
henryju, line 218: Good catch!

core/src/main/java/org/sonarsource/sonarlint/core/tracking/SecurityHotspotMatchingServiceImpl.java
```diff
@@ -0,0 +1,165 @@
...
+    this.configurationRepository = configurationRepository;
+    this.storageService = storageService;
+    this.activeSonarProjectBranchRepository = activeSonarProjectBranchRepository;
+    this.synchronizationService = synchronizationService;
+    this.executorService = Executors.newSingleThreadExecutor(r -> new Thread(r, "sonarlint-server-tracking-hotspot-updater"));
+  }
+
+  @Override
+  public CompletableFuture<MatchWithServerSecurityHotspotsResponse> matchWithServerSecurityHotspots(MatchWithServerSecurityHotspotsParams params) {
+    return CompletableFutures.computeAsync(cancelChecker -> {
```
henryju, line 84: Do we really need to depends on lsp4j cancellation mechanism? Do we support cancellation here?

core/src/main/java/org/sonarsource/sonarlint/core/tracking/SecurityHotspotMatchingServiceImpl.java
```diff
@@ -0,0 +1,165 @@
...
+      var configurationScopeId = params.getConfigurationScopeId();
+      var effectiveBindingOpt = configurationRepository.getEffectiveBinding(configurationScopeId);
+      var activeBranchOpt = activeSonarProjectBranchRepository.getActiveSonarProjectBranch(configurationScopeId);
+      if (effectiveBindingOpt.isEmpty() || activeBranchOpt.isEmpty()) {
```
henryju, line 88: May add a debug log here to know we fall in this case.

core/src/main/java/org/sonarsource/sonarlint/core/tracking/SecurityHotspotMatchingServiceImpl.java
```diff
@@ -0,0 +1,165 @@
...
+      return new MatchWithServerSecurityHotspotsResponse(clientTrackedIssuesByServerRelativePath.entrySet().stream().map(e -> {
+        var serverRelativePath = e.getKey();
+        var serverHotspots = storageService.binding(binding).findings().loadHotspots(activeBranch, serverRelativePath);
+        var clientHotspotTrackables = toTrackables(e.getValue());
+        var matches = matchSecurityHotspots(serverHotspots, clientHotspotTrackables)
+          .stream().<Either<ServerMatchedSecurityHotspotDto, LocalOnlySecurityHotspotDto>>map(result -> {
+            if (result.isLeft()) {
+              var serverSecurityHotspot = result.getLeft();
+              var creationDate = serverSecurityHotspot.getCreationDate();
+              var isOnNewCode = storageService.binding(binding).newCodeDefinition().read().map(definition -> definition.isOnNewCode(creationDate)).orElse(true);
```
henryju, line 110: We can maybe extract the read of the new code definition, to avoid doing it for each finding?

core/src/main/java/org/sonarsource/sonarlint/core/client/api/connected/ConnectedSonarLintEngine.java
```diff
@@ -71,7 +71,9 @@ public interface ConnectedSonarLintEngine extends SonarLintEngine {
    * @param projectBinding information about the project (must have been previously updated with {@link #updateProject(EndpointParams, HttpClient, String, ClientProgressMonitor)})
    * @param filePath       relative to the project.
    * @return All server issues in the local storage for the given file. If file has no issues, an empty list is returned.
+   * @deprecated tracking is managed by the new backend
    */
+  @Deprecated(since = "9.3")
```
henryju, line 76: Is it not too early to deprecate? Do we already have full replacement of the matching/tracking for all type of issues?

henryju (approved):
Very good move! I made a few comments, but overall, LGTM!

client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/backend/tracking/ClientTrackedFindingDto.java
damien-urruty-sonarsource: That's good to know, thanks for sharing!

core/src/main/java/org/sonarsource/sonarlint/core/tracking/SecurityHotspotMatchingServiceImpl.java
damien-urruty-sonarsource: I think it makes sense for this step to be cancelable as the network is involved. lsp4j is not a good way to achieve that but I suggest we keep it and reconsider it when we move to a separate process

core/src/main/java/org/sonarsource/sonarlint/core/tracking/SecurityHotspotMatchingServiceImpl.java
damien-urruty-sonarsource: Yep, good idea!

core/src/main/java/org/sonarsource/sonarlint/core/client/api/connected/ConnectedSonarLintEngine.java
damien-urruty-sonarsource: Your comment is about `getServerIssues()` and yes we have a full replacement for that, since the work on muting local-only issues. It's already not used anymore by SLI and SLLS.

core/src/main/java/org/sonarsource/sonarlint/core/tracking/SecurityHotspotMatchingServiceImpl.java
damien-urruty-sonarsource: This method will also be called in standalone mode, so I'm not sure we should print a trace at the end of each analysis, it's already quite verbose

PR #715, opened 2023-09-14
SLCORE-557 Fix JAR file resource leak

henryju:


henryju:
@damien-urruty-sonarsource can you please review this PR?

plugin-commons/src/main/java/org/sonarsource/sonarlint/core/plugin/commons/loading/PluginInstancesLoader.java
```diff
@@ -169,25 +192,30 @@ Map<String, Plugin> instantiatePluginClasses(Map<PluginClassLoaderDef, ClassLoad
...
+          tryAndCollectIOException(((Closeable) classLoader)::close, exceptions);
         }
       }
+      classloadersToClose.clear();
+    }
+    synchronized (jarFilesToClose) {
+      for (var jarFile : jarFilesToClose) {
+        tryAndCollectIOException(jarFile::close, exceptions);
+      }
+      jarFilesToClose.clear();
```
damien-urruty-sonarsource, line 210: If this is the workaround because of the JDK, a small comment somewhere might be useful with the link to the ticket

plugin-commons/src/main/java/org/sonarsource/sonarlint/core/plugin/commons/loading/PluginInstancesLoader.java
```diff
@@ -85,31 +95,44 @@ public Map<String, Plugin> instantiatePluginClasses(Collection<PluginInfo> plugi
...
   }
 
+  private static Optional<JarFile> getJarFile(Path tmpDepFile) {
+    try {
+      return Optional.of(((JarURLConnection) new URL("jar:" + tmpDepFile.toUri().toURL() + "!/").openConnection()).getJarFile());
+    } catch (ZipException ignore) {
+      // For tests, we are using fake JARs, so ignore ZipException: zip file is empty
+      return Optional.empty();
+    } catch (IOException e) {
+      throw new IllegalArgumentException(e);
```
damien-urruty-sonarsource, line 132: I think this would fail the loading process for all plugins, shouldn't we just ignore the plugin but go on with the rest?

damien-urruty-sonarsource (approved):
LGTM with small suggestions.
Maybe not now but I think we need to define a proper mechanism for errors in the backend, to let users know that something went wrong.

plugin-commons/src/main/java/org/sonarsource/sonarlint/core/plugin/commons/loading/PluginInstancesLoader.java
henryju: Hum, that was not the previous behavior AFAIK. What if a plugin depends on another plugin. We might end up in an inconsistent state.

damien-urruty-sonarsource: I might be wrong but it's a new method, so this is a change in behavior

henryju: My point is that all Jar paths are supposed to exist. If they don't, it would have failed later when building the URLClassloader. So failing here or later is the same.

damien-urruty-sonarsource: The reason for my comment was [this line](https://github.com/SonarSource/sonarlint-core/blob/master/plugin-commons/src/main/java/org/sonarsource/sonarlint/core/plugin/commons/loading/PluginClassloaderFactory.java#L100), I have the impression we are only logging for ClassLoaders that cannot be built. This is something I implemented a while ago but it's maybe not the right way to do that

henryju: I have a similar idea that all classloader definitions could depend on each other. Having a classloader failing to be built, and still trying to build/load others may create unexpected errors.
In the long run, if we feel the need to support the case of one analyzer failing to load, while still being able to run others, I think we should go in a direction where we totally isolate each analyzer into its own container (or even process?).

PR #716, opened 2023-09-15
SLCORE-497 Changes ITs to rely on SSE managed by the backend

henryju:


damien-urruty-sonarsource (approved):
Good move to not rely on the engine anymore. At some point I would like to move some of these ITs to contract tests as we discussed at some point

PR #718, opened 2023-09-19
SLCORE-563 Add support for ProjectSensor

henryju:


analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/container/analysis/sensor/SensorsExecutor.java
```diff
@@ -47,18 +47,18 @@ public class SensorsExecutor {
...
+  public SensorsExecutor(DefaultSensorContext context, SensorOptimizer sensorOptimizer, ProgressMonitor progress, Optional<List<ProjectSensor>> sensors) {
     this.context = context;
     this.sensors = sensors.orElse(List.of());
     this.sensorOptimizer = sensorOptimizer;
     this.progress = progress;
   }
 
   public void execute() {
-    for (Sensor sensor : sort(sensors)) {
+    for (var sensor : sort(sensors)) {
```
damien-urruty-sonarsource, line 61: Isn't there a specific order to run `ProjectSensor`s that are not `Sensor`s? I thought since they are executed once, they should be executed after the other sensors.

damien-urruty-sonarsource (commented):
FYI this made me think of this other [ticket](https://sonarsource.atlassian.net/browse/SLCORE-403) I created when I investigated support for DBD.
Are `ProjectSensor`s a direct replacement for global `Sensor`s? If that's the case then we might also want to execute them at the end (they rely on this logic in DBD if I'm not mistaken)

damien-urruty-sonarsource:
And out of curiosity, what led you to implement this change? Might be good to add a bit of context to the ticket

analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/container/analysis/sensor/SensorsExecutor.java
```diff
@@ -59,9 +59,19 @@ public SensorsExecutor(DefaultSensorContext context, SensorOptimizer sensorOptim
...
 
   public void execute() {
-    var sensorGroups = sensors.stream().collect(Collectors.partitioningBy(Sensor.class::isInstance));
-    var moduleSensors = sensorGroups.get(true);
-    var globalSensors = sensorGroups.get(false);
+    var sensorGroups = sensors.stream().collect(Collectors.partitioningBy(s -> {
+      var isModernGlobalSensor = !(s instanceof Sensor);
+
+      var descriptor = new DefaultSensorDescriptor();
+      s.describe(descriptor);
```
henryju, line 66: You can save the call to this if you already know `isModernGlobalSensor` is true

damien-urruty-sonarsource (approved):
Apart from Julien's comment, LGTM

damien-urruty-sonarsource (approved):

PR #722, opened 2023-09-26
SLCORE-564 Create ShowIssueRequestHandler in order to manage Open issue in IDE

vojtech-suchy-sonarsource:


client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/SonarLintClient.java
```diff
@@ -98,6 +99,12 @@ public interface SonarLintClient {
   @JsonNotification
   void showHotspot(ShowHotspotParams params);
 
+  /**
+   * Sends a notification to the client to show a specific issue (specified by {@link ShowIssueParams} in the IDE
```
damien-urruty-sonarsource, line 103: Missing closing `)`

client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/client/issue/ShowIssueParams.java
```diff
@@ -0,0 +1,85 @@
...
+public class ShowIssueParams {
+  private final String fileUri;
+  private final String message;
+  private final String connectionId;
+  private final String ruleKey;
+  private final String issueKey;
+  private final String creationDate;
+  private final String codeSnippet;
+  private final List<FlowDto> flows;
+  private final TextRangeDto textRange;
```
damien-urruty-sonarsource, line 35: You probably don't need it in VSCode, but here we need to have the configuration scope ID. 
I'm not sure the connectionId is required because if you have the configScopeId you can retrieve the connection on the client side.

client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/client/issue/ShowIssueParams.java
```diff
@@ -0,0 +1,85 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.sonarlint.core.clientapi.client.issue;
+
+import java.util.List;
+import org.sonarsource.sonarlint.core.clientapi.common.FlowDto;
+import org.sonarsource.sonarlint.core.clientapi.common.TextRangeDto;
+
+public class ShowIssueParams {
+  private final String fileUri;
```
damien-urruty-sonarsource, line 27: I am not sure how usable a fileUri would be for us. I need to check. Also there is normally a mapping to do with paths because the server path and the IDE path can be different, not sure we can handle that in the backend today

core/src/test/java/mediumtest/issues/ShowIssueMediumTests.java
```diff
@@ -0,0 +1,79 @@
...
+import org.sonarsource.sonarlint.core.commons.TextRangeWithHash;
+import org.sonarsource.sonarlint.core.embedded.server.ShowIssueRequestHandler;
+
+import static mediumtest.fixtures.ServerFixture.newSonarQubeServer;
+import static mediumtest.fixtures.SonarLintBackendFixture.newBackend;
+import static mediumtest.fixtures.SonarLintBackendFixture.newFakeClient;
+import static mediumtest.fixtures.storage.ServerIssueFixtures.aServerIssue;
+import static org.assertj.core.api.Assertions.assertThat;
+
+class ShowIssueMediumTests {
```
damien-urruty-sonarsource, line 38: I think this feature is a good opportunity to test the full functionality with a medium test:
* trigger an open issue request on the embedded server
* configure the mock web server to return the issue details
* check that the client is called with the right parameters

I can help achieve that if needed

core/src/test/java/org/sonarsource/sonarlint/core/embedded/server/ShowIssueRequestHandlerTest.java
```diff
@@ -0,0 +1,178 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.argThat;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+class ShowIssueRequestHandlerTest {
```
damien-urruty-sonarsource, line 50: I think this could be migrated to a medium test too

core/src/main/java/org/sonarsource/sonarlint/core/embedded/server/ShowIssueRequestHandler.java
```diff
@@ -0,0 +1,211 @@
...
+    }
+  }
+
+  private void showIssueForConnection(String connectionId, String projectKey, String issueKey) {
+    var scopes = configurationService.getConfigScopesWithBindingConfiguredTo(connectionId, projectKey);
+    if (scopes.isEmpty()) {
+      assistBinding(connectionId, projectKey)
+        .thenAccept(newBinding -> showIssueForScope(connectionId, issueKey));
+    } else {
+      showIssueForScope(connectionId, issueKey);
```
damien-urruty-sonarsource, line 117: We probably have the same problem with the show hotspot, but if the `scopes` list has more than one element, should we show the issue in all of them? I can imagine on SLI a situation where a project and a module in a different project are bound to the same connection and projectKey, we probably want to show the issues in the 2 windows

damien-urruty-sonarsource (commented):
Looks good to me overall, but some details need discussion I think

client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/client/issue/ShowIssueParams.java
```diff
@@ -0,0 +1,85 @@
...
+  private final String connectionId;
+  private final String ruleKey;
+  private final String issueKey;
+  private final String creationDate;
+  private final String codeSnippet;
+  private final List<FlowDto> flows;
+  private final TextRangeDto textRange;
+
+  public ShowIssueParams(TextRangeDto textRange, String connectionId, String ruleKey, String issueKey, String filePath, String message,
+    String creationDate, String codeSnippet, List<FlowDto> flows) {
```
damien-urruty-sonarsource, line 38: One thing I forgot is that there is no way for clients to know if the issue is a taint or not. We would need that in case we want to select issues in lists

client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/client/issue/ShowIssueParams.java
vojtech-suchy-sonarsource: Renamed now to serverRelativeFilePath. The actual content is taken from the SQ response, eg _"src/main/java/com/github/kevinsawicki/http/HttpRequest.java"_

client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/client/issue/ShowIssueParams.java
vojtech-suchy-sonarsource: Addressed in [383cbb5](https://github.com/SonarSource/sonarlint-core/pull/722/commits/383cbb5505665f2da97a4c2f86511f34a228339c)

client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/client/issue/ShowIssueParams.java
vojtech-suchy-sonarsource: Now addressed in [c3a404b](https://github.com/SonarSource/sonarlint-core/pull/722/commits/c3a404b9aa999c1e0c62ec5e54c39ff3c06b5d3b)

core/src/main/java/org/sonarsource/sonarlint/core/embedded/server/ShowIssueRequestHandler.java
vojtech-suchy-sonarsource: For now, I've just added a note for future generations as we did for hotspots....

core/src/main/java/org/sonarsource/sonarlint/core/embedded/server/ShowHotspotRequestHandler.java
```diff
@@ -55,21 +53,20 @@
 
 @Named
 @Singleton
-public class ShowHotspotRequestHandler implements HttpRequestHandler {
+public class ShowHotspotRequestHandler extends ShowHotspotOrIssueRequestHandler implements HttpRequestHandler {
```
damien-urruty-sonarsource, line 56: I think it's ok here, but just for sharing, I think it's a common practice to favor [composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance). This could be achieved here by having something like a `BindingAssistantService` that would be used by both handlers

core/src/test/java/mediumtest/issues/ShowIssueMediumTests.java
```diff
@@ -0,0 +1,184 @@
...
+
+    assertThat(backend.telemetryFilePath())
+      .content().asBase64Decoded().asString()
+      .contains("\"showIssueRequestsCount\":0");
+
+    var showIssueRequestHandler = new ShowIssueRequestHandler(newFakeClient().build(), backend.getConnectionConfigurationRepository(),
+      backend.getConfigurationServiceImpl(), backend.getBindingSuggestionProviderImpl(), backend.getServerApiProvider(),
+      backend.getTelemetryServiceImpl());
+
+    showIssueRequestHandler.showIssue(new ShowIssueRequestHandler.ShowIssueQuery(serverWithIssue.baseUrl(), "projectKey", "issueKey"));
```
damien-urruty-sonarsource, line 83: The idea of a medium test is that you exercise the code from the interfaces with the outside world, so in this case I would be more in favor of really making an HTTP request to the server. The reason is that with medium tests we don't want to rely on "internal classes" to not slow down refactoring if needed

core/src/test/java/mediumtest/fixtures/SonarLintTestBackend.java
```diff
@@ -69,6 +74,27 @@ public LocalOnlyIssueStorageService getLocalOnlyIssueStorageService() {
...
+
+  public ServerApiProvider getServerApiProvider() {
+    return getInitializedApplicationContext().getBean(ServerApiProvider.class);
+  }
+
+  public TelemetryServiceImpl getTelemetryServiceImpl() {
+    return getInitializedApplicationContext().getBean(TelemetryServiceImpl.class);
+  }
+
+
```
damien-urruty-sonarsource, line 97: Do you need to expose all of those methods?

core/src/test/java/mediumtest/issues/ShowIssueMediumTests.java
```diff
@@ -0,0 +1,184 @@
...
+    var fakeClient = newFakeClient().build();
+    backend = newBackend()
+      .withSonarQubeConnection(connectionId, serverWithIssue)
+      .withBoundConfigScope(configScopeId, connectionId, projectKey)
+      .withEmbeddedServer()
+      .build(fakeClient);
+
+    HttpRequest request = openIssueWithProjectAndKeyRequest("&issue=" + issueKey, "&project=" + projectKey);
+
+    var response = java.net.http.HttpClient.newHttpClient().send(request, HttpResponse.BodyHandlers.ofString());
```
damien-urruty-sonarsource, line 106: You could maybe extract those 2 lines in a method to improve the readability of the tests

core/src/test/java/mediumtest/issues/ShowIssueMediumTests.java
```diff
@@ -0,0 +1,184 @@
...
+
+  @Test
+  void it_should_fail_request_when_project_parameter_missing() throws Exception{
+    backend = newBackend()
+      .withEmbeddedServer()
+      .build();
+
+    HttpRequest request = openIssueWithProjectAndKeyRequest("", "&project=projectKey");
+    var response = java.net.http.HttpClient.newHttpClient().send(request, HttpResponse.BodyHandlers.ofString());
+    assertThat(response.statusCode()).isEqualTo(400);
```
damien-urruty-sonarsource, line 175: What we also tried to do with style guidelines is to let empty spaces between the Given, When and Then sections, so I would suggest adding an empty line before assertions

core/src/test/java/org/sonarsource/sonarlint/core/embedded/server/ShowIssueRequestHandlerTest.java
damien-urruty-sonarsource: Do you still need these tests now that you have medium ones?

damien-urruty-sonarsource (approved):
Adding more points regarding technical details, feature-wise it seems all good!

PR #725, opened 2023-09-28
SLCORE-553 Handle null previous version and human readable date format

nquinquenel:


commons/src/main/java/org/sonarsource/sonarlint/core/commons/NewCodeDefinition.java
```diff
@@ -20,13 +20,25 @@
...
+  String DATETIME_FORMAT = "MM/dd/yyyy HH:mm:ss";
+  DateTimeFormatter DATETIME_FORMATTER = DateTimeFormatter.ofPattern(DATETIME_FORMAT);
+
   NewCodeMode getMode();
   boolean isOnNewCode(long creationDate);
 
   boolean isSupported();
 
+  static String parseEpochToDate(long epoch) {
+    return ZonedDateTime.ofInstant(Instant.ofEpochMilli(epoch), ZoneId.systemDefault()).format(DATETIME_FORMATTER);
```
damien-urruty-sonarsource, line 39: Potentially this will display a different date than on the server, but not sure it's a problem

damien-urruty-sonarsource:
cc FYI @kirill-knize-sonarsource @jblievremont

server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/newcode/NewCodeApi.java
```diff
@@ -66,6 +66,7 @@ public CompletableFuture<NewCodeDefinition> getNewCodeDefinition(String projectK
       return CompletableFuture.completedFuture(NewCodeDefinition.withNumberOfDays(days, date));
     }
     if (modeString.equalsIgnoreCase("PREVIOUS_VERSION")) {
+      parameter = periodFromWs.hasParameter() ? parameter : null;
```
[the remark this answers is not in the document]
nquinquenel: For now, I only encountered an issue with the `PREVIOUS_VERSION` definition, ideally, we could handle null for each case. Shall we just fall back to `Since DATE` for each scenario?

commons/src/main/java/org/sonarsource/sonarlint/core/commons/NewCodeDefinition.java
nquinquenel: True, in this case, I'd maybe lean towards reusing the date we get from the response and formatting this, although it might still be confusing if the user has a different timezone than the server?

damien-urruty-sonarsource: I think we can keep this like this

damien-urruty-sonarsource (approved):

PR #726, opened 2023-09-29
SLCORE-566 Clear the active branch cache when a config scope binding …

henryju:


henryju:
I did not add a medium test as I think it will be simpler when we rework branch management (there is the FIXME to handle at the same time)

damien-urruty-sonarsource:
Implemented in #768

PR #727, opened 2023-09-29
Fix SonarCloud ITs

henryju:


its/tests/src/test/java/its/SonarCloudTests.java
```diff
@@ -149,6 +156,9 @@ static void prepare() throws Exception {
     randomPositiveInt = new Random().nextInt() & Integer.MAX_VALUE;
 
     adminWsClient = newAdminWsClient();
+    sonarcloudUserToken = adminWsClient.userTokens()
+      .generate(new GenerateRequest().setName(TOKEN_NAME))
+      .getToken();
```
jblievremont, line 161: Could we imagine calling this at the beginning of the test to generate a token, and then using it not only for analysis but also for all server communication? I believe it would be even closer to what users experience in the wild and more future-proof.

jblievremont (approved):
LGTM, and maybe we could make it even better (see my inline comment)

its/tests/src/test/java/its/SonarCloudTests.java
henryju: We could, but I don't think it matters a lot for now as login/password is only deprecated on scanner side AFAIK. When you talk about server communication, I see 2 cases: 
* when configuring the connection in SonarLint: here we are still supporting both login/password and token auth in the UI (at least in SLI), so in theory we should test both
* for ITs setup: here it doesn't matter as long as it works, because this is not real calls made by SonarLint

PR #729, opened 2023-10-02
SLCORE-553 Fetch new code definition only at project level

henryju:


damien-urruty-sonarsource (approved):
LGTM, but I'm really surprised no test had to be modified, as the PR changes the API call and the sync logic

PR #734, opened 2023-10-05
SLCORE-582 CORS preflight request should not be processing the response

nquinquenel:


core/src/main/java/org/sonarsource/sonarlint/core/embedded/server/CorsFilter.java
```diff
@@ -35,25 +36,30 @@ class CorsFilter implements HttpFilterHandler {
...
-      public void sendInformation(ClassicHttpResponse classicHttpResponse) throws HttpException, IOException {
-        responseTrigger.sendInformation(classicHttpResponse);
-      }
 
-      @Override
-      public void submitResponse(ClassicHttpResponse classicHttpResponse) throws HttpException, IOException {
-        if (origin != null) {
-          classicHttpResponse.addHeader("Access-Control-Allow-Origin", origin.getValue());
+    if (Method.OPTIONS.name().equalsIgnoreCase(request.getMethod())) {
+      var response = new BasicClassicHttpResponse(HttpStatus.SC_NO_CONTENT);
```
damien-urruty-sonarsource, line 41: I am not sure about this change of status for the response, I think it was intentional. The only pointer I found is [this SO answer](https://stackoverflow.com/a/58794243) mentioning troubles when returning 204. But maybe it's not a problem anymore. This [link](https://github.com/Kong/kong/pull/4029#issuecomment-648400750) suggests that Mozilla was recommending returning 200 but the page has been updated since and 204 looks fine. Maybe just check it works on major browsers?

damien-urruty-sonarsource (approved):
LTGM, I just left one comment but I think it's OK

PR #735, opened 2023-10-09
SLE-741: Server info about anticipated transitions

thahnen:
## Summary

In order to be used on _SonarLint for Eclipse_, general information about a connection supporting the anticipated transitions should be available. Rather than on an issue level, this information is gathered to configure markers at once and not per marker, saving computing time.

## Testing

Medium tests are added in order to check every possible scenario the new interface method might face.

### DIFF
--- a/client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/backend/issue/CheckAnticipatedStatusChangeSupportedParams.java
+++ b/client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/backend/issue/CheckAnticipatedStatusChangeSupportedParams.java
@@ -1,0 +1,32 @@
+/*
+ * SonarLint Core - Client API
+ * Copyright (C) 2016-2023 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.sonarlint.core.clientapi.backend.issue;
+
+public class CheckAnticipatedStatusChangeSupportedParams {
+  private final String connectionId;
+
+  public CheckAnticipatedStatusChangeSupportedParams(String connectionId) {
+    this.connectionId = connectionId;
+  }
+
+  public String getConnectionId() {
+    return connectionId;
+  }
+}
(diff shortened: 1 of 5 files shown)


client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/backend/issue/CheckAnticipatedStatusChangeSupportedParams.java
```diff
@@ -0,0 +1,32 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.sonarlint.core.clientapi.backend.issue;
+
+public class CheckAnticipatedStatusChangeSupportedParams {
+  private final String connectionId;
+
+  public CheckAnticipatedStatusChangeSupportedParams(String connectionId) {
+    this.connectionId = connectionId;
```
damien-urruty-sonarsource, line 26: Later this feature will also be supported in standalone mode, so maybe passing the `connectionId` is not ideal, and we should pass the `configScopeId`, that the backend is then able to resolve to a connection

client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/backend/issue/IssueService.java
```diff
@@ -70,6 +70,22 @@ public interface IssueService {
...
+   * This method will fail if:
+   * <ul>
+   *   <li>the connectionId provided as a parameter is unknown</li>
+   *   <li>there is a communication problem with the server: network outage, server is down, unauthorized</li>
+   * </ul>
+   * In those cases, a failed future will be returned.
+   * </p>
+   */
+  @JsonRequest
+  CompletableFuture<CheckAnticipatedStatusChangeSupportedResponse> checkAnticipatedStatusChangeSupported(CheckAnticipatedStatusChangeSupportedParams params);
```
damien-urruty-sonarsource, line 87: The name "anticipated status or transition" is the name of the feature and endpoint on the SQ side. We decided to call the feature "resolving local-only issues" because it makes more sense from the SL point of view. But maybe not all our code base uses this name

core/src/test/java/mediumtest/issues/CheckAnticipatedStatusChangeSupportedMediumTests.java
```diff
@@ -0,0 +1,117 @@
...
+    backend = newBackend()
+      .withSonarQubeConnection("connectionId", storage -> storage.withServerVersion("10.2"))
+      .withActiveBranch("configScopeId", "branch")
+      .withBoundConfigScope("configScopeId", "connectionId", "projectKey")
+      .build();
+
+    assertThat(checkAnticipatedStatusChangeSupported("connectionId"))
+      .succeedsWithin(Duration.ofSeconds(2))
+      .extracting(CheckAnticipatedStatusChangeSupportedResponse::isSupported)
+      .isEqualTo(true);
```
damien-urruty-sonarsource, line 111: You can use `isTrue()` or `isFalse()` instead

damien-urruty-sonarsource (approved):
A few comments, but LGTM. Normally you should create a ticket on the core side so that it appears in the release notes. SLE tickets should be reserved for SLE

client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/backend/issue/CheckAnticipatedStatusChangeSupportedParams.java
thahnen: Done, so we don't have to change the API in the future.

PR #739, opened 2023-10-13
SLCORE-589: Make resolved Taint Issues available

thahnen:
To be able to re-open Taint Vulnerabilities from within the IDE we require the option to display the resolved issue again.

### DIFF
--- a/core/src/main/java/org/sonarsource/sonarlint/core/client/api/connected/ConnectedSonarLintEngine.java
+++ b/core/src/main/java/org/sonarsource/sonarlint/core/client/api/connected/ConnectedSonarLintEngine.java
@@ -79,12 +79,13 @@ public interface ConnectedSonarLintEngine extends SonarLintEngine {
   /**
    * Gets locally stored server taint issues for a given file.
    *
-   * @param projectBinding information about the project (must have been previously updated with {@link #updateProject(EndpointParams, HttpClient, String, ClientProgressMonitor)})
-   * @param branchName     branch name
-   * @param filePath       relative to the project.
+   * @param projectBinding  information about the project (must have been previously updated with {@link #updateProject(EndpointParams, HttpClient, String, ClientProgressMonitor)})
+   * @param branchName      branch name
+   * @param filePath        relative to the project.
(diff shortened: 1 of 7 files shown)


thahnen:
@nquinquenel What annoys me is that the SQ analysis is failing because this stuff is tested in ITs but the SonarQube analysis only works with the UTs coverage -.-

core/src/main/java/org/sonarsource/sonarlint/core/client/api/connected/ConnectedSonarLintEngine.java
```diff
@@ -86,6 +86,17 @@ public interface ConnectedSonarLintEngine extends SonarLintEngine {
...
+  /**
+   * Gets locally stored server taint issues for a given file.
+   *
+   * @param projectBinding  information about the project (must have been previously updated with {@link #updateProject(EndpointParams, HttpClient, String, ClientProgressMonitor)})
+   * @param branchName      branch name
+   * @param filePath        relative to the project.
+   * @param includeResolved whether the resolved Taint Vulnerabilities should be included or not
+   * @return All server taint issues in the local storage for the given file. If file has no issues, an empty list is returned.
+   */
+  List<ServerTaintIssue> getServerTaintIssues(ProjectBinding projectBinding, String branchName, String filePath, boolean includeResolved);
```
nquinquenel, line 98: Couldn't we just always include resolved taint instead of overloading the method?

thahnen: Would be an option but in this case everyone implementing it would have to change it. I have no issue with it in this case, but it personally annoys me sometimes that I update SLCORE for my changes and I have to change something completely different (speaking of API changes) ^^

thahnen: Nvm I changed it, I'm too lazy for overloading I noticed

nquinquenel (approved):

PR #747, opened 2023-10-23
Rework the use of completable futures

henryju:


henryju:
@vojtech-suchy-sonarsource sorry, this is a big PR, but we have hopefully already discussed many of the changes.

rpc-impl/src/main/java/org/sonarsource/sonarlint/core/rpc/impl/AbstractRpcServiceDelegate.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Supplier;
+import org.eclipse.lsp4j.jsonrpc.CancelChecker;
+import org.eclipse.lsp4j.jsonrpc.CompletableFutures;
+import org.springframework.beans.factory.BeanFactory;
+
+class AbstractRpcServiceDelegate {
```
vojtech-suchy-sonarsource, line 32: Should we mark this class `abstract`?

rpc-impl/src/main/java/org/sonarsource/sonarlint/core/rpc/impl/AbstractRpcServiceDelegate.java
```diff
@@ -0,0 +1,97 @@
...
+      var wrapper = new CancelCheckerWrapper(cancelChecker);
+      wrapper.checkCanceled();
+      return wrapper;
+    }).thenApplyAsync(cancelChecker -> {
+      cancelChecker.checkCanceled();
+      return code.apply(cancelChecker);
+    }, requestsExecutor);
+  }
+
+  protected <R> CompletableFuture<R> runAsync(Consumer<CancelChecker> code) {
```
vojtech-suchy-sonarsource, line 59: We could lose the generic type and just use `CompletableFuture<Void>` 
(The callers also have `CompletableFuture<Void>` as a return type)

henryju: Good catch, that was the point of this method :)

rpc-impl/src/main/java/org/sonarsource/sonarlint/core/rpc/impl/AbstractRpcServiceDelegate.java
henryju: yes

vojtech-suchy-sonarsource (approved):

PR #750, opened 2023-10-24
SLCORE-594 Replace Guava EventBus by Spring events

henryju:
* Dispatch events synchronously
* Move version soon unsupported check in a background thread

@damien-urruty-sonarsource I think you should review this one as it might conflict with your work


core/src/main/java/org/sonarsource/sonarlint/core/event/ConnectionCredentialsChangedEvent.java
```diff
@@ -19,11 +19,14 @@
...
-public class ConnectionCredentialsChangedEvent {
+import org.springframework.context.ApplicationEvent;
+
+public class ConnectionCredentialsChangedEvent extends ApplicationEvent {
 
   private final String connectionId;
 
-  public ConnectionCredentialsChangedEvent(String connectionId) {
+  public ConnectionCredentialsChangedEvent(Object source, String connectionId) {
+    super(source);
```
damien-urruty-sonarsource, line 29: Do you know what the purpose of this `source` is? It is a bit painful to pass an extra object for each event

damien-urruty-sonarsource (commented):
I am a bit worried that the event handlers are called synchronously for several reasons:
* if an error is thrown in a handler, we need to make sure it does not bubble up to the publisher (maybe already supported by spring events)
* if a handler has a long process or makes a request to the server, we will have to remember to move the computation to a background thread. Each service will have to manage this
* the more handlers, the more it might take time to handle events. The client will get a response to a request only after all event handlers have been processed. 

Also, the events now have a parent that comes from a third-party lib.
Not saying we shouldn't do this move, just that it comes with drawbacks too

core/src/main/java/org/sonarsource/sonarlint/core/event/ConnectionCredentialsChangedEvent.java
henryju: In fact, it is now possible with recent version of Spring to not inherit from the framework abstract class. I will change.

henryju:
> if an error is thrown in a handler, we need to make sure it does not bubble up to the publisher

This is controlled by `eventMulticaster.setErrorHandler(TaskUtils.LOG_AND_SUPPRESS_ERROR_HANDLER)`

> if a handler has a long process or makes a request to the server, we will have to remember to move the computation to a background thread. Each service will have to manage this

yes, but I think it is better than relying on the thread pool of the event bus. At least we can customize the thread name, decide if it should be concurrent, cancel previous tasks if needed, ...

> the more handlers, the more it might take time to handle events. The client will get a response to a request only after all event handlers have been processed.

events are usually created from an RPC notification, so this is not blocking the client. However, it is indeed blocking later events/requests processing.

damien-urruty-sonarsource (approved):

PR #752, opened 2023-10-24
MMF-3220 SonarCloud issues status real-time synchronization

nquinquenel:


core/src/main/java/org/sonarsource/sonarlint/core/websocket/SonarCloudWebSocket.java
```diff
@@ -44,7 +45,9 @@ public static String getUrl() {
     return System.getProperty("sonarlint.internal.sonarcloud.websocket.url", "wss://events-api.sonarcloud.io/");
   }
   private static final Map<String, EventParser<?>> parsersByType = Map.of(
-    "QualityGateChanged", new QualityGateChangedEventParser());
+    "QualityGateChanged", new QualityGateChangedEventParser(),
+    "MyNewIssues", new QualityGateChangedEventParser(),
```
damien-urruty-sonarsource, line 49: Bad copy/paste I think, you need to implement the right parser

core/src/main/java/org/sonarsource/sonarlint/core/websocket/WebSocketService.java
```diff
@@ -212,6 +223,7 @@ private void forgetConnection(String connectionId) {
   private void reopenConnection(String connectionId) {
     closeSocket();
     createConnectionIfNeeded(connectionId);
+    handleEventDispatcher(connectionId);
```
damien-urruty-sonarsource, line 226: I think `reopenConnection` is called only in certain cases, maybe you should call this line from `createConnectionIfNeeded`. To be checked

core/src/main/java/org/sonarsource/sonarlint/core/websocket/WebSocketService.java
```diff
@@ -243,6 +255,14 @@ private void removeProjectsFromSubscriptionListForConnection(String updatedConne
     }
   }
 
+  private void handleEventDispatcher(String connectionId) {
+    this.eventRouterByConnectionId.put(connectionId,
+      new EventDispatcher()
+        .dispatch(QualityGateChangedEvent.class, new ShowSmartNotificationOnQualityGateChangedEvent(client, configurationRepository, telemetryService))
+        .dispatch(IssueChangedEvent.class, new UpdateStorageOnIssueChanged(storageFacade.connection(connectionId)))
```
damien-urruty-sonarsource, line 262: As we have the same handlers for SQ (except `QualityGateChangedEvent`), here you could probably mix this dispatcher with the one configured for SQ. There is probably a way to avoid the double declaration

core/src/main/java/org/sonarsource/sonarlint/core/websocket/WebSocketService.java
```diff
@@ -269,7 +289,9 @@ private void resubscribeAll() {
   private void createConnectionIfNeeded(String connectionId) {
     connectionIdsInterestedInNotifications.add(connectionId);
     if (this.sonarCloudWebSocket == null) {
-      this.sonarCloudWebSocket = SonarCloudWebSocket.create(connectionAwareHttpClientProvider.getHttpClient(connectionId), eventRouter::handle, this::reopenConnectionOnClose);
+      this.sonarCloudWebSocket = SonarCloudWebSocket.create(connectionAwareHttpClientProvider.getHttpClient(connectionId),
+        event -> this.eventRouterByConnectionId.get(connectionId).handle(event),
```
damien-urruty-sonarsource, line 293: Here I think it's better to null-check the value returned from the map, in case we receive an event after the connection is closed. It could happen if we just closed the connection but an event arrived at the same time, but I reckon it's quite unlikely

damien-urruty-sonarsource (commented):
There are also things to adapt in the `SmartNotifications` service. We need to poll only if the web socket is not opened

core/src/main/java/org/sonarsource/sonarlint/core/websocket/SonarCloudWebSocket.java
nquinquenel: I don't think, the payload is exactly the same

damien-urruty-sonarsource: Ah ok, yes good point. Then I suggest to rename it, having a "QualityGate" parser for MyNewIssues could be confusing

core/src/main/java/org/sonarsource/sonarlint/core/websocket/WebSocketService.java
damien-urruty-sonarsource: Thinking about it a bit more, maybe it's better to keep 2 separate dispatchers for now. Even if there is some redundancy/duplication, we don't support exactly the same events.

core/src/main/java/org/sonarsource/sonarlint/core/websocket/parsing/ServerMessageEventParser.java
```diff
@@ -23,29 +23,35 @@
 import java.util.Optional;
 import org.sonarsource.sonarlint.core.commons.log.SonarLintLogger;
 import org.sonarsource.sonarlint.core.serverapi.push.parsing.EventParser;
-import org.sonarsource.sonarlint.core.websocket.events.QualityGateChangedEvent;
+import org.sonarsource.sonarlint.core.websocket.events.ServerMessageEvent;
 
 import static org.sonarsource.sonarlint.core.serverapi.util.ServerApiUtils.isBlank;
 
-public class QualityGateChangedEventParser implements EventParser<QualityGateChangedEvent> {
+public class ServerMessageEventParser implements EventParser<ServerMessageEvent> {
```
damien-urruty-sonarsource, line 30: Maybe we could call it `SmartNotificationEvent`?

core/src/main/java/org/sonarsource/sonarlint/core/websocket/SonarCloudWebSocket.java
```diff
@@ -108,12 +115,16 @@ private static Optional<? extends ServerEvent> parse(WebSocketEvent event) {
...
       SonarLintLogger.get().error("Unknown '{}' event type ", eventType);
       return Optional.empty();
     }
     try {
-      return parsersByType.get(eventType).parse(event.data.toString());
+      var eventParser = parsersByTypeForProjectFilter.get(eventType);
+      if (eventParser == null) {
+        eventParser = parsersByTypeForProjectUserFilter.get(eventType);
+      }
+      return eventParser.parse(event.data.toString());
```
damien-urruty-sonarsource, line 127: We could have the 2 separate maps as you did, and then one field that aggregates the 2 so that it's simpler to use here

core/src/main/java/org/sonarsource/sonarlint/core/smartnotifications/SmartNotifications.java
```diff
@@ -115,21 +115,19 @@ private void manageNotificationsForConnection(ServerApi serverApi, Map<String, S
...
       }
 
       projectKeysByLastEventPolling.keySet()
         .forEach(projectKey -> lastEventPollingService.setLastEventPolling(ZonedDateTime.now(), connectionId, projectKey));
     }
   }
 
-  private boolean shouldIgnoreNotification(AbstractConnectionConfiguration connection, ServerNotification n) {
-    return connection.getKind() == ConnectionKind.SONARCLOUD && webSocketService.hasOpenConnection() && "QUALITY_GATE".equals(n.category());
+  private boolean shouldIgnoreNotification(AbstractConnectionConfiguration connection) {
```
damien-urruty-sonarsource, line 129: Maybe rename it `shouldSkipPolling`?

core/src/main/java/org/sonarsource/sonarlint/core/websocket/SonarCloudWebSocket.java
```diff
@@ -111,16 +131,27 @@ private static Optional<? extends ServerEvent> parse(WebSocketEvent event) {
...
-    } else if (!parsersByType.containsKey(eventType)) {
-      SonarLintLogger.get().error("Unknown '{}' event type ", eventType);
-      return Optional.empty();
     }
+
+    return Stream.of(parsersByTypeForProjectFilter, parsersByTypeForProjectUserFilter)
+      .flatMap(map -> map.entrySet().stream())
+      .filter(entry -> eventType.equals(entry.getKey()))
+      .map(Map.Entry::getValue)
+      .findFirst()
```
damien-urruty-sonarsource, line 140: Do we need to do this every time we receive an event? I think we could have a Map field that combines the 2 other Maps

damien-urruty-sonarsource (approved):
My only concern is that we don't test new events through medium tests, so we don't check if everything is wired properly (parsing and handling), which is normally the purpose of medium tests

nquinquenel:
> My only concern is that we don't test new events through medium tests, so we don't check if everything is wired properly (parsing and handling), which is normally the purpose of medium tests

Everything should be tested now

PR #755, opened 2023-10-25
SLCORE-603 Fix log output when using JDK httpclient

henryju:
The HttpClient uses a [ThreadPool](https://github.com/openjdk/jdk/blob/d2d1592dd94e897fae6fc4098e43b4fffb6d6750/src/java.net.http/share/classes/jdk/internal/net/http/HttpClientImpl.java#L140C50-L140C55) that doesn't copy thread-local variables, so it means our ClientLogOutput might not be the correct one.


http/src/main/java/org/sonarsource/sonarlint/core/http/ApacheHttpClientAdapter.java
```diff
@@ -265,11 +268,14 @@ public void cancel() {
...
     // TODO handle handshake or other errors
     return java.net.http.HttpClient
-        .newHttpClient()
-        .newWebSocketBuilder()
-        .header("Authorization", "Bearer " + usernameOrToken)
-        .buildAsync(URI.create(url), new WebSocketClient(messageConsumer, onClosedRunnable))
-        .join();
+      .newBuilder()
+      // Don't use the default thread pool as it won't allow inheriting thread local variables
+      .executor(Executors.newCachedThreadPool(threadWithNamePrefix("sonarcloud-websocket-")))
```
damien-urruty-sonarsource, line 273: Usually, we try to shutdown created executors, isn't it needed? In fact the same question goes for the client, is there some kind of cleanup to do at the end? I might have missed that when we implemented web sockets

damien-urruty-sonarsource (approved):
Only one question, the rest LGTM

http/src/main/java/org/sonarsource/sonarlint/core/http/ApacheHttpClientAdapter.java
henryju: I think closing the Websocket should be enough to release all resources.

henryju: In the end I made more refactoring and I decided to close the executor service

http/src/main/java/org/sonarsource/sonarlint/core/http/WebSocketClient.java
```diff
@@ -0,0 +1,122 @@
...
+      .newWebSocketBuilder()
+      .header("Authorization", "Bearer " + token)
+      .buildAsync(URI.create(url), new MessageConsummerWrapper(messageConsumer, onClosedRunnable))
+      .join();
+  }
+
+  @Override
+  public void close() {
+    if (!MoreExecutors.shutdownAndAwaitTermination(executor, 1, TimeUnit.SECONDS)) {
+      LOG.warn("Unable to stop hotspot updater executor service in a timely manner");
```
damien-urruty-sonarsource, line 70: Copy/pasta?

damien-urruty-sonarsource (approved):
LGTM

http/pom.xml
```diff
@@ -51,6 +51,10 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+    </dependency>
```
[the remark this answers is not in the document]
henryju: This is for `MoreExecutors`

http/src/main/java/org/sonarsource/sonarlint/core/http/WebSocketClient.java
henryju: oups

PR #756, opened 2023-10-25
SLCORE-579 Redirect logs to the client through the RPC protocol

henryju:


vojtech-suchy-sonarsource (approved):
LGTM.
Please see the unused import warning by SQ

core/src/main/java/org/sonarsource/sonarlint/core/ConnectedSonarLintEngineImpl.java
```diff
@@ -110,7 +110,7 @@ public AnalysisEngine getAnalysisEngine() {
     return analysisContext.get().analysisEngine;
   }
 
-  public AnalysisContext start() {
+  private AnalysisContext start() {
     setLogging(null);
```
vojtech-suchy-sonarsource, line 114: Can this be removed now?

vojtech-suchy-sonarsource (approved):

core/src/main/java/org/sonarsource/sonarlint/core/ConnectedSonarLintEngineImpl.java
henryju: I am not sure as the start() method is also called from another place (restart()). Anyway setting the log output multiple time is harmless.

PR #757, opened 2023-10-26
Feature/vs/slcore 574 bundle platform specific sloop

vojtech-suchy-sonarsource:


embedded-core/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+    <parent>
+      <groupId>org.sonarsource.sonarlint.core</groupId>
+      <artifactId>sonarlint-core-parent</artifactId>
+      <version>9.4-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>sonarlint-embedded-core</artifactId>
+
+    <properties>
+        <maven.compiler.source>17</maven.compiler.source>
```
henryju, line 15: This is suspicious. We should rely on the settings defined in the parent, and not define them in each module.

process-cli/pom.xml
```diff
@@ -18,12 +288,13 @@
...
       <groupId>info.picocli</groupId>
       <artifactId>picocli</artifactId>
-      <version>4.7.1</version>
+      <version>4.7.3</version>
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>sonarlint-rpc-impl</artifactId>
       <version>${project.version}</version>
+      <scope>compile</scope>
```
henryju, line 297:
```suggestion
```

process-cli/src/main/assembly/bin/slcore-process-cli
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env sh
```
henryju, line 1: I would choose the name of the script with care, as it might appear in the OS process explorer. What about `sonarlint-backend`?

process-cli/pom.xml
```diff
@@ -9,6 +9,276 @@
...
+
+  <profiles>
+      <profile>
+        <id>dist-windows</id>
+          <build>
+            <plugins>
+              <plugin>
+                <groupId>com.googlecode.maven-download-plugin</groupId>
+                <artifactId>download-maven-plugin</artifactId>
+                <version>1.6.8</version>  <!--Issue in 1.7.1 https://github.com/maven-download-plugin/maven-download-plugin/issues/260 -->
```
henryju, line 44: Maybe use a property to avoid repeating the version

process-cli/src/main/assembly/bin/slcore-process-cli
```diff
@@ -0,0 +1,28 @@
...
+APP_HOME=`dirname "$SCRIPT_PATH"`/..
+
+# make it fully qualified
+APP_HOME=`cd "$APP_HOME" && pwd -P`
+
+CLASSPATH=$APP_HOME/lib/*:
+
+JAVACMD=$APP_HOME/jre/bin/java
+
+exec "$JAVACMD" -classpath "$CLASSPATH" org.sonarsource.sonarlint.core.rpc.impl.SonarLintCoreProcess "$@"
```
henryju, line 28: Passing `org.sonarsource.sonarlint.core.rpc.impl.SonarLintCoreProcess` should not be needed as it should be the Main-Class in the MANIFEST

process-cli/src/main/assembly/filter-dist.properties
```diff
@@ -0,0 +1 @@
+use_embedded_jre=true
```
henryju, line 1: What is the use of this?

henryju (changes requested):

process-cli/src/main/assembly/filter-dist.properties
vojtech-suchy-sonarsource: Good catch, it's a leftover we don't need. (Was supposed to serve as a property we could query from launcher scripts). Removed in [79ddfa3](https://github.com/SonarSource/sonarlint-core/pull/757/commits/79ddfa394bd9a8f78543ac932e2eda5e69cdc53b)

process-cli/pom.xml
vojtech-suchy-sonarsource: Addressed in [79ddfa3](https://github.com/SonarSource/sonarlint-core/pull/757/commits/79ddfa394bd9a8f78543ac932e2eda5e69cdc53b)

process-cli/pom.xml
vojtech-suchy-sonarsource: Good catch, addressed in [79ddfa3](https://github.com/SonarSource/sonarlint-core/pull/757/commits/79ddfa394bd9a8f78543ac932e2eda5e69cdc53b)

embedded-core/pom.xml
vojtech-suchy-sonarsource: Agree. We don't seem to be setting these properties in other `pom.xml`s either so I removed them all.

henryju (approved):

PR #765, opened 2023-11-01
Create a Java client library for out of process

vojtech-suchy-sonarsource:


its/tests/pom.xml
```diff
@@ -110,6 +116,35 @@
...
           </execution>
+          <execution>
+            <id>fetch-sloop-distribution</id>
+            <phase>generate-test-resources</phase>
+            <goals>
+              <goal>copy</goal>
+            </goals>
+            <configuration>
+              <artifactItems>
+                <artifactItem>
```
henryju, line 127: Minor optimization, but we could use Maven profiles to only download the distribution of the current platform

its/tests/src/test/java/its/SloopLauncherWithJreTest.java
```diff
@@ -0,0 +1,92 @@
...
+  private static Path sonarUserHome;
+
+  private static SonarLintRpcServer server;
+
+  @BeforeEach
+  void setup() {
+    var sloopDistPath = SystemUtils.IS_OS_WINDOWS ? SloopDistLocator.getWindowsDistPath() : SloopDistLocator.getLinux64DistPath();
+    var unArchiver = SystemUtils.IS_OS_WINDOWS ? new ZipUnArchiver() : new TarGZipUnArchiver();
+    var outDirPath = unarchiveDistribution(sloopDistPath.toString(), unArchiver);
+    server = SloopLauncher.startSonarLintRpcServerWithJre(sloopDistPath.toString(), newDummySonarLintClient(), outDirPath.resolve("jre").toAbsolutePath().toString());
```
henryju, line 59: If I understand, you are still using the jre provided in the distribution, so this is difficult to confirm the test is actually doing something. Maybe we could rely on the JAVA_HOME being set (it is probably true on our machine and on Cirrus, and run the backend with the JRE located in JAVA_HOME?
Still remain the question of how to assert. Idea: we could log during the backend startup the version/path of the current JRE, and use that to assert?

process-cli/src/main/java/org/sonarsource/sonarlint/core/rpc/impl/SonarLintCoreProcess.java
```diff
@@ -19,36 +19,63 @@
...
 @CommandLine.Command(name = "slcore", mixinStandardHelpOptions = true, description = "The SonarLint Core backend")
 public class SonarLintCoreProcess implements Callable<Integer> {
 
+  //FIXME remove the logger before merging
+  private static Logger logger;
+
   @Override
-  public Integer call() {
+  public Integer call() throws FileNotFoundException {
+    logger.info("Starting SonarLint Core backend call()...");
```
henryju, line 43: Very dangerous, as I guess Logger will go to stdout by default

process-cli/src/main/java/org/sonarsource/sonarlint/core/rpc/impl/SonarLintCoreProcess.java
```diff
@@ -19,36 +19,63 @@
...
     var originalStdIn = System.in;
     var originalStdOut = System.out;
     System.setIn(new ByteArrayInputStream(new byte[0]));
     // Redirect all logs to stderr for now, would be better to go to a file later
-    System.setOut(System.err);
 
+    System.setErr(new PrintStream(new BufferedOutputStream(new FileOutputStream("SystemErrOoutput.txt"))));
```
henryju, line 49: Was it only for testing? I don't think we agreed on logging to a file for now. And if we do, we should let the client decide of the location (it could be a command line argument)

rpc-java-client/src/main/java/org/sonarsource/sonarlint/core/rpc/client/SloopLauncher.java
```diff
@@ -0,0 +1,132 @@
...
+    if (process.waitFor(1, TimeUnit.MINUTES)){
+      return process.exitValue();
+    } else {
+      process.destroyForcibly();
+      return -1;
+    }
+  }
+
+  private static String unarchiveDistribution(String distArchivePath) {
+    var outputDirPath = Paths.get("").resolve("out").toAbsolutePath().toString();
```
henryju, line 118: Not great to not control where the backend is extracted. In fact I don't think it should be the responsibility of this class to deal with the extraction. I guess IDEs will try to only extract once, to avoid the cost. Similar to what we do with Omnisharp, this class should probably take as argument the path where the backend is extracted.

henryju (changes requested):

its/tests/src/test/java/its/SloopLauncherWithJreTest.java
vojtech-suchy-sonarsource: It's correct that the jre provided from the distribution is used but it's being extracted to a different place than where the SloopLanucher will extract it so I would think it's okay to use it. (It's similar to a case where the user has the same version of jre as the one we provide in the distribution.).
I will update the assertion as suggested.

rpc-java-client/src/main/java/org/sonarsource/sonarlint/core/rpc/client/SloopLauncher.java
vojtech-suchy-sonarsource: Addressed in [3a40e35](https://github.com/SonarSource/sonarlint-core/pull/765/commits/3a40e35e93ba7a3f6bf6823de2bbb829b72b9fc4)

its/tests/src/test/java/its/SloopLauncherWithJreTest.java
vojtech-suchy-sonarsource: Addressed in [55e2773](https://github.com/SonarSource/sonarlint-core/pull/765/commits/55e2773e899579f973ce42cbe69548d564be5dff)

process-cli/src/main/java/org/sonarsource/sonarlint/core/rpc/impl/SonarLintCoreProcess.java
vojtech-suchy-sonarsource: Addressed in [55c408e](https://github.com/SonarSource/sonarlint-core/pull/765/commits/55c408e7ff6269bcbb74e6eb1605a618f60f3987)

its/tests/pom.xml
vojtech-suchy-sonarsource: addressed in [0b76eac](https://github.com/SonarSource/sonarlint-core/pull/765/commits/0b76eacc86d4104f1638cad6914f45b521f5346a)

pom.xml
```diff
@@ -62,6 +62,7 @@
     <mockito.version>5.6.0</mockito.version>
     <kotlin.version>1.6.21</kotlin.version>
     <lsp4j.version>0.21.1</lsp4j.version>
+    <plugin.download.version>1.6.8</plugin.download.version> <!--Issue in 1.7.1 https://github.com/maven-download-plugin/maven-download-plugin/issues/260 -->
```
henryju, line 65: Why not using a `pluginManagement`?

its/tests/src/test/java/its/SloopLauncherWithJreTest.java
```diff
@@ -0,0 +1,94 @@
...
+import org.sonarsource.sonarlint.core.rpc.protocol.SonarLintRpcServer;
+import org.sonarsource.sonarlint.core.rpc.protocol.backend.initialize.ClientInfoDto;
+import org.sonarsource.sonarlint.core.rpc.protocol.backend.initialize.FeatureFlagsDto;
+import org.sonarsource.sonarlint.core.rpc.protocol.backend.initialize.InitializeParams;
+
+import static its.utils.UnArchiveUtils.unarchiveDistribution;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.sonarsource.sonarlint.core.rpc.protocol.common.Language.GO;
+
+class SloopLauncherWithJreTest {
```
henryju, line 46:
```suggestion
class SloopLauncherWithJreTests {
```

its/tests/src/test/java/its/SloopLauncherTest.java
```diff
@@ -0,0 +1,209 @@
...
+import org.sonarsource.sonarlint.core.rpc.protocol.client.progress.ReportProgressParams;
+import org.sonarsource.sonarlint.core.rpc.protocol.client.progress.StartProgressParams;
+import org.sonarsource.sonarlint.core.rpc.protocol.client.smartnotification.ShowSmartNotificationParams;
+import org.sonarsource.sonarlint.core.rpc.protocol.client.sync.DidSynchronizeConfigurationScopeParams;
+
+import static its.utils.UnArchiveUtils.unarchiveDistribution;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.sonarsource.sonarlint.core.rpc.protocol.common.Language.GO;
+
+class SloopLauncherTest {
```
henryju, line 73:
```suggestion
class SloopLauncherTests {
```

rpc-java-client/pom.xml
```diff
@@ -25,6 +25,15 @@
       <artifactId>sonarlint-rpc-protocol</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-compress</artifactId>
```
henryju, line 30: Is it still needed now?

henryju (approved):
Minor feedback then LGTM

PR #773, opened 2023-11-07
Fix some issues found while testing in SLE

henryju:


rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/adapter/EitherCredentialsAdapterFactory.java
```diff
@@ -38,8 +40,8 @@ public class EitherCredentialsAdapterFactory implements TypeAdapterFactory {
   @SuppressWarnings("unchecked")
   @Override
   public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> type) {
-    Predicate<JsonElement> leftChecker = new EitherTypeAdapter.PropertyChecker("token");
-    Predicate<JsonElement> rightChecker = new EitherTypeAdapter.PropertyChecker("username");
-    return (TypeAdapter<T>) new EitherTypeAdapter<>(gson, ELEMENT_TYPE, leftChecker, rightChecker);
+    Predicate<JsonElement> tokenChecker = new EitherTypeAdapter.PropertyChecker("token");
```
henryju, line 43: If the client return a null token or null username (that was the case on my SLE test instance), then Gson was not able to determine the Either "side", and so it was causing a parsing error. I think it is safer like that.

vojtech-suchy-sonarsource (approved):
🚢 it

PR #774, opened 2023-11-07
SLCORE-496 Manage synchronization in the new backend

damien-urruty-sonarsource:


medium-tests/src/test/java/mediumtest/PathMatchingMediumTests.java
```diff
@@ -0,0 +1,147 @@
...
+
+  @Test
+  void it_should_match_without_prefixes_when_local_and_main_branch_server_paths_are_the_same() {
+    var server = newSonarQubeServer("10.3")
+      .withProject("projectKey", project ->
+        project.withBranch("main")
+        .withFile("relative/path/to/a/file"))
+      .start();
+    var client = newFakeClient()
+      .withFile("configScopeId", "relative/path/to/a/file")
```
henryju, line 90: As discussed with Vojtech, I think we could stop coding manually the "stubbing" of the FakeClient. You can directly use Mockito. This allows much more flexibility, and I don't think it is hurting readability. Something like:
```java
var client = spy(newFakeClient().build());
when(client.listAllFilePaths(any()).thenReturn(xxx);
```

medium-tests/src/test/java/mediumtest/fixtures/SonarLintBackendFixture.java
```diff
@@ -486,6 +497,11 @@ public SonarLintClientBuilder withSonarProjectBranchMatchingError() {
       this.shouldErrorProjectBranchResolution = true;
       return this;
     }
+
+    public SonarLintClientBuilder withFile(String configScopeId, String fileRelativePath) {
```
henryju, line 501: As mentioned earlier, I am in favor of stopping manually coding this mock, and instead using Mockito.

medium-tests/src/test/java/mediumtest/synchronization/PluginSynchronizationMediumTests.java
```diff
@@ -0,0 +1,330 @@
...
+    backend = newBackend()
+      .withExtraEnabledLanguagesInConnectedMode(Language.JAVA)
+      .withSonarQubeConnection("connectionId", server)
+      .withBoundConfigScope("configScopeId", "connectionId", "projectKey", "main")
+      .withFullSynchronization()
+      .build(client);
+
+    waitAtMost(3, SECONDS).untilAsserted(() -> {
+      assertThat(getPluginsStorageFolder()).isNotEmptyDirectory();
+      assertThat(client.getConnectionsWhichUpdatedPlugins()).containsExactly("connectionId");
```
henryju, line 91: And same for assertions, we can use Mockito (assuming the fakeClient has been spied):
```java
verify(client).didUpdatePlugins(argCaptor.capture());
```

medium-tests/src/test/java/mediumtest/fixtures/SonarLintTestRpcServer.java
```diff
@@ -158,6 +165,10 @@ public StorageService getIssueStorageService() {
     return serverUsingJava.getIssueStorageService();
   }
 
+  public MatchedSonarProjectBranchRepository getMatchedSonarProjectBranchRepository() {
```
henryju, line 168: Shouldn't we avoid exposing core internals to medium tests that on the long run?

spec/connected_mode/synchronization/pull_synchronization.adoc
henryju, line 1: :heart:

henryju (approved):
Overall looks good to me, and I don't see anything preventing to merge in the feature branch. 2 points we can discuss:
* accessing internal beans from medium tests
* fake client used in medium tests (Mockito vs manual stubbing)

medium-tests/src/test/java/mediumtest/PathMatchingMediumTests.java
damien-urruty-sonarsource: I was wondering about the same thing. Sometimes using Mockito could be simpler, but in the fake client there are some reasonable defaults, e.g. for branch resolution we return the main branch by default, which is used in several scenarios. There might be other examples. We could give it a try, but in a separate PR

medium-tests/src/test/java/mediumtest/fixtures/SonarLintBackendFixture.java
damien-urruty-sonarsource: Same answer

medium-tests/src/test/java/mediumtest/synchronization/PluginSynchronizationMediumTests.java
damien-urruty-sonarsource: Same answer

damien-urruty-sonarsource: Another thing to consider is that verifying this kind of calls is a bit more verbose as we often have to use a captor

medium-tests/src/test/java/mediumtest/fixtures/SonarLintTestRpcServer.java
damien-urruty-sonarsource: I agree it's a shortcut. I call this method to set an initial matched branch when starting the backend, to avoid having to configure the client. It's not used from the medium tests, only from the fixtures. I will think about it, but I prefer to merge this one to avoid further conflicts

PR #775, opened 2023-11-08
Medium test performance fixes

henryju:
This is fixing a thread leak in the WebSocket part.

Medium tests before:
image
After:
image

### DIFF
--- a/http/src/main/java/org/sonarsource/sonarlint/core/http/WebSocketClient.java
+++ b/http/src/main/java/org/sonarsource/sonarlint/core/http/WebSocketClient.java
@@ -19,22 +19,16 @@
  */
 package org.sonarsource.sonarlint.core.http;
 
-import com.google.common.util.concurrent.MoreExecutors;
 import java.net.URI;
 import java.net.http.HttpClient;
 import java.net.http.WebSocket;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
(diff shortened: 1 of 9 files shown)


http/src/main/java/org/sonarsource/sonarlint/core/http/WebSocketClient.java
```diff
@@ -64,22 +56,23 @@ public CompletableFuture<WebSocket> createWebSocketConnection(String url, Consum
       .buildAsync(URI.create(url), new MessageConsummerWrapper(messageConsumer, onClosedRunnable, currentThreadOutput));
   }
 
-  private class MessageConsummerWrapper implements WebSocket.Listener {
+  private static class MessageConsummerWrapper implements WebSocket.Listener {
```
damien-urruty-sonarsource, line 59: Maybe we could fix the typo at the same time: Consummer -> Consumer

rpc-impl/src/main/java/org/sonarsource/sonarlint/core/rpc/impl/SonarLintRpcServerImpl.java
```diff
@@ -179,7 +179,8 @@ public NewCodeRpcService getNewCodeService() {
 
   @Override
   public CompletableFuture<Void> shutdown() {
-    return CompletableFutures.computeAsync(cancelChecker -> {
+    var executor = Executors.newSingleThreadExecutor(r -> new Thread(r, "SonarLint Server shutdown"));
```
damien-urruty-sonarsource, line 182: Shouldn't this executor be shut down too?

damien-urruty-sonarsource (approved):
Small comments, else I don't see anything wrong

PR #776, opened 2023-11-10
RPC feedback

henryju:
Feedback from @eray-felek-sonarsource

### DIFF
--- a/rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/SonarLintRpcClient.java
+++ b/rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/SonarLintRpcClient.java
@@ -145,6 +145,9 @@ public interface SonarLintRpcClient {
   @JsonNotification
   void didSynchronizeConfigurationScopes(DidSynchronizeConfigurationScopeParams params);
 
+  /**
+   * @throws org.eclipse.lsp4j.jsonrpc.ResponseErrorException with {@link org.sonarsource.sonarlint.core.rpc.protocol.client.ClientErrorCode#CONNECTION_NOT_FOUND} if the connection doesn't exist on the client side
+   */
   @JsonRequest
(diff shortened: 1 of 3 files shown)


rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/SonarLintRpcClient.java
```diff
@@ -145,6 +145,9 @@ public interface SonarLintRpcClient {
   @JsonNotification
   void didSynchronizeConfigurationScopes(DidSynchronizeConfigurationScopeParams params);
 
+  /**
+   * @throws org.eclipse.lsp4j.jsonrpc.ResponseErrorException with {@link org.sonarsource.sonarlint.core.rpc.protocol.client.ClientErrorCode#CONNECTION_NOT_FOUND} if the connection doesn't exist on the client side
+   */
   @JsonRequest
   CompletableFuture<GetCredentialsResponse> getCredentials(GetCredentialsParams params);
```
vojtech-suchy-sonarsource, line 152: Should we explicitly add the ResponseErrorException to the method signature so we don't forget to handle it?
`CompletableFuture<GetCredentialsResponse> getCredentials(GetCredentialsParams params) throws ResponseErrorException;`

vojtech-suchy-sonarsource (approved):
LGTM. Please see the comment for SonarLintRpcClient

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/SonarLintRpcClient.java
henryju: ResponseErrorException is a RuntimeException, so I fear we will get warnings from the SQ analysis and from the IDE if we do.

henryju: In fact, any call can possibly throw a ResponseErrorException

PR #777, opened 2023-11-10
Sloop hardening

henryju:


medium-tests/src/test/java/mediumtest/fixtures/SonarLintBackendFixture.java
```diff
@@ -471,15 +467,10 @@ public SonarLintClientBuilder assistingConnectingAndBindingToSonarQube(String sc
...
+        rejectingProgress, proxy, proxyAuth, credentialsByConnectionId, printLogsToStdOut, resolvedSonarProjectBranch, isResolvedSonarProjectBranchSet,
         shouldErrorProjectBranchResolution);
     }
 
-    public SonarLintClientBuilder withDelegate(SonarLintRpcClient delegate) {
-      this.delegate = delegate;
-      return this;
-    }
-
     public SonarLintClientBuilder printLogsToStdOut() {
```
vojtech-suchy-sonarsource, line 474: Not introduced in this PR, but we could remove this printLogsToStdOut() method and the associated code

vojtech-suchy-sonarsource (approved):

medium-tests/src/test/java/mediumtest/fixtures/SonarLintBackendFixture.java
henryju: this is useful when debugging a test. We should just avoid using it on CI to limit log pollution (but it can still be useful if a test fails only on CI)

PR #779, opened 2023-11-14
SLCORE-621 Remove Server-Sent Events subscription from engine side

damien-urruty-sonarsource:
It is no longer used by any client.
This will facilitate other work, e.g. around taint vulnerabilities, as event handling is entirely under the backend responsibility


henryju (approved):
Go go go (but I would create a ticket)

PR #780, opened 2023-11-15
Remove unnecessary RPC wrapping in service layer

henryju:
I am not done yet, but this part can already be reviewed.


core/src/main/java/org/sonarsource/sonarlint/core/ConnectionService.java
```diff
@@ -168,33 +156,32 @@ private void updateConnection(AbstractConnectionConfiguration connectionConfigur
...
-      serverOrganizations.stream().map(o -> new OrganizationDto(o.getKey(), o.getName(), o.getDescription())).collect(Collectors.toList()));
+    return serverOrganizations.stream().map(o -> new OrganizationDto(o.getKey(), o.getName(), o.getDescription())).collect(Collectors.toList());
   }
 
-  public GetOrganizationResponse getOrganization(GetOrganizationParams params, CancelChecker cancelToken) {
-    var helper = buildSonarCloudNoOrgApiHelper(params.getCredentials());
-    var serverOrganization = new OrganizationApi(helper).getOrganization(params.getOrganizationKey(), new ProgressMonitor(null));
-    return new GetOrganizationResponse(serverOrganization.map(o -> new OrganizationDto(o.getKey(), o.getName(), o.getDescription())).orElse(null));
+  @CheckForNull
+  public OrganizationDto getOrganization(Either<TokenDto, UsernamePasswordDto> credentials, String organizationKey) {
```
damien-urruty-sonarsource, line 179: I wonder if we shouldn't do the same for DTOs too, as they are normally specific to RPC

core/src/main/java/org/sonarsource/sonarlint/core/issue/IssueService.java
```diff
@@ -97,17 +89,15 @@ public IssueService(ConfigurationRepository configurationRepository, ServerApiPr
     this.telemetryService = telemetryService;
   }
 
-  public void changeStatus(ChangeIssueStatusParams params, CancelChecker cancelChecker) {
-    var configurationScopeId = params.getConfigurationScopeId();
+  public void changeStatus(String configurationScopeId, String issueKey, ResolutionStatus newStatus, boolean isTaintIssue, CancelChecker cancelChecker) {
```
damien-urruty-sonarsource, line 92: At some point should we introduce an intermediate type if there are a lot of parameters?

damien-urruty-sonarsource (approved):
LGTM

core/src/main/java/org/sonarsource/sonarlint/core/issue/IssueService.java
henryju: This is hiding the dust under the carpet :) A method with many parameters is maybe a sign of bad design. In this example, I expect at some point to have an issue UUID and to not have to tell if this is a taint or not.

core/src/main/java/org/sonarsource/sonarlint/core/ConnectionService.java
henryju: That would be the next step if we want complete isolation between layers, but then we will have more boilerplate code. I suggest we keep using Dtos when needed, and we can adjust later.

damien-urruty-sonarsource (approved):

PR #782, opened 2023-11-16
SLCORE-625 Dispatch server events to services

damien-urruty-sonarsource:


henryju (approved):
Very nice!

henryju (approved):
I like `SonarServer` use, thanks!

PR #783, opened 2023-11-16
SLCORE-626 Remove unnecessary RPC wrapping on client side

henryju:
* don't expose xxxParams and xxxResponse to the client if possible
* provide default implementations on the client
* replace manual stubbing by the use of Mockito in medium tests


medium-tests/src/test/java/mediumtest/BindingSuggestionsMediumTests.java
```diff
@@ -80,15 +88,17 @@ void test_connection_added_should_suggest_binding_with_no_matches() {
     backend.getConnectionService()
       .didUpdateConnections(new DidUpdateConnectionsParams(List.of(new SonarQubeConnectionConfigurationDto(MYSONAR, sonarqubeMock.baseUrl(), true)), List.of()));
 
-    await().atMost(Duration.of(5, ChronoUnit.SECONDS)).until(fakeClient::hasReceivedSuggestions);
-    var bindingSuggestions = fakeClient.getBindingSuggestions();
+    ArgumentCaptor<Map<String, List<BindingSuggestionDto>>> suggestionCaptor = ArgumentCaptor.forClass(Map.class);
+    verify(fakeClient, timeout(5000)).suggestBinding(suggestionCaptor.capture());
+
+    var bindingSuggestions = suggestionCaptor.getValue();
```
damien-urruty-sonarsource, line 94: I am not a fan of mocks in general so my comment might be subjective...
When reading side-by-side, I find the previous version more concise and more readable. If you remove punctuation, it reads almost like natural text. The new version is IMO less easy to read, with lots of `<` and `>` and the captor declaration on the first line is very technical. 
Some other changes read a bit better

henryju: > I am not a fan of mocks in general 
Here I think I am replacing a manually developed mock by Mockito. The principle of mocking (vs using production objects for example) is the same.

For simple cases, mockito + argument captor is maybe a bit more verbose, but as soon as you want to do more complex tests (throwing exceptions, different return based on arguments, delay, ...) it makes things much simpler.

henryju: Also, for simple cases, ArgumentCaptor can be replaced by argument matchers:
Instead of:
```java
ArgumentCaptor<Map<String, List<BindingSuggestionDto>>> suggestionCaptor = ArgumentCaptor.forClass(Map.class);
verify(fakeClient, timeout(5000)).suggestBinding(suggestionCaptor.capture());
assertThat(bindingSuggestions).containsOnlyKeys(CONFIG_SCOPE_ID);
assertThat(bindingSuggestions.get(CONFIG_SCOPE_ID)).isEmpty();
```
we can write:
```java
verify(fakeClient, timeout(5000)).suggestBinding(argThat(map -> map.containsKey(CONFIG_SCOPE_ID) && map.get(CONFIG_SCOPE_ID).isEmpty()));
```
do you like it better?

damien-urruty-sonarsource: > I am not a fan of mocks in general

I was not precise enough, I was talking about mocking frameworks, which are more generic and sometimes lead to less readable constructs. This snippet is a good example I think.

In other cases you mentioned, readability will be ok and it will be easier to implement, so I'm fine ok that. 

There is another aspect that helps readability in general, with the fake client you configure everything in a single statement. As medium tests tend to involve a lot of things to set up, it's IMO better to keep things short and have the GIVEN part of a test as small as possible. With Mockito, you might have several statements to set up the client, so it will be even longer.

All this to say, I'm have mixed feelings with this move

PR #786, opened 2023-11-21
SLCORE-630 Synchronize the storage when credentials are updated

henryju:


core/src/main/java/org/sonarsource/sonarlint/core/sync/SynchronizationService.java
```diff
@@ -224,6 +226,21 @@ public void onBindingChanged(BindingConfigChangedEvent event) {
...
+  @EventListener
+  public void onConnectionCredentialsChanged(ConnectionCredentialsChangedEvent event) {
+    if (!fullSynchronizationEnabled) {
+      return;
+    }
+    var connectionId = event.getConnectionId();
+    LOG.debug("Synchronizing connection '{}' after credential changed", connectionId);
+    var bindingsForUpdatedConnection = configurationRepository.getBoundScopesByConnection(connectionId)
+      .stream()
+      .map(b -> new BoundScope(connectionId, connectionId, b.getSonarProjectKey()))
```
damien-urruty-sonarsource, line 238: This line looks suspicious, the first expected parameter is the config scope ID. Do you need to map at all? The `getBoundScopesByConnection` should return what you need

core/src/main/java/org/sonarsource/sonarlint/core/sync/SynchronizationService.java
```diff
@@ -224,6 +226,21 @@ public void onBindingChanged(BindingConfigChangedEvent event) {
     }
   }
 
+  @EventListener
+  public void onConnectionCredentialsChanged(ConnectionCredentialsChangedEvent event) {
+    if (!fullSynchronizationEnabled) {
+      return;
+    }
+    var connectionId = event.getConnectionId();
+    LOG.debug("Synchronizing connection '{}' after credential changed", connectionId);
```
damien-urruty-sonarsource, line 235: Maybe it's valid but I often see `credentials` used with plural

damien-urruty-sonarsource (approved):
Small comments, else LGTM

core/src/main/java/org/sonarsource/sonarlint/core/sync/SynchronizationService.java
henryju: Good catch. At some point, there was a mix of BoundScope and BoundConfigScope, and I remember I was struggling to convert from one to another.
Seems simpler now (maybe after you merged your changes).

PR #794, opened 2023-11-27
SLCORE-647: "Won't fix" becomes "Accept"

thahnen:
With SonarQube 10.4 and coming SonarCloud the transitions for resolving issues will be changed from **Won't fix** to **Accept**.

### DIFF
--- a/client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/backend/issue/ResolutionStatus.java
+++ b/client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/backend/issue/ResolutionStatus.java
@@ -20,7 +20,7 @@
 package org.sonarsource.sonarlint.core.clientapi.backend.issue;
 
 public enum ResolutionStatus {
-  // order is important here, it will be applied in the UI
+  ACCEPT("Accept", "The issue is valid but will not be fixed now. It represents accepted technical debt."),
   WONT_FIX("Won't Fix", "The issue is valid but does not need fixing. It represents accepted technical debt."),
   FALSE_POSITIVE("False Positive", "The issue is raised unexpectedly on code that should not trigger an issue.");
 
--- a/core/src/main/java/org/sonarsource/sonarlint/core/issue/IssueServiceImpl.java
+++ b/core/src/main/java/org/sonarsource/sonarlint/core/issue/IssueServiceImpl.java
@@ -19,7 +19,6 @@
  */
 package org.sonarsource.sonarlint.core.issue;
 
-import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -31,6 +30,7 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.inject.Named;
 import javax.inject.Singleton;
+
 import org.sonarsource.sonarlint.core.ServerApiProvider;
 import org.sonarsource.sonarlint.core.clientapi.backend.issue.AddIssueCommentParams;
 import org.sonarsource.sonarlint.core.clientapi.backend.issue.ChangeIssueStatusParams;
@@ -51,6 +51,7 @@ import org.sonarsource.sonarlint.core.local.only.XodusLocalOnlyIssueStore;
 import org.sonarsource.sonarlint.core.repository.config.ConfigurationRepository;
 import org.sonarsource.sonarlint.core.serverapi.ServerApi;
 import org.sonarsource.sonarlint.core.serverapi.proto.sonarqube.ws.Issues;
+import org.sonarsource.sonarlint.core.serverconnection.ServerInfoSynchronizer;
 import org.sonarsource.sonarlint.core.serverconnection.storage.ProjectServerIssueStore;
 import org.sonarsource.sonarlint.core.storage.StorageService;
 import org.sonarsource.sonarlint.core.telemetry.TelemetryServiceImpl;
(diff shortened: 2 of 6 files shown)


core/src/main/java/org/sonarsource/sonarlint/core/issue/IssueServiceImpl.java
```diff
@@ -175,28 +178,49 @@ public CompletableFuture<CheckStatusChangePermittedResponse> checkStatusChangePe
     }
     var issueKey = params.getIssueKey();
     var serverApi = serverApiOpt.get();
+
+    // when we checked against SonarCloud we don't have to check if it is SQ anymore
+    var isSonarCloudOrSQ104orNewer = serverApi.isSonarCloud();
+    isSonarCloudOrSQ104orNewer = isSonarCloudOrSQ104orNewer || is104orNewer(connectionId, serverApi);
```
damien-urruty-sonarsource, line 184: It looks a little bit strange to assign the variable twice, what about: 

> var isSonarCloudOrSQ104orNewer = serverApi.isSonarCloud() || is104orNewer(connectionId, serverApi);

It would read pretty well

client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/backend/issue/ResolutionStatus.java
```diff
@@ -39,4 +39,13 @@ public String getTitle() {
...
   }
+
+  public static ResolutionStatus[] get103orOlderStatuses() {
+    return new ResolutionStatus[] { ResolutionStatus.WONT_FIX, ResolutionStatus.FALSE_POSITIVE};
+  }
+
+  /** With SQ >= 10.4: "Won't fix" becomes "Accept" */
+  public static ResolutionStatus[] get104orNewerStatuses() {
+    return new ResolutionStatus[] { ResolutionStatus.ACCEPT, ResolutionStatus.FALSE_POSITIVE};
+  }
```
damien-urruty-sonarsource, line 50: I would not put those in this class, because it's not something we want to expose to clients. I would move them to the service. And you could return a List directly instead of using an array then converted to a list

core/src/main/java/org/sonarsource/sonarlint/core/issue/IssueServiceImpl.java
```diff
@@ -175,28 +178,49 @@ public CompletableFuture<CheckStatusChangePermittedResponse> checkStatusChangePe
...
     // the 2 required transitions are not available when the 'Administer Issues' permission is missing
     // normally the 'Browse' permission is also required, but we assume it's present as the client knows the issue key
     var possibleTransitions = new HashSet<>(issue.getTransitions().getTransitionsList());
+
+    final Set<String> requiredTransitions;
+    if (useNewStatuses) {
+      requiredTransitions = Set.of(Transition.ACCEPT.getStatus(), Transition.FALSE_POSITIVE.getStatus());
+    } else {
+      requiredTransitions = Set.of(Transition.WONT_FIX.getStatus(), Transition.FALSE_POSITIVE.getStatus());
+    }
```
damien-urruty-sonarsource, line 223: I have the feeling that we have duplicated knowledge here and in the enum: there are 2 places where we configure the list of allowed statuses. Maybe we could have the mapping only defined once?

damien-urruty-sonarsource (approved):
Small improvement suggestions, else LGTM

core/src/main/java/org/sonarsource/sonarlint/core/issue/IssueServiceImpl.java
thahnen: A leftover I missed

thahnen:
This PR will be on hold until the week before the release of SLI / SLE / SLVSCODE in order for SonarCloud to catch up.

core/src/main/java/org/sonarsource/sonarlint/core/issue/IssueServiceImpl.java
```diff
@@ -175,29 +181,67 @@ public CompletableFuture<CheckStatusChangePermittedResponse> checkStatusChangePe
...
+
+
+
     return asUUID(issueKey)
       .flatMap(localOnlyIssueRepository::findByKey)
       .map(r -> {
+        // For anticipated issues we currently don't get the information from SonarQube (as there is no web API
+        // endpoint) regarding the available transitions. SonarCloud doesn't provide it currently anyway.
+        // -> Here we just rely on the SonarQube version alone. God have mercy with the poor soul who has to change it
+        //    once SonarCloud is supporting anticipated transitions or SonarQube wants to change it once again :>
```
jblievremont, line 193: I don't believe that calling to a higher being is necessary here 😉; from experience, a joke in a comment is little comfort when trying to understand why a change was made / how a bug happens.

If the code is not clear about its intention, and if the goal is to express empathy with the future reader, I would rather change the code to make it clearer, if possible.

core/src/main/java/org/sonarsource/sonarlint/core/issue/IssueServiceImpl.java
```diff
@@ -175,29 +181,67 @@ public CompletableFuture<CheckStatusChangePermittedResponse> checkStatusChangePe
...
-      permitted ? null : reason,
-      // even if not permitted, return the possible statuses, if clients still want to show users what's supported
-      Arrays.asList(ResolutionStatus.values()));
+  /** Create a response for normal issues already found on SonarQube */
+  private static CheckStatusChangePermittedResponse toResponse(List<ResolutionStatus> statuses) {
+    var permitted = !statuses.isEmpty();
+
+    // When it is not permitted, we won't return the possible statuses as we don't know which statuses exist: SonarCloud
+    // might use the new transitions, might not - we don't know. For SonarQube, we could determine the available
+    // transitions but in order to have the "same" behaviour for SonarQube / SonarCloud we just return none.
```
jblievremont, line 215: If SLCORE does not return possible statuses when the user does not have the correct permissions, then we can completely remove the explicit version check.

* On future SonarCloud and SonarQube 10.4+, the permission is granted if both `accept` and `falsepositive` are there; in this case, the allowed transitions are `Accept` and `False positive`
* On current SonarCloud and pre-10.4 SonarQube, the permission is granted if both `wontfix` and `falsepositive` are there; in this case, the allowed transitions are `Won't fix` and `False positive`

So IMHO, the code should:
* Check whether `accept` is in the available transitions
* If it is there, also check for `falsepositive`, and eventually return `Accept` and `False positive`
* If _not_, check for `wontfix` and `falsepositive`

This would work for all current and foreseeable future server products.

jblievremont (changes requested):
Please see my comments below (and I'm also available for a chat if needed).

jblievremont (approved):
LGTM 👍🏻!

PR #798, opened 2023-11-28
Refactor sychronization and introduce filesystem service

henryju:


vojtech-suchy-sonarsource (approved):
I haven't noticed anything obvious that would need changing. Test coverage is good so fingers crossed 🤞

PR #803, opened 2023-12-04
Remove downloadAllServerIssues() and getServerIssues() from SonarLintEngine in favour of RPC

vojtech-suchy-sonarsource:


henryju (approved):
LGTM!

PR #804, opened 2023-12-04
Packaging cleanup (part 1)

henryju:
I have started refactoring things with the idea to isolate all non migrated stuff in a separate "legacy" module. This is not yet complete, but I think you can already review this part.

I was about to move issue tracking stuff that is only used by SLLS to the legacy package, but it was also partly used by the core, so I decided to take a chance to do a big cleanup and drop every non-used stuff.
If needed, SLLS can copy the missing classes on their side, waiting for the full issue tracking to be migrated to the core.


backend/server-connection/src/test/java/testutils/InMemoryIssueStore.java
```diff
@@ -44,7 +44,7 @@
 import static org.sonarsource.sonarlint.core.serverconnection.storage.StorageUtils.deserializeLanguages;
 
 public class InMemoryIssueStore implements ProjectServerIssueStore {
-  private final Map<String, Map<String, List<ServerIssue>>> issuesByFileByBranch = new HashMap<>();
+  private final Map<String, Map<String, List<ServerIssue<?>>>> issuesByFileByBranch = new HashMap<>();
```
damien-urruty-sonarsource, line 47: I saw some of those issues about ServerIssue being a generic class. I'm not sure it should be, but it's probably not related to your PR

spec/glossary.adoc
```diff
@@ -11,7 +11,9 @@
...
 [[effective_rule_details]]Effective Rule Details:: <<rule_definition>> but taking into account the <<binding,binding>> and the <<finding_context,finding context>>. Some attributes can be different from the <<rule_definition, rule definition>>.
-[[finding]]Finding:: Generic concept that includes all things reported by <<sonar_analyzer,Sonar analyzers>>: issues, security hotspots, taint vulnerabilities, ...
+[[finding]]Finding:: Generic concept that includes all things reported by <<sonar_analyzer,Sonar analyzers>>: issues, security hotspots, taint vulnerabilities, ... With the new Clean Code Taxonomy, security hotspots should disappear, and so everything will be an issue, so this concept might disappear.
+[[issue_matching]]Issue Matching:: Refer to the process, algorithm or xref:issue_tracking.adoc#matching_heuristic[heuristic] that we use to decide if one or multiple issues reported by an analysis are the same as the ones previously reported. Important part of <<issue_tracking,issue tracking>>.
```
damien-urruty-sonarsource, line 15: > decide if one or multiple issues reported by an analysis are the same as the ones previously reported

The wording gives the impression that the output of the process is just a yes or no, but it's more than that. What about:

> the process that takes freshly detected issues and tries to match them with already known issues. The result is a list of matched issues and a list of non-matched issues (they are for example new).

damien-urruty-sonarsource (approved):
Small comments, else LGTM

backend/server-connection/src/test/java/testutils/InMemoryIssueStore.java
henryju: We have a rule in SQ to not return wildcards. But then we have other rules (or IDE) complaining about using types without a generic type. So this I decided to go in this direction, and we can mark issues as WF on SQ (or disable the rule).

henryju: Ah yes, you are right. Let's keep that for later, as it might take some effort to refactor.

PR #810, opened 2023-12-07
SLCORE-628 Don't ask client for telemetry information we can infer on backend

vojtech-suchy-sonarsource:


backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryManager.java
```diff
@@ -46,15 +46,15 @@ TelemetryLocalStorageManager newTelemetryStorage(Path path) {
...
+  void enable(TelemetryLiveAttributesDto telemetryPayload) {
     storage.tryUpdateAtomically(data -> data.setEnabled(true));
     uploadLazily(telemetryPayload);
   }
 
   /**
    * Disable telemetry (opt-out).
    */
-  void disable(TelemetryLiveAttributesResponse telemetryPayload) {
+  void disable(TelemetryLiveAttributesDto telemetryPayload) {
```
henryju, line 57: same

backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryManager.java
```diff
@@ -67,7 +67,7 @@ void disable(TelemetryLiveAttributesResponse telemetryPayload) {
    * - the grace period has elapsed since the last upload
    * To be called periodically once a day.
    */
-  void uploadLazily(TelemetryLiveAttributesResponse telemetryPayload) {
+  void uploadLazily(TelemetryLiveAttributesDto telemetryPayload) {
```
henryju, line 70: same

core/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryServerLiveAttributesProvider.java
```diff
@@ -0,0 +1,87 @@
...
+    ConnectionConfigurationRepository connectionConfigurationRepository,
+    RulesService rulesService, RulesRepository rulesRepository) {
+    this.configurationRepository = configurationRepository;
+    this.connectionConfigurationRepository = connectionConfigurationRepository;
+    this.rulesService = rulesService;
+    this.rulesRepository = rulesRepository;
+  }
+
+  public TelemetryServerLiveAttributesDto getTelemetryServerLiveAttributes() {
+    var allBindings = configurationRepository.getConfigScopeIds().stream()
```
henryju, line 54: Can be replaced by `configurationRepository.getAllBoundScopes()` maybe?

core/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryServerLiveAttributesProvider.java
```diff
@@ -0,0 +1,87 @@
...
+      .map(configurationRepository::getBindingConfiguration)
+      .filter(bindingConfiguration -> bindingConfiguration != null && bindingConfiguration.getConnectionId() != null)
+      .collect(Collectors.toList());
+
+    var usesConnectedMode = !allBindings.isEmpty();
+
+    var usesSonarCloud = allBindings.stream()
+      .anyMatch(binding -> connectionConfigurationRepository.getConnectionById(binding.getConnectionId()) instanceof SonarCloudConnectionConfiguration);
```
henryju, line 62: I don't like this "instanceOf". We could maybe add a `isSonarCloud` attribute on the `BoundScope`? I remember it would have been useful for another use case.

medium-tests/src/test/java/mediumtest/TelemetryMediumTests.java
```diff
@@ -179,7 +178,7 @@ void it_should_ping_telemetry_endpoint() throws ExecutionException, InterruptedE
         "  \"ide_version\" : \"4.5.6\",\n" +
         "  \"platform\" : \"linux\",\n" +
         "  \"architecture\" : \"x64\",\n" +
-        "  \"connected_mode_used\" : true,\n" +
+        "  \"connected_mode_used\" : false,\n" +
```
henryju, line 181: Why is it not `true` ? There is a connection and a binding declared.

medium-tests/src/test/java/mediumtest/TelemetryMediumTests.java
```diff
@@ -206,7 +205,7 @@ void it_should_disable_telemetry() throws ExecutionException, InterruptedExcepti
...
```
henryju, line 208: Same question

medium-tests/src/test/java/mediumtest/TelemetryMediumTests.java
vojtech-suchy-sonarsource: This is an interesting one.  We set a `System.setProperty("sonarlint.internal.telemetry.initialDelay", "0")` at the start of the test so we don't need to wait for the telemetry scheduling to start in one minute (that's the default).
In this case, this means the telemetry endpoint is contacted pretty much immediately at the point when we build the backend - even before the [configuration is added](https://github.com/SonarSource/sonarlint-core/blob/task/vs/slcore-628_get_telemetry_from_core/medium-tests/src/test/java/mediumtest/fixtures/SonarLintBackendFixture.java#L353).

So the test used to return true just because that's what we mocked the client to return in the live attributes 
` when(fakeClient.getTelemetryLiveAttributes()).thenReturn(new TelemetryLiveAttributesResponse(true, false, null, false, emptyList(), emptyList(), emptyMap()));`  (and declaring the connection and binding on the backend didn't have impact).
Since the `connected_mode_used` value is now computed on the backend, it got correctly computed to false as there is no binding at the moment of the computation.
Rather than trying to fiddle with the initial scheduling delay for Telemetry I think it's ok to have `false` returned for this test (as the actual computation is tested by other tests). I will just remove the `.....withSonarQubeConnection("connectionId").withBoundConfigScope("scopeId", "connectionId", "projectKey")` bit so it's not confusing.

henryju: Sounds good to me, this will be less confusing indeed!

PR #815, opened 2023-12-08
SLCORE-624 Notify clients when the Sloop process is killed

damien-urruty-sonarsource:


vojtech-suchy-sonarsource (approved):
Nice, :shipit:

vojtech-suchy-sonarsource:
Still looking good 👍

PR #818, opened 2023-12-14
SLCORE-661 Support analysis of Gradle Kotlin DSL files

nquinquenel:


damien-urruty-sonarsource (approved):
LGTM, I think you could have at least one IT, checking that we are able to detect kts issues

PR #821, opened 2023-12-15
SLCORE-611 Test .kts file analysis and issue reporting

nquinquenel:


thahnen (approved):
LGTM if the test now passes 😛

PR #823, opened 2023-12-15
SLCORE-631 Stop asking client to manipulate server file paths

kirill-knize-sonarsource:


backend/core/src/test/java/org/sonarsource/sonarlint/core/file/PathTranslationServiceTests.java
```diff
@@ -0,0 +1,54 @@
...
+import org.sonarsource.sonarlint.core.ServerApiProvider;
+import org.sonarsource.sonarlint.core.fs.ClientFileSystemService;
+import org.sonarsource.sonarlint.core.repository.config.ConfigurationRepository;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+class PathTranslationServiceTests {
```
henryju, line 35: Shouldn't you test the retry mechanism?

backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/storage/XodusServerIssueStore.java
```diff
@@ -714,6 +716,13 @@ private static Optional<Entity> findUniqueAmong(EntityIterable iterable, String
       .findFirst();
   }
 
+  private static Optional<Entity> findUniquePathAmong(EntityIterable iterable, String propertyName, Path path) {
+    return StreamSupport.stream(iterable.spliterator(), false)
+      .filter(e -> path.equals(Path.of((String) e.getProperty(propertyName))))
```
henryju, line 721: :+1:

henryju (changes requested):
Good work overall. I don't think I will have time for an in-depth review, so I left a few comments already, and then please ask another review from @damien-urruty-sonarsource before merging (or Damien can merge next week).
I am not sure about the need for the retry mechanism in the Path translation service. I think it could be a separate task if we feel the need.

damien-urruty-sonarsource (approved):
One small comment, else LGTM

PR #824, opened 2023-12-18
Add "SONARLINT_VM_OPTS" to startup scripts

vojtech-suchy-sonarsource:


henryju (approved):
You can add maybe an integration test. I remember testing something similar by passing a very small Xmx value so that the process would die with OOM (that would allow to assert that the passed value was taken into account).

PR #825, opened 2023-12-18
OSGI packaging

henryju:


damien-urruty-sonarsource (approved):
LGTM, waiting for Tobias's check on SLE side before merging

client/osgi/pom.xml
```diff
@@ -58,10 +59,7 @@
               org.slf4j.spi;version="1.2.3",
             </Import-Package>
             <Export-Package>
-              org.sonarsource.sonarlint.core.*;version="${project.version}",
-              org.sonarsource.sonarlint.shaded.com.google.gson.*;version="${gson.version}",
-              com.google.protobuf;version="${protobuf.version}",
-              org.eclipse.lsp4j.jsonrpc.*;version="${lsp4j.version}",
+              org.sonarsource.sonarlint.core.client.*;version="${project.version}",
```
thahnen, line 62: Don't we also have to export `org.sonarsource.sonarlint.core.commons.*` and `org.sonarsource.sonarlint.core.rpc.*`?
I tried using the bundle in SLE and it didn't work (mostly due to other reasons but also because when accessing stuff in both this packages Tycho complained about it not being API).

client/embedded/pom.xml
```diff
@@ -0,0 +1,132 @@
...
+                  </includes>
+                  <excludes>
+                    <exclude>**/JsonWriter.class</exclude>
+                    <exclude>**/internal/google/gson/**</exclude>
+                    <!-- Overridden by SonarLint -->
+                    <exclude>org/sonar/api/utils/log/Loggers.class</exclude>
+                  </excludes>
+                </filter>
+                <filter>
+                  <!-- Shade the SonarLint Plugin API (avoids minimization) as it is used by plugins -->
```
thahnen, line 94: Both shading the API and ProtoBuf didn't work, it is not present on the artifact.

thahnen (changes requested):
The shading seems not to work correctly and the `Export-Package` statement is not yet correct.

client/embedded/pom.xml
henryju: OK, I will fix that in another PR

PR #827, opened 2023-12-19
 Invalidate file exclusions after sync if file inclusion/exclusion settings changed

vojtech-suchy-sonarsource:


backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/LocalStorageSynchronizer.java
```diff
@@ -74,12 +75,28 @@ public boolean synchronize(ServerApi serverApi) {
...
+
+    storage.project(projectKey).analyzerConfiguration().store(updatedAnalyzerConfiguration);
     var version = storage.serverInfo().read().orElseThrow().getVersion();
     serverApi.newCodeApi().getNewCodeDefinition(projectKey, null, version)
       .ifPresent(ncd -> storage.project(projectKey).newCodeDefinition().store(ncd));
+    return configUpdateSummary;
+  }
+
+  private static AnalyzerConfigUpdateSummary diffAnalyzerConfiguration(AnalyzerConfiguration original, AnalyzerConfiguration updated) {
+    var originalFileExclusions = original.getSettings().getAll().get("sonar.exclusions");
```
henryju, line 96: I don't like to have the responsibility of knowing what are the properties involved in the exclusion process here. This should be in the `FileExclusionService`.
I would raise an event as soon as some property changed in the local storage (with a diff, or simply the old and new maps). Then the `FileExclusionService` would receive this event, and look if one of the exclusions/inclusions properties have been updated.

backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/LocalStorageSynchronizer.java
```diff
@@ -74,12 +75,28 @@ public boolean synchronize(ServerApi serverApi) {
...
+    var updatedFileExclusions = updated.getSettings().getAll().get("sonar.exclusions");
```
henryju, line 97: BTW there are a few other properties involved in the computation of exclusions:
* sonar.inclusions
* sonar.test.exclusions
* sonar.test.inclusions

core/src/main/java/org/sonarsource/sonarlint/core/fs/FileExclusionService.java
```diff
@@ -114,6 +115,12 @@ public void onFileSystemUpdated(FileSystemUpdatedEvent event) {
     event.getAddedOrUpdated().forEach(f -> serverExclusionByUri.invalidate(f.getUri()));
   }
 
+  @EventListener
+  public void onFileExclusionSettingsChanged(FileExclusionChangedEvent event) {
```
henryju, line 119:
```suggestion
  public void onSonarServerSettingsChanged(SonarServerChangedEvent event) {
```
I think the event should represent a change of the settings on the Sonar server side.

core/src/main/java/org/sonarsource/sonarlint/core/fs/FileExclusionService.java
```diff
@@ -114,6 +115,12 @@ public void onFileSystemUpdated(FileSystemUpdatedEvent event) {
...
+    event.getConfigScopeIds().forEach(configScopeId -> clientFileSystemService.getFiles(configScopeId)
```
henryju, line 120: Maybe add a debug log

core/src/main/java/org/sonarsource/sonarlint/core/sync/SynchronizationService.java
```diff
@@ -240,7 +248,12 @@ private void synchronizeConnectionAndProjectsIfNeeded(String connectionId, Serve
       var scopesPerProjectKey = scopesToSync.stream().collect(groupingBy(BoundScope::getSonarProjectKey, mapping(BoundScope::getConfigScopeId, toSet())));
       scopesPerProjectKey.forEach((projectKey, configScopeIds) -> {
         LOG.debug("Synchronizing storage of Sonar project '{}' for connection '{}'", projectKey, connectionId);
-        serverConnection.sync(serverApi, projectKey);
+        var analyzerConfigUpdateSummary = serverConnection.sync(serverApi, projectKey);
+        if(analyzerConfigUpdateSummary != null) {
```
henryju, line 252: Code style

its/tests/src/test/java/its/FileExclusionTests.java
```diff
@@ -0,0 +1,221 @@
...
+
+    // Change file exclusion settings on SQ which should not affect Foo.java
+    adminWsClient.settings().set(new SetRequest()
+      .setKey("sonar.exclusions")
+      .setValues(singletonList("**/*.js"))
+      .setComponent(projectKey));
+
+    // Check Foo.java is still included
+    await().atMost(10, SECONDS).untilAsserted(() ->
+      assertThat(backend.getFileService().getFilesStatus(getFilesStatusParams).get().getFileStatuses().get(filePath.toUri()).isExcluded()).isFalse());
```
henryju, line 181: I think this assertion is useless because anyway, it is unlikely the new exclusion would be processed that fast.

What you could do is move this case after a case where the file was excluded:

assert excluded = false

set sonar.exclusion = **/*.java
await excluded = true

set sonar.exclusion = **/*.js
await excluded = false

set sonar.inclusion = **/*.js
await excluded = true

reset sonar.inclusion and sonar.exclusion
await excluded = false

henryju (changes requested):
Nice integration test. I just wonder if a medium test would not be enough, since there is little integration with SonarQube for this feature (except the sync of settings).

Overall looks good, my main issue is with the granularity of the event and the responsibility of deciding which properties are interesting.

backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/LocalStorageSynchronizer.java
vojtech-suchy-sonarsource: Good idea. In [157701d](https://github.com/SonarSource/sonarlint-core/pull/827/commits/157701d423cb61219539615451f78474538d5cb5) we make diff on the ServerSettings and send the summary to the `SynchronizationService` which then raises `SonarServerSettingsChangedEvent`.
The reason for raising the event in `SynchronizationService` is that the `LocalStorageSynchronizer` is not a bean (and is used from `ServerConnection` which is used from the `ConnectedEngine` so it wouldn't be straightforward making it a bean)

core/src/main/java/org/sonarsource/sonarlint/core/fs/FileExclusionService.java
```diff
@@ -114,6 +115,22 @@ public void onFileSystemUpdated(FileSystemUpdatedEvent event) {
     event.getAddedOrUpdated().forEach(f -> serverExclusionByUri.invalidate(f.getUri()));
   }
 
+  private static boolean isFileExclusionSettingsDifferent(Map<String, String> updatedSettingsValueByKey) {
+    return updatedSettingsValueByKey.containsKey("sonar.exclusions") ||
```
henryju, line 119:
```suggestion
 // See org.sonar.api.scan.filesystem.FileExclusions
  public static final Set<String> ALL_EXCLUSION_RELATED_SETTINGS = Set.of(CoreProperties.PROJECT_INCLUSIONS_PROPERTY, CoreProperties.PROJECT_TEST_INCLUSIONS_PROPERTY, CoreProperties.GLOBAL_EXCLUSIONS_PROPERTY, CoreProperties.PROJECT_EXCLUSIONS_PROPERTY, CoreProperties.GLOBAL_TEST_EXCLUSIONS_PROPERTY, CoreProperties.PROJECT_TEST_EXCLUSIONS_PROPERTY); 
    return ALL_EXCLUSION_RELATED_SETTINGS.stream.anyMatch(updatedSettingsValueByKey::containsKey);
```

henryju (approved):
I forgot global exclusions properties, but then LGTM

PR #828, opened 2023-12-20
Deploy embedded CLI as correct dependency

thahnen:
## Summary

In order for the flavors (e.g. SonarLint for Eclipse) to consume the embedded CLI packages correctly, they must be published not as assemblies "added" to the Maven module JAR artifact but alongside their own POM. This way they can be consumed like normal Maven artifacts (e.g. in Eclipse Target Platform, like the Embedded Analyzers being copied locally).


thahnen:
Misconfiguration on IDE side, nothing to be changed on SLOCRE

PR #829, opened 2023-12-21
Add custom adapters to remove the need for Either wrapping

henryju:
Waiting for the final decision, if we finally decide to keep Eithers, here is a fix to remove the ugly "wrapper" I had to introduce since I was not able to properly deserialize `Map<String, List<Either<A, B>>`.


rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/SonarLintLauncherBuilder.java
```diff
@@ -49,10 +55,9 @@ protected MessageJsonHandler createJsonHandler() {
       public GsonBuilder getDefaultGsonBuilder() {
         // We don't want the EnumTypeAdapter from lsp4j, as we want to serialize enums as string (this is the default in Gson)
         return new GsonBuilder()
-          .registerTypeAdapterFactory(new CollectionTypeAdapter.Factory())
-          .registerTypeAdapterFactory(new ThrowableTypeAdapter.Factory())
           .registerTypeAdapterFactory(new EitherTypeAdapter.Factory())
-          .registerTypeAdapterFactory(new TupleTypeAdapters.TwoTypeAdapterFactory())
+          .registerTypeAdapterFactory(new EitherServerOrLocalHotspotAdapterFactory())
```
henryju, line 59: The order is important here: the last added will have higher priority. So we should add more specific Either adapters after the generic `EitherTypeAdapter`.

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/adapter/EitherServerOrLocalHotspotAdapterFactory.java
```diff
@@ -0,0 +1,59 @@
...
+  private static final TypeToken<Either<ServerMatchedSecurityHotspotDto, LocalOnlySecurityHotspotDto>> ELEMENT_TYPE = new TypeToken<>() {
+  };
+
+  @SuppressWarnings("unchecked")
+  @Override
+  public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> type) {
+    if (!TypeUtils.isEither(type.getType())) {
+      return null;
+    }
+    Type[] typeParameters = ((ParameterizedType) type.getType()).getActualTypeArguments();
```
henryju, line 49: We have to check the generic types so that this adapter will only be used for `Either<ServerMatchedSecurityHotspotDto, LocalOnlySecurityHotspotDto>`

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/SonarLintLauncherBuilder.java
damien-urruty-sonarsource: Maybe worth adding a comment in the code?

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/adapter/EitherServerOrLocalIssueAdapterFactory.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+    Type[] typeParameters = ((ParameterizedType) type.getType()).getActualTypeArguments();
+    var leftType = typeParameters[0];
+    var rightType = typeParameters[1];
+    if (!ServerMatchedIssueDto.class.isAssignableFrom((Class<?>) leftType) && !LocalOnlyIssueDto.class.isAssignableFrom((Class<?>) rightType)) {
+      return null;
+    }
+    Predicate<JsonElement> serverChecker = new EitherTypeAdapter.PropertyChecker("serverKey");
+    Predicate<JsonElement> localChecker = not(serverChecker);
+    return (TypeAdapter<T>) new EitherTypeAdapter<>(gson, ELEMENT_TYPE, serverChecker, localChecker);
```
damien-urruty-sonarsource, line 57: The 2 Either adapters have a lot in common, do we really need 2 separate adapters?

damien-urruty-sonarsource (approved):
A few minor comments, but LGTM! Should we also add a comment somewhere in the code about why we need dedicated adapters, and why we can't reuse the provided EitherTypeAdapter?

PR #832, opened 2024-01-02
SLCORE-665: Automatic connected mode setup

thahnen:
When using the "Open in IDE" feature on SonarQube, we want to get a token (if possible) for an automated connected mode setup. It should still work with older SonarQube servers not able to generate a token as well as opening Security Hotspots in the IDE - this feature won't provide a token.
The changes on the SonarQube side were already done (**old SonarLint - new SonarQube** and **new SonarLint - old SonarQube** will work).

Follow-up ticket: [SLCORE-666 Make token revocation transparent for the client](https://sonarsource.atlassian.net/browse/SLCORE-666)

### DIFF
--- a/client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/client/connection/AssistCreatingConnectionParams.java
+++ b/client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/client/connection/AssistCreatingConnectionParams.java
@@ -19,15 +19,32 @@
  */
 package org.sonarsource.sonarlint.core.clientapi.client.connection;
 
+import javax.annotation.Nullable;
+
 public class AssistCreatingConnectionParams {
   private final String serverUrl;
+  @Nullable
+  private final String tokenName;
+  @Nullable
+  private final String tokenValue;
 
-  public AssistCreatingConnectionParams(String serverUrl) {
+  public AssistCreatingConnectionParams(String serverUrl, @Nullable String tokenName, @Nullable String tokenValue) {
     this.serverUrl = serverUrl;
+    this.tokenName = tokenName;
+    this.tokenValue = tokenValue;
   }
 
   public String getServerUrl() {
     return serverUrl;
   }
 
+  @Nullable
+  public String getTokenName() {
+    return tokenName;
+  }
+
+  @Nullable
+  public String getTokenValue() {
+    return tokenValue;
+  }
 }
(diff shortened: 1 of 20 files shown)


client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/client/connection/AssistCreatingConnectionParams.java
```diff
@@ -19,15 +19,32 @@
  */
 package org.sonarsource.sonarlint.core.clientapi.client.connection;
 
+import javax.annotation.Nullable;
+
 public class AssistCreatingConnectionParams {
   private final String serverUrl;
+  @Nullable
+  private final String tokenName;
```
thahnen, line 27: I've included the `tokenName` alongside the `tokenValue` if at some point we want to store this information on the IDE side. As an example: When the user uses the "Open in IDE" feature but cancels (I dunno why) and then proceeds to connect manually, we could already have the token in stock. Of course only for some time and not persistent ^^

thahnen: The `tokenName` is now also used when revoking tokens from the IDE side after a failed automatic connected mode setup.

its/tests/src/test/java/its/SonarQubeCommunityEditionTests.java
```diff
@@ -152,6 +154,19 @@ static void createSonarLintUser() {
...
+  void test_revoke_token() {
+    var tokenName = "testTokenForRevoking";
+    var testToken = adminWsClient.userTokens()
+      .generate(new GenerateRequest().setName(tokenName));
+    assertThat(testToken.getName()).isEqualTo(tokenName);
+
+    // Using the credentials we used throughout the test, we want to remove the additional token!
+    backend
+      .getUserTokenService()
+      .revokeToken(new RevokeTokenParams(ORCHESTRATOR.getServer().getUrl(), testToken.getName()));
```
sophio-japharidze-sonarsource, line 167: is an assertion missing here? 🤔

thahnen: Yes you're right, I did not commit the changes here.

sophio-japharidze-sonarsource (approved):
LGTM with a future improvement in [SLCORE-666](https://sonarsource.atlassian.net/browse/SLCORE-666)

PR #836, opened 2024-01-16
Hardening (WIP)

henryju:
* Use Failsafe library instead of custom retry and for cancellation propagation
* Log RPC messages in medium tests (TODO: should probably be optional as it is very verbose)
* Added Thread name to logs
* Tried to avoid cascading synchronization (sync -> branch matching -> sync) (TODO: not reliable solution)
* Moved synchronization tasks to the executor service


medium-tests/src/test/java/mediumtest/fixtures/SonarLintBackendFixture.java
```diff
@@ -378,11 +379,66 @@ public SonarLintTestRpcServer build(SonarLintRpcClientDelegate client) {
...
+        public int nextContentSize = -1;
+
+        @Override
+        public void write(@NotNull byte[] b) throws IOException {
+          mem.append(new String(b));
+          flushIfNeeded();
+          super.write(b);
+        }
+
+        public void flushIfNeeded() {
```
damien-urruty-sonarsource, line 394: Why public?

medium-tests/src/test/java/mediumtest/fixtures/SonarLintBackendFixture.java
```diff
@@ -378,11 +379,66 @@ public SonarLintTestRpcServer build(SonarLintRpcClientDelegate client) {
...
+        @Override
+        public synchronized int read(byte[] b, int off, int len) throws IOException {
+          int readLength = super.read(b, off, len);
+          if (readLength > 0) {
+            System.out.println("<-- " + new String(b, off, readLength));
+          }
+          return readLength;
+        }
+
+        public void flushIfNeeded() {
```
damien-urruty-sonarsource, line 426: Is this one used at all?

backend/core/src/main/java/org/sonarsource/sonarlint/core/commons/SmartCancelableLoadingCache.java
```diff
@@ -0,0 +1,186 @@
...
+              oldValueFuture.complete(newValue);
+              valueFuture.complete(newValue);
+            }
+          }
+        } finally {
+          lock.writeLock().unlock();
+        }
+      }, executorService);
+      computeFuture = result;
+      lock.writeLock().unlock();
```
damien-urruty-sonarsource, line 100: To be on the safe side, I would put all unlocking in a `finally` block, maybe nothing throws at the moment but later something could. This comment applies for all unlocks in this file

backend/core/src/main/java/org/sonarsource/sonarlint/core/commons/SmartCancelableLoadingCache.java
```diff
@@ -0,0 +1,186 @@
...
+  private static final SonarLintLogger LOG = SonarLintLogger.get();
+
+  private final ExecutorService executorService;
+  private final String threadName;
+  private final BiFunction<K, SonarLintCancelChecker, V> valueComputer;
+  private final ConcurrentHashMap<K, ValueAndComputeFutures> cache = new ConcurrentHashMap<>();
+
+  private class ValueAndComputeFutures {
+    private final K key;
+    private CompletableFuture<V> valueFuture = new CompletableFuture<>();
```
damien-urruty-sonarsource, line 50: I always have a doubt, shouldn't this one be volatile? Or use an AtomicReference?

backend/core/src/main/java/org/sonarsource/sonarlint/core/commons/SmartCancelableLoadingCache.java
```diff
@@ -0,0 +1,186 @@
...
+      lock.readLock().lock();
+      try {
+        return valueFuture;
+      } finally {
+        lock.readLock().unlock();
+      }
+    }
+
+    public void cancel() {
+      lock.writeLock().lock();
```
damien-urruty-sonarsource, line 118: This means there is no possible cancelation while a computation is in progress, right?

damien-urruty-sonarsource (approved):
Probably some small problems with the current implementation. Overall I think the smart cache complexity is pretty high. This makes me feel that Sloop's code is slowly becoming more and more complex :(
As I tested it with SLI, it works, so I think this PR can be merged (after some fixes). I'm a bit concerned about the direction all this is taking. We can discuss this later if you want

backend/core/src/main/java/org/sonarsource/sonarlint/core/commons/SmartCancelableLoadingCache.java
```diff
@@ -0,0 +1,186 @@
...
+            if (error != null) {
+              oldValueFuture.completeExceptionally(error);
+              valueFuture.completeExceptionally(error);
+            } else {
+              oldValueFuture.complete(newValue);
+              valueFuture.complete(newValue);
+            }
+          }
+        } finally {
+          lock.writeLock().unlock();
```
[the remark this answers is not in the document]
henryju: The code in between is only swapping future and scheduling a new computation. It should be fast and non blocking.

backend/core/src/main/java/org/sonarsource/sonarlint/core/fs/ClientFileSystemService.java
```diff
@@ -58,17 +58,16 @@ public class ClientFileSystemService {
 
   private final ExecutorService executorService = Executors.newSingleThreadExecutor(r -> new Thread(r, "sonarlint-filesystem"));
 
-  private final AsyncLoadingCache<String, Map<URI, ClientFile>> filesByConfigScopeId = Caffeine.newBuilder()
-    .executor(executorService)
-    .buildAsync(this::initializeFileSystem);
+  private final SmartCancelableLoadingCache<String, Map<URI, ClientFile>> filesByConfigScopeIdCache =
```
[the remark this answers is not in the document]
henryju: Good catch!

backend/core/src/main/java/org/sonarsource/sonarlint/core/fs/ClientFileSystemService.java
```diff
@@ -58,17 +58,16 @@ public class ClientFileSystemService {
 
   private final ExecutorService executorService = Executors.newSingleThreadExecutor(r -> new Thread(r, "sonarlint-filesystem"));
```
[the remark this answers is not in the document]
henryju: Nope, good catch!

medium-tests/src/test/java/mediumtest/fixtures/SonarLintBackendFixture.java
henryju: Oups

medium-tests/src/test/java/mediumtest/fixtures/SonarLintBackendFixture.java
henryju: Nope, this is a remain from a first attempt

backend/core/src/main/java/org/sonarsource/sonarlint/core/commons/SmartCancelableLoadingCache.java
henryju: Yes, I think it is a good idea

backend/core/src/main/java/org/sonarsource/sonarlint/core/commons/SmartCancelableLoadingCache.java
henryju: No, the write lock is only taken to schedule a new computation, since we might have to swap the completable future. The computation itself runs out of the lock.

PR #837, opened 2024-01-18
SLCORE-668 Add textenterprise to the plugin allow list in connected mode

jblievremont:


plugin-commons/src/main/java/org/sonarsource/sonarlint/core/plugin/commons/PluginsLoader.java
```diff
@@ -79,10 +81,17 @@ public PluginsLoadResult load(Configuration configuration) {
...
     var pluginInstancesByKeys = instancesLoader.instantiatePluginClasses(nonSkippedPlugins);
 
-    return new PluginsLoadResult(new LoadedPlugins(pluginInstancesByKeys, instancesLoader, maybeDbdAllowedPlugins(configuration.enableDataflowBugDetection)),
+    return new PluginsLoadResult(new LoadedPlugins(pluginInstancesByKeys, instancesLoader, additionalAllowedPlugins(configuration)),
       pluginCheckResultByKeys);
   }
 
+  private static Set<String> additionalAllowedPlugins(Configuration configuration) {
+    var allowedPluginsIds = new HashSet<String>();
+    allowedPluginsIds.add("textenterprise");
```
thahnen, line 90: Will it work when we allow the `sonar-text-enterprise` but SQ does not have it in stock? Like community or older editions?

jblievremont: Yes: this is only a list used to filter which plugins are allowed among the ones downloaded from the server.

thahnen: Perfect, I wasn't quite sure about this.

thahnen (approved):
LGTM

PR #838, opened 2024-01-19
SLCORE-671 Suggest automatic binding following connection creation

nquinquenel:


client-api/src/main/java/org/sonarsource/sonarlint/core/clientapi/client/binding/NoBindingSuggestionFoundParams.java
```diff
@@ -0,0 +1,37 @@
...
+package org.sonarsource.sonarlint.core.clientapi.client.binding;
+
+import org.eclipse.lsp4j.jsonrpc.validation.NonNull;
+
+public class NoBindingSuggestionFoundParams {
+
+  @NonNull
+  private final String projectKey;
+
+  public NoBindingSuggestionFoundParams(String projectKey) {
```
nquinquenel, line 29: We could also send the text to display. The only unique thing is the action for showing the troubleshooting documentation, it differs for each IDE.

thahnen (approved):
Looks good to me.

thahnen (approved):
Took some more time to understand the changes in-depth. LGTM, nice work!
I think for now we will go with the current implementation (of course with your changes) of a 2-3-step setup.
But I'll take it with me for the future after Sloop is bundled that we maybe can re-work it in Eclipse to maybe get rid of one step 👍

PR #841, opened 2024-01-23
NO-JIRA: Remove flaky integration test

thahnen:
The test was added with #832 but turned out to be too flaky.

### DIFF
--- a/its/tests/src/test/java/its/SonarCloudTests.java
+++ b/its/tests/src/test/java/its/SonarCloudTests.java
@@ -23,7 +23,6 @@ import java.io.File;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.time.Duration;
 import java.time.Instant;
 import java.util.Collections;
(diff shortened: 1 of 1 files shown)


its/tests/src/test/java/its/SonarCloudTests.java
```diff
@@ -471,6 +472,7 @@ void testConnection() throws ExecutionException, InterruptedException {
   }
 
   @Test
+  @Disabled("Because the the feature is intended for SQ and the IT was only added 'on-top' but is too flaky!")
```
henryju, line 475: Can't we simply drop it then? I am a bit worried by all those @disabled tests.

thahnen: I'm in favor, I see SLCORE already has some ignored / disabled tests. At some point we should get rid of the rest too.

thahnen: Done

PR #846, opened 2024-01-26
Fix OSGi metadata

henryju:


backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/log/LogOutput.java
```diff
@@ -21,13 +21,29 @@
...
-  void log(String formattedMessage, Level level);
+  /**
+   * @deprecated please implement {@link #log(String, Level, String)} instead
+   */
+  @Deprecated(since = "10.0")
+  default void log(String formattedMessage, Level level) {
+    log(formattedMessage, level, null);
+  }
+
+  default void log(@Nullable String formattedMessage, Level level, @Nullable String stacktrace) {
```
damien-urruty-sonarsource, line 39: I was a bit surprised with this API, why not take the Throwable here instead of the stacktrace?

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/log/LogOutput.java
```diff
@@ -21,13 +21,29 @@
...
  * Allow to redirect SonarLint logs to a custom output on client side
  */
 public interface LogOutput {
 
-  void log(String formattedMessage, Level level);
+  /**
+   * @deprecated please implement {@link #log(String, Level, String)} instead
+   */
+  @Deprecated(since = "10.0")
+  default void log(String formattedMessage, Level level) {
```
damien-urruty-sonarsource, line 35: Why deprecate this method? I assume most places logging won't have a stack trace, so it could act as a shortcut for the other one

backend/core/src/main/java/org/sonarsource/sonarlint/core/NodeJsHelper.java
```diff
@@ -163,7 +161,7 @@ private String runSimpleCommand(Command command) {
     if (!stdErr.isEmpty()) {
       msg.append("\nstderr: ").append(String.join("\n", stdErr));
     }
-    logOutput.log(msg.toString(), LogOutput.Level.DEBUG);
+    LOG.debug("{}", msg);
```
damien-urruty-sonarsource, line 164: -> LOG.debug(msg);

damien-urruty-sonarsource (approved):
Minor comments, else LGTM

damien-urruty-sonarsource:
And there are SQ issues and a bit low coverage

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/log/LogOutput.java
henryju: I guess most clients will print the stacktrace anyway. This will save them to implement the "stacktraceToString"

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/log/LogOutput.java
henryju: The goal is for clients to have only one method to implement. In fact, I could have removed it totally since we are not really concerned about backward compatibility in 10.x...

backend/core/src/main/java/org/sonarsource/sonarlint/core/NodeJsHelper.java
henryju: Nope, msg is a StringBuilder. `LOG.debug(msg.toString())` raises an issue about performance.
With slf4j2 we will be able to do:
`LOG.debug(() -> msg.toString())` to not pay the cost of the toString when DEBUG is not enabled

henryju:
@vojtech-suchy-sonarsource Can you review the last commit of this PR please? I had to rework the SloopLauncher to have a better coverage.

PR #847, opened 2024-01-26
SLCORE-668: Fix download of sonar-text

thahnen:
## Summary

There was an issue that connecting SonarLint to SQ 10.3 and earlier worked the same as before. However, with 10.4, the `sonar-text` analyzer wasn't downloaded anymore because it was still seen as embedded while the `PluginsSynchonizer` was synchronizing the plug-ins for the connection.

The [first approach](https://github.com/SonarSource/sonarlint-core/commit/6aeaf961919d1eb5cd880b50c9da41168d119e1d) to mitigate this issue only worked when the connection was established; the IDE was closed and then opened again for the synchronization to start again. This was due to the check to remove the embedded instance of `sonar-text` not being able to access the server information yet to check the SQ version.
The synchronization of the SQ server information was not finished the first time, but when the IDE was restarted, the information was already present.

With this PR, the changes are now finally in place to wait for the server information to be synchronized before making the check on SQ 10.4+ (when we don't use the embedded `sonar-text` but rather fetch the one from the server alongside `sonar-text-enterprise`) and before handing everything over to the `PluginsSynchonizer` to synchronize the plug-ins to be downloaded.

## Testing

Due to SQ 10.4 DEV currently not yet including the new `sonar-text` and `sonar-text-enterprise` versions, which support SonarLint via the manifest attributes, there are unit as well as medium tests in place to emulate the behavior of having SQ 10.4 and SQ 10.3!
Additionally, manual tests were done with the new `sonar-text` and `sonar-text-enterprise` versions and SQ 10.4 DEV locally, where the plug-ins were exchanged alongside manual testing against SQ 10.3 and SQ 10.2 where `sonar-text` was not to be downloaded but where the embedded one should be used.

### DIFF
--- a/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/LocalStorageSynchronizer.java
+++ b/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/LocalStorageSynchronizer.java
@@ -52,7 +52,13 @@ public class LocalStorageSynchronizer {
     serverInfoSynchronizer.synchronize(serverApi);
     var version = storage.serverInfo().read().orElseThrow().getVersion();
 
-    var anyPluginUpdated = pluginsSynchronizer.synchronize(serverApi, progressMonitor);
+    // INFO: In order to download `sonar-text` alongside `sonar-text-enterprise` on SQ 10.4+ we have to change the
+    //       plug-in synchronizer to work correctly the moment the connection is established and the plug-ins are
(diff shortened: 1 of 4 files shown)


jblievremont (approved):
LGTM!

Tested in SLVSCODE:
 * Standalone mode uses the embedded analyzer
 * Connected mode with SQ 9.9 _also_ uses embedded analyzer
 * Connected mode with SQ 10.4 uses the one from the server (with custom rule enabled)

PR #849, opened 2024-01-29
SLCORE-683 Fix race condition in the SmartCancellableLoadingCache

henryju:
I can reproduce using a debugger to force a certain execution order, but not easy to write a unit test. It was failing randomly on CI.

### DIFF
--- a/backend/core/src/main/java/org/sonarsource/sonarlint/core/commons/SmartCancelableLoadingCache.java
+++ b/backend/core/src/main/java/org/sonarsource/sonarlint/core/commons/SmartCancelableLoadingCache.java
@@ -47,13 +47,12 @@ public class SmartCancelableLoadingCache<K, V> implements AutoCloseable {
 
   private class ValueAndComputeFutures {
     private final K key;
-    private volatile CompletableFuture<V> valueFuture = new CompletableFuture<>();
-    private volatile CompletableFuture<V> computeFuture;
+    private CompletableFuture<V> valueFuture = new CompletableFuture<>();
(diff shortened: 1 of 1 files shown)


backend/core/src/main/java/org/sonarsource/sonarlint/core/commons/SmartCancelableLoadingCache.java
```diff
@@ -47,13 +47,12 @@ public class SmartCancelableLoadingCache<K, V> implements AutoCloseable {
 
   private class ValueAndComputeFutures {
     private final K key;
-    private volatile CompletableFuture<V> valueFuture = new CompletableFuture<>();
-    private volatile CompletableFuture<V> computeFuture;
+    private CompletableFuture<V> valueFuture = new CompletableFuture<>();
```
henryju, line 50: volatile are not needed since we are using a Lock

backend/core/src/main/java/org/sonarsource/sonarlint/core/commons/SmartCancelableLoadingCache.java
```diff
@@ -179,19 +178,27 @@ public void clear(K key) {
    * Awaiting #get() will receive the newly computed value
    */
   public void refreshAsync(K key) {
-    var valueAndComputeFutures = cache.get(key);
```
henryju, line 182: The race condition was here, between the `get` and the `computeIfAbsent`

damien-urruty-sonarsource (approved):

PR #853, opened 2024-02-01
SLCORE-396 Notify clients when analysis is ready to be triggered

damien-urruty-sonarsource:


backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisService.java
```diff
@@ -340,4 +355,81 @@ private static Predicate<? super SonarLintRuleDefinition> isIncludedByConfigurat
...
+  public void onConfigurationScopeRemoved(ConfigurationScopeRemovedEvent event) {
+    analysisReadinessByConfigScopeId.remove(event.getRemovedConfigurationScopeId());
+  }
+
+  @EventListener
+  public void onBindingConfigurationChanged(BindingConfigChangedEvent event) {
+    var configScopeId = event.getConfigScopeId();
+    analysisReadinessByConfigScopeId.remove(configScopeId);
+    if (!checkIfReadyForAnalysis(configScopeId)) {
+      client.didChangeAnalysisReadiness(new DidChangeAnalysisReadinessParams(Set.of(configScopeId), false));
```
henryju, line 380: Shouldn't the sending of the event managed at the same place? In the `saveAndNotifyReadyForAnalysis`?

backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisService.java
```diff
@@ -340,4 +355,81 @@ private static Predicate<? super SonarLintRuleDefinition> isIncludedByConfigurat
...
+
+  private void checkIfReadyForAnalysis(Set<String> configurationScopeIds) {
+    var readyConfigScopeIds = configurationScopeIds.stream()
+      .filter(this::isReadyForAnalysis)
+      .collect(toSet());
+    saveAndNotifyReadyForAnalysis(readyConfigScopeIds);
+  }
+
+  private boolean checkIfReadyForAnalysis(String configurationScopeId) {
+    if (isReadyForAnalysis(configurationScopeId)) {
```
henryju, line 402: This is not atomic if I am not mistaken

backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/storage/PluginsStorage.java
```diff
@@ -49,6 +49,15 @@ public PluginsStorage(Path connectionStorageRoot) {
     this.pluginReferencesFilePath = rootPath.resolve(PLUGIN_REFERENCES_PB);
   }
 
+  public boolean isValid() {
+    try {
+      rwLock.read(() -> ProtobufFileUtil.readFile(pluginReferencesFilePath, Sonarlint.PluginReferences.parser()));
```
henryju, line 54: Shouldn't you also check if all plugin JARs are available?

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/client/log/LogParams.java
```diff
@@ -45,6 +51,7 @@ public LogParams(LogLevel level, @Nullable String message, @Nullable String conf
     this.threadName = threadName;
     this.loggerName = loggerName;
     this.stackTrace = stackTrace;
+    this.loggedAt = Instant.now();
```
henryju, line 54: If you want the timestamp, I suggest to not collect it here, but make it a parameter, since logback can already provide it. See in `SonarLintRpcClientLogbackAppender::append`, you can get the log timestamp from `eventObject.getTimeStamp()`

henryju (changes requested):
This looks good overall, but I have doubt about the thread safety of the AnalysisService. I also find the `DidChangeAnalysisReadinessParams` strange. Why not having a boolean for each configScope, so that you can transmit both true -> false and false -> true in the same event? Or maybe simply have 2 sets in the Dto: configScopesChangedToReady and configScopesChangedToNotReady or something like that.

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/client/log/LogParams.java
damien-urruty-sonarsource: Makes sense, will change that

backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/storage/PluginsStorage.java
damien-urruty-sonarsource: There is a log later when starting the analysis if one jar is missing/corrupted. I think it's better to proceed with the analysis at this point

backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisService.java
damien-urruty-sonarsource: These methods are invoked from event listeners, which are called synchronously and serially, so they are not accessed concurrently. We make this assumption in other places too. I agree later if we change the events delivery it might break, but I don't want to over-complexify the code. Do you have a suggestion?

backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisService.java
damien-urruty-sonarsource: It could, but the method would be a bit more complex

damien-urruty-sonarsource:
> Why not having a boolean for each configScope, so that you can transmit both true -> false and false -> true in the same event

In practice, it cannot happen that configuration scopes readiness change in opposite directions at the same time. Hence the simpler API. I suggest we discuss about some of your comments tomorrow

henryju (approved):

PR #856, opened 2024-02-02
Use JDK 17 to build since Next requires it

henryju:


nquinquenel (approved):

henryju:
In the end it seems to not be required (yet), so closing.

PR #859, opened 2024-02-05
Fix failing test consecutive to analyzer metadata update

damien-urruty-sonarsource:
The rule description of javasecurity:S2083 was updated in SQ 10.4


henryju (approved):
You can probably cherry-pick it on branch-9.x

PR #860, opened 2024-02-06
Hardening SLOOP

henryju:


backend/cli/src/main/assembly/bin/sonarlint-backend.bat
```diff
@@ -58,7 +31,10 @@ if defined JAVA_HOME_ARG (
 set CLASSPATH=%APP_HOME%\lib\*;
 
 @rem Execute SLCORE
-"%JAVA_EXE%" %DEFAULT_JVM_OPTS% -classpath "%CLASSPATH%" org.sonarsource.sonarlint.core.backend.cli.SonarLintServerCli %args%
+"%JAVA_EXE%" ^
+   %SONARLINT_JVM_OPTS% ^
```
vojtech-suchy-sonarsource, line 35: I tried to test passing in `SONARLINT_JVM_OPTS ` like this ` cmd.exe /c 'C:\work\sonarlint-core\backend\cli\src\main\assembly\bin\sonarlint-backend.bat -DSONARLINT_JVM_OPTS=hello'` but that kept the `DSONARLINT_JVM_OPTS` unset. What is the correct way to set it?

henryju: I don't think it is the correct way, as it is supposed to be an environment variable, but we made a test as well with @thahnen and it was not working either. Not sure what is missing...

henryju: I think the way to test it is more like: `export SONARLINT_JVM_OPTS=-Xmx1G && ./sonarlint-backend` or something similar in Windows

vojtech-suchy-sonarsource: Nice. It worked with `set SONARLINT_JVM_OPTS=-Xmx1G && C:\work\sonarlint-core\backend\cli\src\main\assembly\bin\sonarlint-backend.bat`

vojtech-suchy-sonarsource (approved):

PR #861, opened 2024-02-06
fix/vs/SLCORE-691-fix-AnalyzerSettingsUpdateSummary

vojtech-suchy-sonarsource:
In [LocalStorageSynchronizer](https://github.com/SonarSource/sonarlint-core/blob/master/backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/LocalStorageSynchronizer.java#L75) we return null as a AnalyzerSettingsUpdateSummary if the local analyzer configuration (represented by the analyzer_config.pb file) can not be read. 

Might be better to take the configuration downloaded from a server and consider it as a AnalyzerSettingsUpdateSummary as the local setting likely doesn’t exist


backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/LocalStorageSynchronizer.java
```diff
@@ -79,7 +78,7 @@ public AnalyzerSettingsUpdateSummary synchronizeAnalyzerConfig(ServerApi serverA
       var originalAnalyzerConfiguration = storage.project(projectKey).analyzerConfiguration().read();
       configUpdateSummary = diffAnalyzerConfiguration(originalAnalyzerConfiguration, updatedAnalyzerConfiguration);
     } catch (StorageException e) {
-      configUpdateSummary = null;
+      configUpdateSummary = new AnalyzerSettingsUpdateSummary(updatedAnalyzerConfiguration.getSettings().getAll());
```
damien-urruty-sonarsource, line 81: As this `synchronizeAnalyzerConfig` method always returns non-null, you can also remove the null check in `SynchronizationService`.

PR #863, opened 2024-02-06
Update parent pom and license headers

henryju:
The update of license headers have been lost in one of the rebase of the sloop branch. I took the opportunity to fix some other headers.


damien-urruty-sonarsource (approved):
I did not check every single file, but LGTM

PR #864, opened 2024-02-07
Hardening for 10.0

damien-urruty-sonarsource:


vojtech-suchy-sonarsource (approved):
LGTM 👍 

I've noticed we use hard coded strings for request headers (such as [User-Agent](https://github.com/SonarSource/sonarlint-core/pull/864/files#diff-adbebdd2b5940ee06f48effa07a62fb4bc449c635f6c5e5b2de4582c36972213R58) or `Authorization`). There is an option to use eg. [google's HTTPHeaders class](https://github.com/google/guava/blob/master/guava/src/com/google/common/net/HttpHeaders.java#L190)  to get these constants (Or we could make one ourselves)

damien-urruty-sonarsource:
@vojtech-suchy-sonarsource I extracted constants, I prefer not to rely on 3rd party libs for such trivial things (even if we already have it today)

vojtech-suchy-sonarsource (approved):
Sweet 👍

PR #866, opened 2024-02-09
SLCORE-693 send notification about release to the new squad channel

vojtech-suchy-sonarsource:


.github/workflows/release.yml
```diff
@@ -14,4 +14,4 @@ jobs:
     with:
       publishToBinaries: false
       mavenCentralSync: true
-      slackChannel: team-sonarlint-java
+      slackChannel: squad-sonar-solution
```
damien-urruty-sonarsource, line 17: Shouldn't it be `#squad-sonar-solution-bots`? In fact if it's just for the release, I'm fine having it in the discussion channel

damien-urruty-sonarsource (approved):

.github/workflows/release.yml
vojtech-suchy-sonarsource: I think `#squad-sonar-solution-bots` is mostly for squad members to be aware of JIRA, Github and Cirrus events and so we don't polite our official `#squad-sonar-solution` where we have even external members (PMs, UXs, people from other squads). That said, I think these external members might be interested in getting the release notification (which won't happen often) so I think the `#squad-sonar-solution` channel might be good place.

PR #868, opened 2024-02-13
SLCORE-689 Support SQ 9.9

serhat-yenican-sonarsource:


its/tests/src/test/java/its/SonarQubeDeveloperEditionTests.java
```diff
@@ -1056,9 +1030,6 @@ void stop() {
     }
 
     @Test
-    // SonarQube should support opening security hotspots
-    @OnlyOnSonarQube(from = "8.6")
-    @Disabled
```
serhat-yenican-sonarsource, line 1061: I am not sure why it was disabled. I tried to enable it, and now it passes

nquinquenel: It was suddenly making the pipeline fail a few days ago, my guess was that it was because the rule has been deprecated and it wasn't loaded anymore

serhat-yenican-sonarsource: Thanks for the heads up. It looks like it's working again

PR #869, opened 2024-02-14
SLCORE-695 Prefer suggesting parent scope to bind even if children are eligible

damien-urruty-sonarsource:


backend/core/src/main/java/org/sonarsource/sonarlint/core/BindingCandidatesFinder.java
```diff
@@ -55,16 +56,26 @@ public Set<String> findConfigScopesToBind(String connectionId, String projectKey
...
+    // we prefer to bind at the broadest possible scope
+    return filterOutLeafCandidates(goodConfigScopeCandidates);
+  }
+
+  private static Set<String> filterOutLeafCandidates(Set<ConfigurationScope> candidates) {
+    var candidateIds = candidates.stream().map(ConfigurationScope::getId).collect(Collectors.toSet());
+    return candidates.stream().filter(scope -> {
+      var parentId = scope.getParentId();
+      return parentId == null || !candidateIds.contains(parentId);
+    }).map(ConfigurationScope::getId).collect(Collectors.toSet());
```
serhat-yenican-sonarsource, line 78: It's a minor thing but it would be nicer if we could avoid adding a leaf candidate to the collection in the first place rather than filtering them out later.

serhat-yenican-sonarsource: But then we should be careful whether the child comes first which increases the check method complexity.

In this case, your version is preferable, yes

serhat-yenican-sonarsource (approved):
Looks good!

backend/core/src/main/java/org/sonarsource/sonarlint/core/BindingCandidatesFinder.java
damien-urruty-sonarsource: Yes, I think it's simpler to remove it at the end, and performance-wise I don't think it makes a big difference

PR #873, opened 2024-02-23
SLCORE-702 Register for server-sent events from a background thread

damien-urruty-sonarsource:


vojtech-suchy-sonarsource (approved):
Neat :shipit: 
Good we are shutting down from SonarLintServerCli now 👍

PR #878, opened 2024-03-04
SLCORE-633 Avoid noisy error when storage is not ready for exclusions computation

damien-urruty-sonarsource:


kirill-knize-sonarsource (approved):
LGTM

PR #879, opened 2024-03-05
SLCORE-705: Slack notification on pipeline failure

thahnen:
## INFO: This will be merged AFTER the next release, just in case to not break anything.


damien-urruty-sonarsource (commented):
Will a notification be sent also for non-main branches?

thahnen:
@damien-urruty-sonarsource the notification will be sent by all failing tasks on CI builds. So the same as it is one currently ^^

nquinquenel (approved):
🐐

PR #883, opened 2024-03-07
SLCORE-713 Close HTML tag in context rule description to simplify par…

kirill-knize-sonarsource:


georgii-borovinskikh-sonarsource (approved):
LGTM!

PR #885, opened 2024-03-08
SLCORE-714 Do not throw an exception when client file encoding is unknown

damien-urruty-sonarsource:


kirill-knize-sonarsource (approved):
LGTM!

PR #886, opened 2024-03-08
SLCORE-715 Error if rule set changing between the start of analysis and reporting issue

kirill-knize-sonarsource:


damien-urruty-sonarsource (approved):
LGTM, the tests are pretty complicated, but let's rework that as part of the next sprint

PR #887, opened 2024-03-11
NO-JIRA: Raise awareness of compatibility

thahnen:
See [SonarSource/sonar-plugin-api#155](https://github.com/SonarSource/sonar-plugin-api/pull/155)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,6 +49,7 @@
   </modules>
 
   <properties>
+    <!-- INFO: When changing the Plugin API version here, also update it in the Plugin API repo README! -->
     <sonar-plugin-api.version>10.1.0.809</sonar-plugin-api.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -49,6 +49,7 @@
   </modules>
 
   <properties>
+    <!-- INFO: When changing the Plugin API version here, also update it in the Plugin API repo README! -->
```
damien-urruty-sonarsource, line 52: I think the comment can help to remind us to change the version, but it cannot be changed right away, as the core needs to be released and used by a released version of the IDE before it can be actually changed in the README. Maybe a better place to put a reminder would be the IDE release process?

thahnen:
Closed due to the information was added to the release process of the IDEs instead of putting it on the VCS.

PR #888, opened 2024-03-11
Prepare next dev iteration

kirill-knize-sonarsource:


damien-urruty-sonarsource (approved):
You could have used `10.1-SNAPSHOT` instead of `10.1.0-SNAPSHOT`, but I think it works too

PR #889, opened 2024-03-12
SLCORE-727 Use feature flag rather than system property to enable/disable telemetry on startup

vojtech-suchy-sonarsource:


damien-urruty-sonarsource (approved):
Good job!

PR #890, opened 2024-03-13
SLCORE-734 Directly invoke java to launch the process

damien-urruty-sonarsource:


client/rpc-java-client/src/main/java/org/sonarsource/sonarlint/core/rpc/client/SloopLauncher.java
```diff
@@ -122,6 +92,27 @@ private Sloop execute(Path distPath, @Nullable Path jrePath) throws IOException
     return new Sloop(serverProxy, process);
   }
 
+  private List<String> createCommand(Path distPath, @Nullable Path customJreHomePath) {
+    var jreHomePath = customJreHomePath == null ? distPath.resolve("jre") : customJreHomePath;
```
vojtech-suchy-sonarsource, line 96: I think we need to account that on Windows the java executable  has `.exe` exension so ` distPath.resolve("jre.exe")`

damien-urruty-sonarsource: True, and the tests were failing because of that. Fixed it

vojtech-suchy-sonarsource (approved):
LGTM.

Please see the comment about the process working directory.

client/rpc-java-client/src/main/java/org/sonarsource/sonarlint/core/rpc/client/SloopLauncher.java
```diff
@@ -86,23 +86,14 @@ private boolean isWindows() {
...
     }
-    if (jrePath != null) {
-      logToClient(LogLevel.INFO, "Using JRE from " + jrePath, null);
-      commands.add("-j");
-      commands.add(jrePath.toString());
-    }
-
-    var processBuilder = processBuilderFactory.apply(commands);
+    var processBuilder = processBuilderFactory.apply(createCommand(distPath, jreJavaExePath));
     processBuilder.directory(binDirPath.toFile());
```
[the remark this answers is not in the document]
damien-urruty-sonarsource: I was wondering why it was done like this before, and I just switched to keep using the folder where the script/exe is located as the current working directory. Anyway I don't think functionally the location is important, the CORE does not try to access files that are relative to the current directory.

For your point regarding security policies that would block the execution, I don't know. I am not aware of any restrictions for launching processes from some folders. I hope I'm wrong :slightly_smiling_face:

PR #891, opened 2024-03-13
SLCORE-732 Avoid null prompt when requesting proxy credentials to client

nquinquenel:


kirill-knize-sonarsource:
According to `org.apache.hc.client5.http.auth.AuthScope` protocol field also can be null. Maybe worth adding it to our API as well.

PR #893, opened 2024-03-14
SLCORE-726 Add debug telemetry option

vojtech-suchy-sonarsource:


backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryHttpClient.java
```diff
@@ -124,10 +125,19 @@ private void sendDelete(TelemetryPayload payload) {
...
+    }
     try (var response = client.post(endpoint, HttpClient.JSON_CONTENT_TYPE, payload.toJson())) {
       if (!response.isSuccessful() && InternalDebug.isEnabled()) {
         LOG.error("Failed to upload telemetry data: {}", response.toString());
       }
     }
   }
+
+  @VisibleForTesting
+  boolean isDebugTelemetryEnabled(){
```
serhat-yenican-sonarsource, line 140: How about `isTelemetryLogEnabled` for the method name and `SONARLINT_TELEMETRY_LOG` for the env var name?

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/initialize/FeatureFlagsDto.java
```diff
@@ -34,10 +34,11 @@ public class FeatureFlagsDto {
   private final boolean shouldManageServerSentEvents;
   private final boolean enableDataflowBugDetection;
   private final boolean shouldManageFullSynchronization;
+  private final boolean enableTelemetry;
 
 
   public FeatureFlagsDto(boolean shouldManageSmartNotifications, boolean taintVulnerabilitiesEnabled, boolean shouldSynchronizeProjects, boolean shouldManageLocalServer,
-    boolean enableSecurityHotspots, boolean shouldManageServerSentEvents, boolean enableDataflowBugDetection, boolean shouldManageFullSynchronization) {
+    boolean enableSecurityHotspots, boolean shouldManageServerSentEvents, boolean enableDataflowBugDetection, boolean shouldManageFullSynchronization, boolean enableTelemetry) {
```
serhat-yenican-sonarsource, line 41: What do you think about creating a default constructor for an instance with default values? Using a builder would even be better, actually.
So we don't need to worry about the usage of IDEs.

backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryHttpClient.java
vojtech-suchy-sonarsource: Addressed as requested

serhat-yenican-sonarsource (approved):
Validated on Eclipse. Working as expected!
LGTM!

PR #897, opened 2024-03-15
SLCORE-720 add back issue_status_changed_count to telemetry

serhat-yenican-sonarsource:
Validated in IntelliJ


damien-urruty-sonarsource (approved):
LGTM! We should keep the IDE squad informed, as they will need to update the telemetry sample on the IDE side

PR #901, opened 2024-03-19
SLCORE-735 Introduce provider for connected mode suggestions

nquinquenel:


backend/core/src/test/java/org/sonarsource/sonarlint/core/BindingClueProviderTests.java
```diff
@@ -189,14 +190,144 @@ void ignore_scanner_file_invalid_content() {
     assertThat(logTester.logs(LogOutput.Level.ERROR)).contains("Unable to parse content of file 'file://path/to/sonar-project.properties'");
   }
 
+  @Test
```
damien-urruty-sonarsource, line 193: I suggest (pun intended) to write medium tests as much as possible, as per our guidelines. Unit tests should be used when it's too hard to test at the medium test level. Here I think some of the cases could be converted, some others are already covered by a medium test, and maybe some might stay as unit tests

backend/core/src/main/java/org/sonarsource/sonarlint/core/ConnectionSuggestionProvider.java
```diff
@@ -0,0 +1,138 @@
...
+  private final ConfigurationRepository configRepository;
+  private final ConnectionConfigurationRepository connectionRepository;
+  private final SonarLintRpcClient client;
+  private final BindingClueProvider bindingClueProvider;
+  private final ExecutorServiceShutdownWatchable<?> executorService;
+  private final BindingSuggestionProvider bindingSuggestionProvider;
+
+  @Inject
+  public ConnectionSuggestionProvider(ConfigurationRepository configRepository, ConnectionConfigurationRepository connectionRepository, SonarLintRpcClient client,
+                                   BindingClueProvider bindingClueProvider, BindingSuggestionProvider bindingSuggestionProvider) {
```
damien-urruty-sonarsource, line 60: The formatting looks strange here

backend/core/src/main/java/org/sonarsource/sonarlint/core/BindingClueProvider.java
```diff
@@ -141,38 +170,57 @@ private Set<String> matchConnections(BindingClue bindingClue, Set<String> eligib
...
   }
 
-  private static class ScannerProperties {
+  private static class ConnectionProperties {
     private final String projectKey;
     private final String organization;
     private final String serverUrl;
 
-    private ScannerProperties(@Nullable String projectKey, @Nullable String organization, @Nullable String serverUrl) {
+    private ConnectionProperties(@Nullable String projectKey, @Nullable String organization, @Nullable String serverUrl) {
```
damien-urruty-sonarsource, line 215: Not sure it's a "connection" properties, as there is a project key involved. Maybe `BindingProperties`?

backend/core/src/main/java/org/sonarsource/sonarlint/core/BindingClueProvider.java
```diff
@@ -141,38 +170,57 @@ private Set<String> matchConnections(BindingClue bindingClue, Set<String> eligib
...
-    var properties = new Properties();
+  private static ConnectionProperties  extractSonarLintConfiguration(ClientFile sonarLintConfigurationFile) {
     try {
-      properties.load(new StringReader(matchedFile.getContent()));
-    } catch (Exception e) {
-      LOG.error("Unable to parse content of file '{}'", matchedFile, e);
+      var configuration = new Gson().fromJson(sonarLintConfigurationFile.getContent(), JsonObject.class);
+      var projectKey = configuration.get("projectKey");
+      var organization = configuration.get("organization");
+      var serverUrl = configuration.get("serverUrl");
```
damien-urruty-sonarsource, line 178: Is this compatible with the VS file format?

backend/core/src/main/java/org/sonarsource/sonarlint/core/ConnectionSuggestionProvider.java
```diff
@@ -0,0 +1,138 @@
...
+      .collect(Collectors.toSet());
+
+    if (!listConfigScopeIds.isEmpty()) {
+      var cancelMonitor = new SonarLintCancelMonitor();
+      cancelMonitor.watchForShutdown(executorService);
+      executorService.submit(() -> suggestConnectionForGivenScopes(listConfigScopeIds, cancelMonitor));
+    }
+  }
+
+  private void suggestConnectionForGivenScopes(Set<String> listOfFilesPerConfigScopeIds, SonarLintCancelMonitor cancelMonitor) {
```
damien-urruty-sonarsource, line 84: The only occasion where we call this method is when the filesystem is updated. Shouldn't we have other triggers? For example, if a scope is added?

backend/core/src/main/java/org/sonarsource/sonarlint/core/ConnectionSuggestionProvider.java
```diff
@@ -0,0 +1,138 @@
...
+
+      var bindingClues = bindingClueProvider.collectBindingClues(configScopeId, cancelMonitor);
+      if (bindingClues.size() == 1) {
+        var bindingClue = bindingClues.get(0);
+        var projectKey = bindingClue.getSonarProjectKey();
+        if (projectKey != null) {
+          handleBindingClue(connectionSuggestionsByConfigScopeIds, bindingClue, configScopeId, projectKey);
+        }
+      } else if (bindingClues.size() > 1) {
+        LOG.debug("Multiple binding clues found, skipping the suggestion");
```
damien-urruty-sonarsource, line 102: I think we could try to better handle this case: we could suggest several connections. Or suggest one of them. I agree this needs to be adapted in the protocol and the UI

damien-urruty-sonarsource (approved):
Lots of small cosmetic comments, the rest LGTM

backend/core/src/main/java/org/sonarsource/sonarlint/core/BindingClueProvider.java
sophio-japharidze-sonarsource: The keys we agreed on are
`sonarCloudOrganization`, `projectKey`, `sonarQubeUri` to be compatible with VS

backend/core/src/main/java/org/sonarsource/sonarlint/core/fs/ClientFile.java
```diff
@@ -123,6 +126,11 @@ public void setClean() {
     this.clientProvidedContent = null;
   }
 
+  public boolean isSonarlintConfigurationFile() {
+    var sonarlintPath = getClientRelativePath().getParent();
+    return "json".equals(FileNameUtils.getExtension(getClientRelativePath())) && sonarlintPath != null && SONARLINT_FOLDER_NAME.equals(sonarlintPath.getFileName().toString());
```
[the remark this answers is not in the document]
sophio-japharidze-sonarsource: Yes, VS names the files by either `<folderName>.json` or `<solutionName>.json`, and Rider will also call .NET project config files with `<solutionName>.json`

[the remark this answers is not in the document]
nquinquenel: Yes, the backend does not care about the filename because of the specificities of C# projects - we store the module name in the name, whereas in other languages we'll store them as connectedMode.json (or something similar) to make it clearer in the VCS.

Later, the files are parsed, and if the parsing does not work we skip the file

sophio-japharidze-sonarsource (commented):
Looks good, same comments as Damien.

Just one thing that I think needs change - for [telemetry](https://sonarsource.atlassian.net/browse/SLCORE-740) we want to differentiate between "imported bindings" (based on .sonarlint files) vs "automatic bindings" (currently in place based on scanner files), so I think we will need either a separate rpc notification for binding import, or a changed signature in `BindingSuggestionDto` like `isImported` or similar.

backend/core/src/main/java/org/sonarsource/sonarlint/core/BindingClueProvider.java
```diff
@@ -49,6 +53,7 @@
 public class BindingClueProvider {
 
   private static final SonarLintLogger LOG = SonarLintLogger.get();
+  private static final String CLUES = "CLUES";
```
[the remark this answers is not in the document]
nquinquenel: Oops, nope

backend/core/src/main/java/org/sonarsource/sonarlint/core/BindingClueProvider.java
nquinquenel: Thanks, I'll fix this

backend/core/src/main/java/org/sonarsource/sonarlint/core/BindingClueProvider.java
```diff
@@ -141,38 +170,57 @@ private Set<String> matchConnections(BindingClue bindingClue, Set<String> eligib
...
-    } catch (Exception e) {
-      LOG.error("Unable to parse content of file '{}'", matchedFile, e);
+      var configuration = new Gson().fromJson(sonarLintConfigurationFile.getContent(), JsonObject.class);
+      var projectKey = configuration.get("projectKey");
+      var organization = configuration.get("organization");
+      var serverUrl = configuration.get("serverUrl");
+      return new ConnectionProperties(projectKey != null ? projectKey.getAsString() : null,
+        organization != null ? organization.getAsString() : null,
+        serverUrl != null ? serverUrl.getAsString() : null);
+    } catch (JsonSyntaxException e) {
```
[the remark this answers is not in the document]
nquinquenel: I wouldn't expect to log something here, since we fetch all `.json` files under `.sonarlint`, maybe in the future we'll have other files, and here the parsing is used to differentiate whether it's a SonarLint file for sharing connected mode or not. It could be possible and normal to fail here.

backend/core/src/main/java/org/sonarsource/sonarlint/core/ConnectionSuggestionProvider.java
nquinquenel: After discussing with the sprint members, we'll send all the suggestions to the client, and it will be up to him to decide how to choose

backend/core/src/main/java/org/sonarsource/sonarlint/core/ConnectionSuggestionProvider.java
nquinquenel: My code style was wrong for method parameter multilines

PR #902, opened 2024-03-20
SLCORE-686 Don't load OS-specific SSL store on Windows

vojtech-suchy-sonarsource:


backend/http/src/main/java/org/sonarsource/sonarlint/core/http/HttpClientProvider.java
```diff
@@ -105,6 +108,10 @@ private static SSLContext configureSsl(SslConfig sslConfig, @Nullable Predicate<
     return sslFactoryBuilder.build().getSslContext();
   }
 
+  private static boolean isNotWindows() {
+    return System.getProperty("os.name") != null && !System.getProperty("os.name").startsWith("Windows");
```
henryju, line 112: Maybe adding a dependency on `commons-lang3`  and using SystemUtils would be safer/more maintainable.

vojtech-suchy-sonarsource: Yes, I was pondering that as well

damien-urruty-sonarsource (approved):
LGTM

PR #904, opened 2024-03-20
SLCORE-749 Add a cache mechanism for telemetry local storage

serhat-yenican-sonarsource:


serhat-yenican-sonarsource:
> The code looks good!
> 
> I wonder if we should add some tests which will check that the cache gets invalidated and refreshed if the storage file is modified under its hands (ie by external_user/other_process in real life)?

Thanks for the review @vojtech-suchy-sonarsource, that's what I'm exactly working on now.

PR #905, opened 2024-03-20
SLCORE-751 Use Spring to create and inject Telemetry singletons

serhat-yenican-sonarsource:


backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryHttpClient.java
```diff
@@ -39,10 +43,10 @@
 import org.sonarsource.sonarlint.core.telemetry.payload.cayc.CleanAsYouCodePayload;
 import org.sonarsource.sonarlint.core.telemetry.payload.cayc.NewCodeFocusPayload;
 
+@Named
+@Singleton
 public class TelemetryHttpClient {
```
vojtech-suchy-sonarsource, line 48: If we start using `SystemUtils` in `TelemetryHttpClient` (as suggested in another comment) we also can use it to detect OS in the `createPayload()` method to replace ` var os = System.getProperty("os.name");`

backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryManager.java
```diff
@@ -40,8 +45,22 @@ public class TelemetryManager {
...
-    uploadLazily(telemetryLiveAttributes);
+    storageManager.tryUpdateAtomically(localStorage -> {
+      localStorage.setEnabled(true);
+      if (isGracePeriodElapsed(localStorage.lastUploadTime())) {
+        uploadAndClearTelemetry(telemetryLiveAttributes, localStorage);
+      }
+    });
+  }
+
+  private static boolean isGracePeriodElapsed(LocalDateTime lastUploadTime) {
```
vojtech-suchy-sonarsource, line 56: I really like the `isGracePeriodElapsed` method name as it actually reflects what it's doing as opposed to the `dayChanged`. 
I know it's a years-old thing not introduced in this PR but I wonder whether to rename the `dayChanged` to `isGracePeriodElapsedAndDayChanged` or similar and use directly that,

vojtech-suchy-sonarsource (commented):
Very useful changes which make the code more readable and maintainable 👍 

Please see the minor comments

backend/core/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetrySpringConfig.java
```diff
@@ -0,0 +1,57 @@
...
+    return userHome.resolve("telemetry").resolve(productKey).resolve("usage");
+  }
+
+  @Bean(name = "telemetryEndpoint")
+  String provideTelemetryEndpoint() {
+    return System.getProperty(PROPERTY_TELEMETRY_ENDPOINT, TELEMETRY_ENDPOINT);
+  }
+
+  @Bean
+  TelemetryServerConstantAttributes provideServerConstantAttribute() {
```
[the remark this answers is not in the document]
serhat-yenican-sonarsource: I thought the same at first, but then I noticed some [tests](https://github.com/SonarSource/sonarlint-core/blob/21e0c3a107ba31854c02b44384d22a74222c093b/backend/telemetry/src/test/java/org/sonarsource/sonarlint/core/telemetry/TelemetryHttpClientTests.java#L68) checking injected mock values. Do you see any easy way to get rid of it?

backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryHttpClient.java
serhat-yenican-sonarsource: After your comment, I noticed that `os` and `platform` values are the same in the telemetry payload. We can remove one of both. I'll create a ticket for it.

serhat-yenican-sonarsource: Good point!

backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryManager.java
serhat-yenican-sonarsource: I renamed both the `dayChanged` utility method and the private method to what you suggested. I kept the instance method in order to use one fewer parameter.

vojtech-suchy-sonarsource (approved):
🚀

PR #907, opened 2024-03-21
SLCORE-745 give clients way tor request shared connected mode file contents

sophio-japharidze-sonarsource:


rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/connection/auth/UserTokenRpcService.java
```diff
@@ -17,7 +17,7 @@
  * along with this program; if not, write to the Free Software Foundation,
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
-package org.sonarsource.sonarlint.core.rpc.protocol.backend.usertoken;
+package org.sonarsource.sonarlint.core.rpc.protocol.backend.connection.auth;
```
damien-urruty-sonarsource, line 20: This will lead to a breaking API change. Not a big deal, but as I suggested in a message earlier, I think we should document those API changes somewhere.

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/connection/config/SharedConnectedModeSettingsRpcService.java
```diff
@@ -0,0 +1,43 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.sonarlint.core.rpc.protocol.backend.connection.config;
+
+import java.util.concurrent.CompletableFuture;
+import org.eclipse.lsp4j.jsonrpc.services.JsonRequest;
+
+public interface SharedConnectedModeSettingsRpcService {
```
damien-urruty-sonarsource, line 25: I was surprised to see a new service, I think this new method could fit in the `BindingRpcService` class. But no strong objection

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/connection/config/SharedConnectedModeSettingsRpcService.java
```diff
@@ -0,0 +1,43 @@
...
+   }</li>
+   OR
+   <li>   {
+   "sonarQubeUri": "$serverUrl",
+   "projectKey": "$projectKey"
+   }</li>
+
+   */
+  @JsonRequest
+  CompletableFuture<GetSharedConnectedModeConfigFileResponse> getSharedConnectedModeConfigFileContents(GetSharedConnectedModeConfigFileParams params);
```
damien-urruty-sonarsource, line 41: You could maybe document when this method might fail

backend/core/src/main/java/org/sonarsource/sonarlint/core/SharedConnectedModeSettingsProvider.java
```diff
@@ -0,0 +1,72 @@
...
+      var projectKey = bindingConfiguration.getSonarProjectKey();
+      var connectionId = bindingConfiguration.getConnectionId();
+
+      var connection =  Objects.requireNonNull(connectionRepository.getConnectionById(Objects.requireNonNull(connectionId)));
+      if (connection.getEndpointParams().isSonarCloud()) {
+        AtomicReference<String> fileContents = new AtomicReference<>();
+        connection.getEndpointParams().getOrganization().ifPresent(organization ->
+          fileContents.set(String.format(SONARCLOUD_CONNECTED_MODE_CONFIG, organization, projectKey))
+        );
+        return fileContents.get();
```
damien-urruty-sonarsource, line 63: This will return `null` if there is no organization? I think it's not documented in the API, but is it what we want? 

`connectionRepository.getConnectionById` returns an `AbstractConnectionConfiguration`, and you can use `instanceof` or `getKind` to cast it to the concrete type, where organization won't be nullable in the case of SC. So it removes this edge case

backend/core/src/main/java/org/sonarsource/sonarlint/core/SharedConnectedModeSettingsProvider.java
```diff
@@ -0,0 +1,72 @@
...
+      } else {
+        return String.format(SONARQUBE_CONNECTED_MODE_CONFIG, connection.getEndpointParams().getBaseUrl(), projectKey);
+      }
+    } else {
+      LOG.warn("Request for generating shared Connected Mode configuration file content failed; Binding not yet available for '{}'", configScopeId);
+      throw new SonarLintException("Binding not found; Cannot generate shared Connected Mode file contents");
```
damien-urruty-sonarsource, line 69: You could maybe add the configuration scope ID to the message, to help identify the source better

backend/core/src/test/java/org/sonarsource/sonarlint/core/SharedConnectedModeSettingsProviderTest.java
```diff
@@ -0,0 +1,110 @@
...
+import org.sonarsource.sonarlint.core.repository.connection.ConnectionConfigurationRepository;
+import org.sonarsource.sonarlint.core.repository.connection.SonarCloudConnectionConfiguration;
+import org.sonarsource.sonarlint.core.repository.connection.SonarQubeConnectionConfiguration;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+class SharedConnectedModeSettingsProviderTest {
```
damien-urruty-sonarsource, line 45: Are those unit tests covering a case that cannot be covered with a medium test? If not, I suggest to prefer medium tests, as per our guidelines.

damien-urruty-sonarsource (commented):
Looks mostly good, I left some ideas for improvement, the most important one about the elimination of an edge case

damien-urruty-sonarsource (approved):
LGTM

PR #908, opened 2024-03-21
SLCORE-718: Add support for JCL

thahnen:


nquinquenel (commented):
LGTM, I'll let someone from Sonar Solution squad approve

PR #909, opened 2024-03-21
Fix flaky test in telemetry

serhat-yenican-sonarsource:


backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryLocalStorageManager.java
```diff
@@ -101,16 +101,9 @@ private void updateLastModified() {
   }
 
   private TelemetryLocalStorage read() throws IOException {
-    var bytes = Files.readAllBytes(path);
-    if (bytes.length == 0) {
-      return new TelemetryLocalStorage();
+    try (var fileChannel = FileChannel.open(path, StandardOpenOption.READ)) {
+      return read(fileChannel);
```
damien-urruty-sonarsource, line 105: Jus tout of curiosity, could you share a bit more why the implementation was changed?

serhat-yenican-sonarsource: Sure! We designed the telemetry cache based on the local file modification time for cache invalidation. This flaky test result showed that we depend on OS platforms to get the modification time and at least Windows doesn't always provide this information correctly. Even though it is a rare case that the file is modified by the user or by another IDE (a different flavor of the same IDE), it is still a risk to update the file on top of out-dated cache data. To prevent this situation, I changed to code to always update the data in the file itself which was already the previous behavior before introducing the cache.
So, after this change, we only benefit from the cache while reading not updating.

damien-urruty-sonarsource (approved):

PR #910, opened 2024-03-26
SLCORE-753 Skip project synchronization when project key was recently synchronized

damien-urruty-sonarsource:


backend/core/src/main/java/org/sonarsource/sonarlint/core/sync/SynchronizationService.java
```diff
@@ -230,57 +252,74 @@ public void onConnectionCredentialsChanged(ConnectionCredentialsChangedEvent eve
...
+      .map(lastSync -> lastSync.isBefore(Instant.now().minus(syncPeriod, ChronoUnit.SECONDS)))
+      .orElse(true);
+    if (!result) {
+      LOG.debug("Skipping synchronization of binding '{}' because it was synchronized recently", binding);
+    }
+    return result;
+  }
+
   private boolean shouldSynchronizeScope(BoundScope configScope) {
     var syncPeriod = Long.parseLong(System.getProperty("sonarlint.internal.synchronization.scope.period", "300"));
```
serhat-yenican-sonarsource, line 321: `syncPeriod` can be defined as a constant. Actually, all System properties can be defined as constants or fields to be set in post construct method.

backend/core/src/main/java/org/sonarsource/sonarlint/core/sync/SynchronizationService.java
```diff
@@ -230,57 +252,74 @@ public void onConnectionCredentialsChanged(ConnectionCredentialsChangedEvent eve
...
     var cancelMonitor = new SonarLintCancelMonitor();
     cancelMonitor.watchForShutdown(scheduledSynchronizer);
-    scheduledSynchronizer.submit(() ->
-      serverApiProvider.getServerApi(connectionId)
-        .ifPresent(serverApi -> synchronizeConnectionAndProjectsIfNeededSync(connectionId, serverApi, boundScopes, cancelMonitor)));
+    scheduledSynchronizer.submit(() -> serverApiProvider.getServerApi(connectionId)
+      .ifPresent(serverApi -> synchronizeConnectionAndProjectsIfNeededSync(connectionId, serverApi, boundScopes, cancelMonitor)));
   }
 
   private void synchronizeConnectionAndProjectsIfNeededSync(String connectionId, ServerApi serverApi, Collection<BoundScope> boundScopes, SonarLintCancelMonitor cancelMonitor) {
```
serhat-yenican-sonarsource, line 269: `synchronizeConnectionAndProjectsIfNeededSync` method seems a little bit too complex to me. It would be nice if you could extract some parts into new method(s).

serhat-yenican-sonarsource (approved):
Nice improvement!
I added two minor comments. It is up to you to implement them now or later. 
LGTM!

backend/core/src/main/java/org/sonarsource/sonarlint/core/sync/SynchronizationService.java
damien-urruty-sonarsource: I agree. I created this [ticket](https://sonarsource.atlassian.net/browse/SLCORE-761) to share what we discussed in the past about this synchronization process, so I think we can tackle the complexity as part of this ticket

backend/core/src/main/java/org/sonarsource/sonarlint/core/sync/SynchronizationService.java
damien-urruty-sonarsource: The properties are set from tests to influence the period, so I think they have to be read each time. Maybe we could come up with a better approach, but I will keep it for later as it's not directly related to this PR

PR #911, opened 2024-03-27
SLCORE-757 Expose API to get a SQ/SC projectName based on a projectKey

vojtech-suchy-sonarsource:


backend/core/src/main/java/org/sonarsource/sonarlint/core/ConnectionService.java
```diff
@@ -193,4 +195,10 @@ public List<SonarProjectDto> getAllProjects(Either<TransientSonarQubeConnectionD
       .collect(Collectors.toList());
   }
 
+  public String getProjectNameByKey(Either<TransientSonarQubeConnectionDto, TransientSonarCloudConnectionDto> transientConnection,
+    String projectKey, SonarLintCancelMonitor cancelMonitor) {
+    var serverApi = serverApiProvider.getForTransientConnection(transientConnection);
+    return serverApi.component().getProject(projectKey, cancelMonitor).map(ServerProject::getName).orElse("");
```
damien-urruty-sonarsource, line 201: Maybe returning null or an optional would better reflect that the project cannot be retrieved? If it was deleted for example.

medium-tests/src/test/java/mediumtest/fixtures/ServerFixture.java
```diff
@@ -1014,6 +1015,18 @@ private void registerComponentsSearchApiResponses() {
...
     }
 
+    private void registerComponentsShowApiResponses() {
+      projectsByProjectKey.forEach((projectKey, project) -> {
+        var url = "/api/components/show.protobuf?component=" + projectKey;
+        var projectComponent = projectsByProjectKey.entrySet().stream().filter(e -> e.getKey().equals(projectKey)).map(entry -> Components.Component.newBuilder().setKey(entry.getKey()).setName(entry.getValue().name).build()).findFirst().get();
+        mockServer.stubFor(get(url)
+          .willReturn(aResponse().withResponseBody(protobufBody(Components.ShowWsResponse.newBuilder()
+            .setComponent(projectComponent)
+            .setPaging(Common.Paging.newBuilder().setTotal(project.relativeFilePaths.size()).build())
```
damien-urruty-sonarsource, line 1025: Is this intentional? I think the response is not related to files

damien-urruty-sonarsource (approved):
LGTM, the only thing is that before we were doing a single request to retrieve names for N project keys, now we will do N requests. Not sure it's a big deal. If it's a problem, at some point we could also fetch the name during the project synchronization and cache it locally.

medium-tests/src/test/java/mediumtest/fixtures/ServerFixture.java
vojtech-suchy-sonarsource: not intentional, fixed in [29c62f5](https://github.com/SonarSource/sonarlint-core/pull/911/commits/29c62f5cc34d26ba05bb75249c0a79a05ed3d03b)

vojtech-suchy-sonarsource:
> LGTM, the only thing is that before we were doing a single request to retrieve names for N project keys, now we will do N requests. Not sure it's a big deal. If it's a problem, at some point we could also fetch the name during the project synchronization and cache it locally.

Now addressed in [81f8635](https://github.com/SonarSource/sonarlint-core/pull/911/commits/81f863508c87fc355fd1560dc9f438c5f6bcf8a8);

Also note the change to[ serializing nulls](https://github.com/SonarSource/sonarlint-core/blob/81f863508c87fc355fd1560dc9f438c5f6bcf8a8/rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/SonarLintLauncherBuilder.java#L66)

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/SonarLintLauncherBuilder.java
```diff
@@ -62,7 +62,8 @@ public GsonBuilder getDefaultGsonBuilder() {
           .registerTypeHierarchyAdapter(Instant.class, new InstantTypeAdapter())
           .registerTypeHierarchyAdapter(UUID.class, new UuidTypeAdapter())
           .registerTypeHierarchyAdapter(URI.class, new UriTypeAdapter())
-          .registerTypeHierarchyAdapter(Duration.class, new DurationTypeAdapter());
+          .registerTypeHierarchyAdapter(Duration.class, new DurationTypeAdapter())
+          .serializeNulls();
```
damien-urruty-sonarsource, line 66: I think I understand Kirill's point. If the key is not in the map, it means we have no name for a given project key, I don't see how having a key with a null value provides more details. I checked the VSCode [implementation](https://github.com/SonarSource/sonarlint-vscode/blob/master/src/connected/connections.ts#L164) and it should work even with "missing" keys in the map. But no big deal.

damien-urruty-sonarsource (approved):
LGTM

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/SonarLintLauncherBuilder.java
vojtech-suchy-sonarsource: My point of view is that if the client sends a list of keys, we send them the same amount of keys back. We expressed that we did some operations for all the keys and didn't forget anything.
I will merge it with the `serializeNulls()`, but I don't hold a strong opinion on that. @damien-urruty-sonarsource & @kirill-knize-sonarsource, feel free to revert it if you don't see the above as valid.

PR #912, opened 2024-03-27
SLCORE-755 introduce a cache for server paths by binding

serhat-yenican-sonarsource:


backend/core/src/main/java/org/sonarsource/sonarlint/core/file/ServerFilePathsProvider.java
```diff
@@ -0,0 +1,169 @@
...
+  private final Map<Binding, Path> cachedResponseFilePathByBinding = new HashMap<>();
+  private final Path cacheDirectoryPath;
+  private final Cache<Binding, List<Path>> temporaryInMemoryFilePathCacheByBinding;
+
+  public ServerFilePathsProvider(ServerApiProvider serverApiProvider, Path storageRoot) {
+    this.serverApiProvider = serverApiProvider;
+    this.cacheDirectoryPath = storageRoot.resolve("cache");
+    this.temporaryInMemoryFilePathCacheByBinding = CacheBuilder.newBuilder()
+      .expireAfterWrite(Duration.of(1, ChronoUnit.MINUTES))
+      .maximumSize(3)
```
serhat-yenican-sonarsource, line 69: These two configurations might be calibrated

serhat-yenican-sonarsource: I gave small values so as not to tire the memory too much. We can even think of smaller values.

damien-urruty-sonarsource (approved):
I was a bit surprised that we don't have medium tests impacted by this change, but the unit tests seem to cover most of the logic. LGTM

backend/core/src/main/java/org/sonarsource/sonarlint/core/file/ServerFilePathsProvider.java
```diff
@@ -0,0 +1,169 @@
...
+      return Files.readAllLines(responsePath).stream().map(Paths::get).collect(toList());
+    } catch (IOException e) {
+      LOG.debug("Error occurred while reading the file server path response file cache {}", responsePath);
+      return Collections.emptyList();
+    }
+  }
+
+  private Optional<List<Path>> getPathsFromInMemoryCache(Binding binding) {
+    return Optional.ofNullable(temporaryInMemoryFilePathCacheByBinding.getIfPresent(binding));
+  }
```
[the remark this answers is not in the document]
serhat-yenican-sonarsource: Make sense, thanks. I was using depth-first order. Now I changed it to breadth-first.

PR #913, opened 2024-03-27
SLCORE-759 Add a property to enable RPC logs

damien-urruty-sonarsource:


backend/rpc-impl/src/main/java/org/sonarsource/sonarlint/core/rpc/impl/SonarLintRpcServerImpl.java
```diff
@@ -105,12 +110,23 @@ public SonarLintRpcServerImpl(InputStream in, OutputStream out, ExecutorService
...
     LOG.info("SonarLint backend started, instance={}", this);
   }
 
+  private static PrintWriter getMessageTracer() {
+    if ("true".equals(System.getProperty("sonarlint.debug.rpc"))) {
+      try {
+        return new PrintWriter(Paths.get(System.getProperty("user.home")).resolve(".sonarlint").resolve("rpc_session.log").toFile(), StandardCharsets.UTF_8);
+      } catch (IOException e) {
+        System.err.println("Cannot write rpc debug logs file");
+        e.printStackTrace();
```
vojtech-suchy-sonarsource, line 124: Personally, I might prefer not to print the stack trace but it's ok.
Also, should we use the `e.printStackTrace(System.err)` for printing?

vojtech-suchy-sonarsource (approved):
LGTM.
 One comment about printing stack trace.

backend/rpc-impl/src/main/java/org/sonarsource/sonarlint/core/rpc/impl/SonarLintRpcServerImpl.java
damien-urruty-sonarsource: `e.printStackTrace()` already calls `e.printStackTrace(System.err)`.

I decided to print the stack trace else we would lose the cause. This flag is for internal debug purposes so I think it's fine

PR #916, opened 2024-03-28
SLCORE-760 Define explicit JVM heap size on SLOOP start

kirill-knize-sonarsource:


damien-urruty-sonarsource (commented):
I am concerned that setting the value to 2G might in fact lower the max heap size on some environments where the default is higher. On my Linux, if I run the following:
> java -client -XX:+PrintFlagsFinal 2>&1 | grep -i -E 'MaxHeapSize'

I get:
> size_t MaxHeapSize                              = 8340373504                                {product} {ergonomic}

Which is close to 8G if I'm not wrong. So with the Xmx option set to 2028M, we would be in fact lowering the max heap size on some machines.

My second concern is, if I understand correctly, starting the JVM will fail if the OS is not able to allocate the amount of memory requested.

I'm not an expert with those settings, so I can't say with confidence if we should move forward with this change

damien-urruty-sonarsource:
Closing as per the concern shared above about having a hardcoded size lower than the default

PR #917, opened 2024-03-28
Use embedded JS plugin for ITs

damien-urruty-sonarsource:
This makes the test faster as we don't need to sync the plugin


vojtech-suchy-sonarsource (approved):
LGTM.

Just to double check, do we need to update other integration tests as well to use the new `PluginLocator.getEmbeddedPluginsByKeyForTests(),` method?

damien-urruty-sonarsource:
@vojtech-suchy-sonarsource I already modified the ITs that were relevant to use this method

PR #919, opened 2024-03-28
Build with Java 17 (but keep targeting 11)

damien-urruty-sonarsource:
This fixes a problem running the Sonar analysis that now requires Java 17


jblievremont (approved):
LGTM 👍🏻 tested by manually building with Java 17 and checking that the generated classes have version 55 (Java 11).

PR #922, opened 2024-04-02
SLCORE-568 Propagate CCT attributes from server

vojtech-suchy-sonarsource:


backend/core/src/main/java/org/sonarsource/sonarlint/core/rules/RuleDetails.java
```diff
@@ -79,8 +79,8 @@ public static RuleDetails merging(ServerActiveRule activeRuleFromStorage, Server
         .collect(Collectors.groupingBy(DescriptionSection::getKey)),
       Optional.ofNullable(activeRuleFromStorage.getSeverity()).orElse(serverRule.getSeverity()),
       serverRule.getType(),
-      null, // TODO get clean code attribute from storage or server rule?
-      Map.of(), // TODO get impacts from storage?
+      serverRule.getCleanCodeAttribute(),
+      serverRule.getImpacts(),
```
serhat-yenican-sonarsource, line 83: It would be really nice if you could move the class fields and the constructor to the top.

serhat-yenican-sonarsource (approved):
Looks good!
Just two minor comments.

LGTM!

PR #923, opened 2024-04-03
SLCORE-738 Consider shared Connected Mode settings when responding to binding suggestions request

nquinquenel:


thahnen (approved):
LGTM!

backend/core/src/main/java/org/sonarsource/sonarlint/core/BindingClueProvider.java
```diff
@@ -107,7 +107,7 @@ public Set<String> getConnectionIds() {
   public List<BindingClue> collectBindingClues(String checkedConfigScopeId, SonarLintCancelMonitor cancelMonitor) {
     var sonarlintConfigurationFiles = clientFs.findSonarlintConfigurationFilesByScope(checkedConfigScopeId);
     if (!sonarlintConfigurationFiles.isEmpty()) {
-      var collectedClues = collectFromFiles(sonarlintConfigurationFiles, cancelMonitor);
+      var collectedClues = collectFromFiles(sonarlintConfigurationFiles, cancelMonitor, true);
```
vojtech-suchy-sonarsource, line 110: I have noticed that we are sending the `isFromSharedConfiguration` param 3 levels down into SonarCloudBindingClue/SonarQubeBindingClue constructor but it's not really needed for the computation of binding clues.
I wonder if it would make the code cleaner if we had a seter and marked them with the `isFromSharedConfiguration` here on the `collectedClues`. (So practically we are balancing having a setter on a DTO (not great) and having to modify method signature and pass parameters down few levels)

vojtech-suchy-sonarsource (approved):
LGTM; but see the comment about `isFromSharedConfiguration` param

backend/core/src/main/java/org/sonarsource/sonarlint/core/BindingClueProvider.java
nquinquenel: Good point. I'm not a fan of modifying a DTO object, so I went another solution which does not rely on passing a value in every method signature: the information about whether it's from a shared configuration also comes from the BindingProperties object, which we have access to

PR #924, opened 2024-04-03
DO NOT MERGE: Use LSP4J 0.15.0 instead of 0.21.0

thahnen:
TODO: Follow up with information


damien-urruty-sonarsource:
This won't be merged as it is too dangerous to downgrade. #933 will fix the root issue by shading & relocating lsp4j

PR #927, opened 2024-04-05
SLCORE-664 Protect from nulls in RuleSetChangedEvent

vojtech-suchy-sonarsource:
See https://sonarsource.atlassian.net/browse/SLCORE-664?focusedCommentId=596799

### DIFF
--- a/backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/push/parsing/RuleSetChangedEventParser.java
+++ b/backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/push/parsing/RuleSetChangedEventParser.java
@@ -48,7 +48,7 @@ public class RuleSetChangedEventParser implements EventParser<RuleSetChangedEven
         changedRule.key,
         changedRule.language,
         IssueSeverity.valueOf(changedRule.severity),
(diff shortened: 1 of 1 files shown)


backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/push/parsing/RuleSetChangedEventParser.java
```diff
@@ -48,7 +48,7 @@ public Optional<RuleSetChangedEvent> parse(String jsonData) {
         changedRule.key,
         changedRule.language,
         IssueSeverity.valueOf(changedRule.severity),
-        changedRule.params.stream().collect(Collectors.toMap(p -> p.key, p -> p.value)),
+        changedRule.params.stream().filter(p -> p.value != null).collect(Collectors.toMap(p -> p.key, p -> p.value)),
```
serhat-yenican-sonarsource, line 51: Normally, a map is supposed to allow null values. I had a quick search about it. As mentioned [here](https://stackoverflow.com/a/24634007), it is a known JDK bug. We can apply this suggested workaround if we want to allow the null values. If it is fine to filter them out, we can also keep the change which is much simpler.

serhat-yenican-sonarsource (approved):
Looks good!
We can also store the null values with a workaround. It's up to you to implement.

PR #928, opened 2024-04-08
SLCORE-740 Telemetry for sharing connected mode

nquinquenel:


backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/payload/ShareConnectedModePayload.java
```diff
@@ -0,0 +1,47 @@
...
+
+import com.google.gson.annotations.SerializedName;
+import javax.annotation.Nullable;
+
+public class ShareConnectedModePayload {
+
+  @SerializedName("manual_added_bindings_count")
+  public final Integer manualAddedBindingsCount;
+
+  @SerializedName("imported_added_bindings_count")
```
sophio-japharidze-sonarsource, line 30: minor note - I think `imported_bindings_count` & `auto_bindings_count` would be enough and also in line with what DPO [expects](https://docs.google.com/spreadsheets/d/1oIPd5dZrjWqtujEQ5hQIJ_fVe3heCUIzmNmEaoAC_W4/edit#gid=368366020)

sophio-japharidze-sonarsource (approved):
The changes look good to me. The only question is if all these fields should indeed be nested under `shared_connected_mode` payload? And if there is any preference for the exact field keys. I think both are questions for our DPO.

backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/payload/ShareConnectedModePayload.java
nquinquenel: DPO needs the total binding count, which to get we need to add the manual binding counts. I don't think we can skip it, WDYT?

sophio-japharidze-sonarsource: Yes yes for sure, I just meant for the naming `imported_added_bindings_count` -> `imported_bindings_count`

nquinquenel: oh I see 😄

vojtech-suchy-sonarsource (approved):
LGTM.

As discussed, I find it a bit odd that we stopped cleaning the telemetry data after the telemetry ping. Thank you for confirming it with DPOs. 

Also, a reminder that we will need to update the telemetry samples + [[Data Team](https://xtranet-sonarsource.atlassian.net/wiki/spaces/DATA?homepageId=2237989083)
[Data sources, catalogs and lineage](https://xtranet-sonarsource.atlassian.net/wiki/spaces/DATA/pages/2348187685/Data+sources+catalogs+and+lineage)
[Telemetry data catalog for SonarLint](https://xtranet-sonarsource.atlassian.net/wiki/spaces/DATA/pages/2347991072/Telemetry+data+catalog+for+SonarLint#) (unless DPO will do it)





Summarize



Share

Telemetry data catalog](https://xtranet-sonarsource.atlassian.net/wiki/spaces/DATA/pages/2347991072/Telemetry+data+catalog+for+SonarLint)

PR #929, opened 2024-04-08
Avoid sending multiple binding suggestions

nquinquenel:
- Since `ConnectionSuggestionProvider` was listening to the same events at the `BindingSuggestionProvider`, it is not useful to keep the listeners in `BindingSuggestionProvider` because anyway the connection provider will fallback to the binding provider if there is no connection to be suggested
- Avoid the `BindingSuggestionProvider` to propose a binding while the connection is being created on the client side
- Fix test error with trailing '/'


vojtech-suchy-sonarsource (approved):
📦

PR #933, opened 2024-04-09
SLCORE-663 Stop leaking LSP4J types in API and services

vojtech-suchy-sonarsource:


client/rpc-java-client/src/main/java/org/sonarsource/sonarlint/core/rpc/client/SonarLintCancelChecker.java
```diff
@@ -0,0 +1,54 @@
...
+package org.sonarsource.sonarlint.core.rpc.client;
+
+import java.util.concurrent.CancellationException;
+
+/*
+ * A class to use in place of {@link org.eclipse.lsp4j.jsonrpc.CancelChecker} to stop depending on lsp4j types in API
+ * and services.
+ * See SLCORE-663 for details.
+ */
+public class SonarLintCancelChecker implements org.eclipse.lsp4j.jsonrpc.CancelChecker {
```
damien-urruty-sonarsource, line 29: Do we need to inherit from lsp4j's `CancelChecker`? Is there a situation where we pass our own instance to a LSP4J API?

client/rpc-java-client/src/main/java/org/sonarsource/sonarlint/core/rpc/client/SonarLintRpcClientImpl.java
```diff
@@ -89,8 +88,8 @@ public SonarLintRpcClientImpl(SonarLintRpcClientDelegate delegate, ExecutorServi
     this.requestAndNotificationsSequentialExecutor = requestAndNotificationsSequentialExecutor;
   }
 
-  protected <R> CompletableFuture<R> requestAsync(Function<CancelChecker, R> code) {
-    CompletableFuture<CancelChecker> start = new CompletableFuture<>();
+  protected <R> CompletableFuture<R> requestAsync(Function<org.eclipse.lsp4j.jsonrpc.CancelChecker, R> code) {
+    CompletableFuture<org.eclipse.lsp4j.jsonrpc.CancelChecker> start = new CompletableFuture<>();
```
damien-urruty-sonarsource, line 92: Maybe we don't need to use the FQN here, as our own cancel checker is called differently. But if you think it makes things clearer, I'm fine keeping it

its/plugins/java-custom-rules/pom.xml
```diff
@@ -65,6 +65,7 @@
           <pluginKey>custom</pluginKey>
           <sonarLintSupported>true</sonarLintSupported>
           <requirePlugins>java:${sonarjava.version}</requirePlugins>
+          <requiredForLanguages>java</requiredForLanguages>
```
damien-urruty-sonarsource, line 68: You can remove this when you rebase onto master

medium-tests/src/test/java/mediumtest/hotspots/MatchWithServerHotspotsMediumTests.java
```diff
@@ -96,7 +96,7 @@ void it_should_track_local_only_hotspots() {
       .satisfies(result -> assertThat(result.getSecurityHotspotsByIdeRelativePath())
         .hasEntrySatisfying(Path.of("file/path"), hotspots -> {
           assertThat(hotspots).hasSize(1).allSatisfy(hotspot -> assertThat(hotspot.isRight()).isTrue());
-          assertThat(hotspots).usingRecursiveComparison().ignoringFields("right.id")
+          assertThat(hotspots).usingRecursiveComparison().ignoringFields("lsp4jEither.right.id")
```
damien-urruty-sonarsource, line 99: Isn't there a way to avoid referencing the lsp4j field in the suite?

medium-tests/src/test/java/mediumtest/issues/TrackWithServerIssuesMediumTests.java
```diff
@@ -107,7 +107,7 @@ void it_should_track_local_only_issues() {
       .satisfies(result -> assertThat(result.getIssuesByIdeRelativePath())
         .hasEntrySatisfying(Path.of("file/path"), issues -> {
           assertThat(issues).hasSize(1).allSatisfy(issue -> assertThat(issue.isRight()).isTrue());
-          assertThat(issues).usingRecursiveComparison().ignoringFields("right.id")
+          assertThat(issues).usingRecursiveComparison().ignoringFields("lsp4jEither.right.id")
```
damien-urruty-sonarsource, line 110: Same here

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/adapter/EitherTypeAdapter.java
```diff
@@ -0,0 +1,256 @@
...
+import javax.annotation.Nullable;
+import org.sonarsource.sonarlint.core.rpc.protocol.common.Either;
+import org.sonarsource.sonarlint.core.rpc.protocol.common.Either3;
+
+/*
+ * A class to use in place of {@link org.eclipse.lsp4j.jsonrpc.json.adapters.EitherTypeAdapter} to stop depending on lsp4j types in API
+ * and services.
+ * See SLCORE-663 for details.
+ */
+public class EitherTypeAdapter<L, R> extends TypeAdapter<Either<L, R>> {
```
damien-urruty-sonarsource, line 46: I presume this class is a copy/paste from the original Either, now that I see it, I think we could just wrap the original adapter instead of copying it here. We would benefit from bug fixes on their side if any, and it would simplify the code

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/adapter/JsonElementTypeAdapter.java
```diff
@@ -0,0 +1,77 @@
...
+import com.google.gson.stream.JsonWriter;
+import java.io.IOException;
+import javax.annotation.Nullable;
+
+/*
+ * A class to use in place of {@link org.eclipse.lsp4j.jsonrpc.json.adapters.JsonElementTypeAdapter} to stop depending on lsp4j types in API
+ * and services.
+ * See SLCORE-663 for details.
+ */
+public class JsonElementTypeAdapter extends TypeAdapter<Object> {
```
damien-urruty-sonarsource, line 37: Do we really need to clone this one?

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/adapter/Tuple.java
```diff
@@ -0,0 +1,101 @@
...
+package org.sonarsource.sonarlint.core.rpc.protocol.adapter;
+
+import javax.annotation.Nullable;
+
+/*
+ * A class to use in place of {@link org.eclipse.lsp4j.jsonrpc.messages.Tuple} to stop depending on lsp4j types in API
+ * and services.
+ * See SLCORE-663 for details.
+ */
+public interface Tuple {
```
damien-urruty-sonarsource, line 29: Do we also need to clone this one?

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/adapter/TypeUtils.java
```diff
@@ -0,0 +1,240 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import org.sonarsource.sonarlint.core.rpc.protocol.common.Either;
+
+/*
+ * A class to use in place of {@link org.eclipse.lsp4j.jsonrpc.json.adapters.TypeUtils} to stop depending on lsp4j types in API
+ * and services.
+ * See SLCORE-663 for details.
+ */
+public final class TypeUtils {
```
damien-urruty-sonarsource, line 40: Same

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/common/Either3.java
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.sonarlint.core.rpc.protocol.common;
+
+import java.util.function.Function;
+
+public class Either3<T1, T2, T3> extends Either<T1, Either<T2, T3>> {
```
damien-urruty-sonarsource, line 24: Do we need to clone this one too? Do we have any usage of Either3 in our APIs?

damien-urruty-sonarsource (commented):
Looks mostly good to me, but I think there is room for simplification

client/rpc-java-client/src/main/java/org/sonarsource/sonarlint/core/rpc/client/SonarLintCancelChecker.java
thahnen: Yes I think so, as it is used in the `SonarLintRpcClientDelegate` that is inherited on the IDEs.

PR #936, opened 2024-04-11
SLCORE-753 Skip project synchronization when branch was recently synchronized

damien-urruty-sonarsource:


backend/core/src/main/java/org/sonarsource/sonarlint/core/sync/SynchronizationService.java
```diff
@@ -173,15 +178,25 @@ private void synchronizeProjectsOfTheSameConnection(String connectionId, Map<Str
     }
     serverApiProvider.getServerApi(connectionId).ifPresent(serverApi -> {
       var subProgressGap = progressGap / boundScopeBySonarProject.size();
-      var subProgress = progress;
+      var subProgress = new AtomicDouble(progress);
```
serhat-yenican-sonarsource, line 181: I don't see any concurrent access to the `subProgress` variable. Is there any special reason to use Atomic var?

backend/core/src/main/java/org/sonarsource/sonarlint/core/sync/SynchronizationService.java
```diff
@@ -173,15 +178,25 @@ private void synchronizeProjectsOfTheSameConnection(String connectionId, Map<Str
...
+          .forEach((branchNameOpt, scopes) -> branchNameOpt.ifPresent(branchName -> {
+            var branchBinding = new BranchBinding(new Binding(connectionId, sonarProjectKey), branchName);
+            if (shouldSynchronizeBranch(branchBinding)) {
+              branchSynchronizationTimestampRepository.setLastSynchronizationTimestampToNow(branchBinding);
+              notifier.notify("Synchronizing project '" + sonarProjectKey + "'...", (int) Math.round(subProgress.get()));
+              issueSynchronizationService.syncServerIssuesForProject(serverApi, connectionId, sonarProjectKey, branchName, cancelMonitor);
+              taintSynchronizationService.synchronizeTaintVulnerabilities(serverApi, connectionId, sonarProjectKey, branchName, cancelMonitor);
+              hotspotSynchronizationService.syncServerHotspotsForProject(serverApi, connectionId, sonarProjectKey, branchName, cancelMonitor);
+              synchronizedConfScopeIds.addAll(entry.getValue().stream().map(BoundScope::getConfigScopeId).collect(toSet()));
+              subProgress.getAndAccumulate(subProgressGap, Double::sum);
```
serhat-yenican-sonarsource, line 197: I think the accumulation of subProgress should be right under the for loop that iterates the `boundScopeBySonarProject` collection, and `subProgress` was divided into its size.

serhat-yenican-sonarsource (commented):
Looks good!
Just two minor comments

backend/core/src/main/java/org/sonarsource/sonarlint/core/sync/SynchronizationService.java
damien-urruty-sonarsource: The variable is used in a lambda, so I could not modify it directly and was getting a compilation error:

> Variable used in lambda expression should be final or effectively final

The other technique I encountered would be to use an array of 1 element, and to modify it everytime

If you see a better approach, let me know

serhat-yenican-sonarsource: Ahh, lambda, true. I didn't think that. I think your approach is quite reasonable.

backend/core/src/main/java/org/sonarsource/sonarlint/core/sync/SynchronizationService.java
damien-urruty-sonarsource: I think you're right. It was already wrong before, I will fix this

backend/core/src/main/java/org/sonarsource/sonarlint/core/sync/SynchronizationService.java
damien-urruty-sonarsource: I ended up extracting the lambda to a separate method

serhat-yenican-sonarsource (approved):
LGTM!

PR #937, opened 2024-04-15
Remove useless usage of CompletableFuture for suggesting connection

nquinquenel:
Since the `suggestConnection` only shows a notification, it's purely asynchronous and we cannot wait.

I'd suggest using `didUpdateBinding` and modifying the `bindingSuggestionDisabled` field on the client side before and after processing the binding.


thahnen (approved):
LGTM!

PR #938, opened 2024-04-15
Fix IT failing since SQ 10.5 release

damien-urruty-sonarsource:
Declare the <requiredForLanguages> tag so that the custom Java plugin used in tests is considered optional. This prevents the plugin from being loaded at the beginning of scanner execution, because the SonarJava dependency is not loaded either


vojtech-suchy-sonarsource (approved):
LGTM 🚢

PR #940, opened 2024-04-16
SLCORE-663 Wrap lsp4j Either in sonar packaging

vojtech-suchy-sonarsource:


backend/core/src/main/java/org/sonarsource/sonarlint/core/ServerApiProvider.java
```diff
@@ -95,8 +95,8 @@ public ServerApi getForTransientConnection(Either<TransientSonarQubeConnectionDt
     var endpointParams = transientConnection.map(
       sq -> new EndpointParams(sq.getServerUrl(), false, null),
       sc -> new EndpointParams(sonarCloudUri.toString(), true, sc.getOrganization()));
-    var httpClient = getClientFor(transientConnection
-      .map(TransientSonarQubeConnectionDto::getCredentials, TransientSonarCloudConnectionDto::getCredentials));
+    var httpClient = getClientFor((transientConnection
+      .map(TransientSonarQubeConnectionDto::getCredentials, TransientSonarCloudConnectionDto::getCredentials)));
```
damien-urruty-sonarsource, line 99: Unnecessary parenthesis?

medium-tests/src/test/java/mediumtest/hotspots/MatchWithServerHotspotsMediumTests.java
```diff
@@ -96,7 +96,7 @@ void it_should_track_local_only_hotspots() {
       .satisfies(result -> assertThat(result.getSecurityHotspotsByIdeRelativePath())
         .hasEntrySatisfying(Path.of("file/path"), hotspots -> {
           assertThat(hotspots).hasSize(1).allSatisfy(hotspot -> assertThat(hotspot.isRight()).isTrue());
-          assertThat(hotspots).usingRecursiveComparison().ignoringFields("right.id")
+          assertThat(hotspots).usingRecursiveComparison().ignoringFields("lsp4jEither.right.id")
```
damien-urruty-sonarsource, line 99: As I commented earlier, I don't like to expose this thing in tests. But I'm not sure there is a better way

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/Lsp4jUtils.java
damien-urruty-sonarsource, line 1: Small nitpick: this is not related to the protocol itself, it's more a utility to help during deserialization. I would suggest to move it to a different packege

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/adapter/EitherTypeAdapter.java
```diff
@@ -0,0 +1,138 @@
...
+
+    ParameterizedTypeImpl(Type ownerType, Type rawType, Type[] typeArguments) {
+      this.ownerType = ownerType;
+      this.rawType = rawType;
+      this.actualTypeArguments = typeArguments;
+    }
+
+    @Override
+    public Type getOwnerType() {
+      return ownerType;
```
damien-urruty-sonarsource, line 96: After reading the Javadoc for this class, I think we could return null here everytime

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/adapter/EitherTypeAdapter.java
```diff
@@ -0,0 +1,138 @@
...
+      return result.toString();
+    }
+
+    private static String toString(Type type) {
+      if (type instanceof Class<?>) {
+        return ((Class<?>) type).getName();
+      } else {
+        return String.valueOf(type);
+      }
+    }
```
damien-urruty-sonarsource, line 134: I'm tempted to remove this to try and keep it simple

damien-urruty-sonarsource (approved):
LGTM, some minor comments

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/Lsp4jUtils.java
vojtech-suchy-sonarsource: The `sonarlint-rpc-protocol` artefact is self-contained and doesn't declare any other dependencies, so I would probably keep it.

damien-urruty-sonarsource: Ah I see now, in fact I think it's the whole `org.sonarsource.sonarlint.core.rpc.protocol.adapter` package that should be moved out of the `rpc-protocol` module. These are low-level stuff about how to serialize/deserialize things, I don't think it belongs to the protocol itself. And this new `Lsp4jUtils` class could also belong to this `adapter` package. But this can be kept for later if you prefer

PR #941, opened 2024-04-17
SLCORE-733 Publish artifacts needed by clients to Maven Central

serhat-yenican-sonarsource:
There are two commits:
1. Remove embedded modules from the parent pom because we don't use them currently. 
2. Create a dummy Javadoc jar for these modules to pass the deploy-check in case we decide to use them in the future
    - It doesn't generate Javadoc by default because these modules don't have any source files.

### DIFF
--- a/backend/embedded/javadoc/README.md
+++ b/backend/embedded/javadoc/README.md
@@ -1,0 +1,1 @@
+# Embedded backend modules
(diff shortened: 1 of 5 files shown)


vojtech-suchy-sonarsource (approved):
LGTM 🚢

PR #951, opened 2024-04-26
SLCORE-780 Do not fail to promote languages when none was detected

damien-urruty-sonarsource:


kirill-knize-sonarsource (approved):

backend/core/src/main/java/org/sonarsource/sonarlint/core/promotion/PromotionService.java
thahnen, line 1: Wouldn't it be easier to not fix it in `promoteExtraEnabledLanguagesInConnectedMode()` but already in the if statement of `onAnalysisFinished()`? Like catching if the detected languages are empty.
This way `promoteExtraEnabledLanguagesInConnectedMode()` doesn't even have to be called and we save some CPU cycles :smile: Ain't much but a lot of "not much" is speeding up everything

damien-urruty-sonarsource: I guess it's a common practice for me to use private methods to document the code by giving it a meaningful name. I think the impact on the CPU is negligible, and here I prefer readability over "performance". So instead I modified the code again for even better readability :)

PR #952, opened 2024-04-29
SLCORE-777 Perform issue tracking before raising issue

kirill-knize-sonarsource:


rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/analysis/AnalysisRpcService.java
```diff
@@ -65,8 +65,19 @@ public interface AnalysisRpcService {
   CompletableFuture<GetAutoDetectedNodeJsResponse> getAutoDetectedNodeJs();
 
   /**
+   * @deprecated
    * Analyze the provided files.
+   * Use {@link #analyzeFilesAndTrack} instead.
+   * It will require implementing {@link org.sonarsource.sonarlint.core.rpc.protocol.SonarLintRpcClient#publishIssues}
+   * on the client side to get tracked issues.
    */
+  @Deprecated(since = "10.2")
```
damien-urruty-sonarsource, line 74: If we deprecate this one, we can also deprecate `didRaiseIssues` for the client

damien-urruty-sonarsource (approved):
SOme minor comments, else LGTM

PR #953, opened 2024-04-30
SLCORE-776 Store tracked issues from previous analysis

damien-urruty-sonarsource:


kirill-knize-sonarsource (approved):
LGTM. 
For the duplication whatever solution is fine. Either abstract class or just move logic to static methods where possible. We just need to remember that it's a temporary solution and not intended design.

PR #955, opened 2024-04-30
SLCORE-737: Create OSGi bundle for Java Client

thahnen:
## TL;DR:

Using shading/relocating, an OSGi bundle for Java Clients (e.g., Eclipse) was created. Due to conflicts in the Maven Shade and Bndtools plug-in, a custom transformer for the META-INF/MANIFEST.MF file had to be created.

## Summary

OSGi bundles using [bnd](https://github.com/bndtools/bnd) are not created for `sonarlint-rpc-implementation`, `sonarlint-rpc-java-client`, `sonarlint-java-client-legacy` and `sonarlint-java-client-utils` anymore. There is only one bundle called `sonarlint-java-client-osgi` that contains these four as well as the dependencies that are either shaded and relocated (LSP4J, GSON) or just shaded (ProtoBuf) or are Eclipse-specific for the bundle classpath.

The new module `sonarlint-java-client-dependencies` is currently just used for being consumed inside `sonarlint-java-client-osgi` as both modules cannot be merged to due conflicts between the Maven Shade and bnd plug-in used. The upside is that it is easier now to specify which dependencies should be put into the OSGi bundle and which shaded modules.

As said, there are conflicts between the Maven Shade and the bnd plug-in used. I've tried the only other option, which is the [Apache Felix plug-in](https://felix.apache.org/documentation/subprojects/apache-felix-maven-bundle-plugin-bnd.html) that we used until 9.8, but it works even worse.
The solution was to write a custom transformer for the MANIFEST.MF in order to take into account the file created by bnd for the normal or sources JAR while still handling the shading and relocating correctly.

## Testing

The testing is done on the Eclipse side by consuming the new and single bundle; see [SLE-793](https://github.com/SonarSource/sonarlint-eclipse/pull/656). Currently, all tests are successful.

[SLE-793]: https://sonarsource.atlassian.net/browse/SLE-793?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shared.bnd
thahnen, line 1: Dropped because only the module `sonarlint-java-client-osgi` is using [bnd](https://github.com/bndtools/bnd) now.

medium-tests/src/test/java/mediumtest/websockets/WebSocketMediumTests.java
thahnen, line 1: Changes to this file are coming from the cherry-picked commit from Damien. Tried to rebase instead before and it failed all the time - drove me crazy.

buildSrc/README.md
thahnen, line 1: I choose to name the folder `buildSrc` because it is used by Gradle for example

buildSrc/README.md
```diff
@@ -0,0 +1,18 @@
+# SonarLint Core: Build dependencies & logic
+
+This directory contains libraries, tools or scripts that are used by the build, e.g. as
+dependencies. These are not mend to be published but will be used / build for every build.
```
vojtech-suchy-sonarsource, line 4: [mend](https://www.mend.io/) to *meant*

pom.xml
```diff
@@ -24,6 +24,7 @@
   </issueManagement>
 
   <modules>
+    <module>buildSrc/maven-shade-ext-bnd-transformer</module>
```
vojtech-suchy-sonarsource, line 27: As discussed over Slack: 
At the moment, all modules defined in [modules](https://github.com/SonarSource/sonarlint-core/blob/d73add3c7c8db586ede432b2a995966a87da0b8a/pom.xml#L26) section are published to Maven Central as those modules are direct or transitive dependencies needed by clients. 
Also, all those modules generate a javadoc.jar as that is needed for a successful deployment to Maven Central (we discovered this the hard way in [BUILD-4655](https://sonarsource.atlassian.net/browse/BUILD-4655)).

So for the successful release of SLCore, these new modules will also have to generate a javadoc jar (can be a dummy one, eg. [here](https://github.com/SonarSource/sonarlint-core/blob/eadfd469b0bfd77256baee561d0af4110f73774d/backend/embedded/pom.xml#L133)) or define some maven property (is there one?) which will exclude it from deployment.

vojtech-suchy-sonarsource: @serhat-yenican-sonarsource has discovered https://github.com/SonarSource/sonarlint-core/blob/403cbb13f4967aa00be8c1e2bce44310279ff02c/its/pom.xml#L20

buildSrc/README.md
thahnen: Done

pom.xml
thahnen: Done

vojtech-suchy-sonarsource (approved):
LGTM; 

Make sure that we don't need a dummy content to generate the dummy javadoc jar (eg. [Readme](https://github.com/SonarSource/sonarlint-core/blob/eadfd469b0bfd77256baee561d0af4110f73774d/backend/embedded/javadoc/README.md) )

PR #956, opened 2024-04-30
Fix flaky tests around WebSockets

damien-urruty-sonarsource:


damien-urruty-sonarsource:
@kirill-knize-sonarsource some tests were flaky due to some tests sometimes going too fast and incomplete checks:
* in many tests we were checking if a WebSocket connection was opened at the end of each test setup, before doing further checks
* we were only checking if the connection was opened, we were not checking if the subscription correctly happened
* in some further assertions, we were checking the messages received by the server

When those tests are too fast, the connection is opened but the subscription messages are not received yet by the server. Further assertions were making the assumptions that those subscription messages were received. 

I fixed it in almost each test by verifying the connection is opened AND subscription received before going further with the test

kirill-knize-sonarsource (approved):
Great fix. LGTM!

PR #960, opened 2024-05-03
Fix flaky tests (again)

damien-urruty-sonarsource:


damien-urruty-sonarsource:
Real issue discovered, see SLCORE-787

PR #962, opened 2024-05-06
Fix line matching problem

kirill-knize-sonarsource:


damien-urruty-sonarsource (approved):
Some minor comments, else LGTM!

PR #964, opened 2024-05-07
SLCORE-788 Perform hotspots tracking before raising hotspot

kirill-knize-sonarsource:


backend/core/src/main/java/org/sonarsource/sonarlint/core/tracking/IssueMatchingService.java
```diff
@@ -256,12 +234,28 @@ private void processEvent(AnalysisFinishedEvent event) {
...
+    return newIssues;
+  }
+
+  public static Map<URI, List<RaisedIssueDto>> getIssuesToRaise(Map<Path, List<TrackedIssue>> updatedIssues) {
     return updatedIssues.values().stream().flatMap(Collection::stream)
       .collect(groupingBy(TrackedIssue::getFileUri, Collectors.mapping(DtoMapper::toRaisedIssueDto, Collectors.toList())));
   }
 
-  public void storeTrackedIssues(XodusKnownIssuesStore knownIssuesStore, String configurationScopeId, Path clientRelativePath, List<TrackedIssue> newKnownIssues) {
+  static void storeTrackedIssues(XodusKnownIssuesStore knownIssuesStore, String configurationScopeId, Path clientRelativePath, List<TrackedIssue> newKnownIssues) {
```
damien-urruty-sonarsource, line 258: The fact that we use the same method both for hotspots and issues means that the last one to call it might overwrite what was previously stored. I will push a fix on this PR

PR #970, opened 2024-05-13
SLCORE-737: Simplify / remove duplication

thahnen:
## Summary

The OSGi bundle created via Bndtools copied the exported packages from the included resources (other modules' JAR archives) into the shaded one.

After getting in touch with the Bndtools developers a simpler solution was found (that [lacks documentation](https://bnd.bndtools.org/instructions/exportcontents.html)) that does not include copying exported packages from the included JAR archives.

## Information

A small explanation of the `Export-Packages` manifest header and how it is computed additionally using the `-exportcontents` instruction can be found [HERE](https://bnd.discourse.group/t/more-information-about-exportcontent/214/2)!

### DIFF
--- a/client/java-client-osgi/java-client-osgi.bnd
+++ b/client/java-client-osgi/java-client-osgi.bnd
@@ -2,20 +2,22 @@
 -include: shared.bnd
 
 # Manifest entries to configure the OSGi attributes for the normal JAR archive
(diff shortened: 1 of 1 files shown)


vojtech-suchy-sonarsource (approved):
Nice :shipit:

PR #972, opened 2024-05-13
Add hotspot status to raised hotspot

kirill-knize-sonarsource:


backend/core/src/main/java/org/sonarsource/sonarlint/core/tracking/TrackedIssue.java
```diff
@@ -167,6 +170,10 @@ public URI getFileUri() {
     return fileUri;
   }
 
+  public HotspotStatus getHotspotStatus() {
```
damien-urruty-sonarsource, line 173: Maybe `CheckForNull` too?

PR #977, opened 2024-05-13
SLCORE-783 Fix server to client path translation when there is no server prefix

damien-urruty-sonarsource:


vojtech-suchy-sonarsource (approved):
LGTM!

PR #981, opened 2024-05-14
SLCORE-708 Use recommended methods for deprecated methods of 3rd party libraries

serhat-yenican-sonarsource:
# For SonarSourcers:

- [ ] When changing an API:

# For external contributors:


backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/container/analysis/filesystem/FileMetadata.java
```diff
@@ -161,10 +161,13 @@ public Metadata readMetadata(InputStream stream, Charset encoding, URI fileUri,
 
   private static InputStream streamFile(File file) {
     try {
-      return new BOMInputStream(new FileInputStream(file),
-        ByteOrderMark.UTF_8, ByteOrderMark.UTF_16LE, ByteOrderMark.UTF_16BE, ByteOrderMark.UTF_32LE, ByteOrderMark.UTF_32BE);
+      return BOMInputStream.builder().setInputStream(new FileInputStream(file))
```
damien-urruty-sonarsource, line 164: I have the impression this `streamFile` is dead code. Maybe we could instead remove the method?

damien-urruty-sonarsource (approved):
I think there is an opportunity to remove code.

I also have the impression this is related to some deprecations introduced by the bump of commons-compress, so instead of NO-JIRA, we could link it to SLCORE-708?

backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/container/analysis/filesystem/FileMetadata.java
serhat-yenican-sonarsource: Good catch! I'll remove the code and attach the ticket to the related Jira ticket. Thanks

PR #986, opened 2024-05-15
SLCORE-774 Do not send empty string suggestions to the client

kirill-knize-sonarsource:
In the SonarLint configuration file, it's possible to have content with a "valid" connection config with empty values. Today, we report these empty lines as suggestions when we detect such config files. This fix ensures that we will only show useful suggestions for the user in this case.

### DIFF
--- a/medium-tests/src/test/java/mediumtest/ConnectionSuggestionMediumTests.java
+++ b/medium-tests/src/test/java/mediumtest/ConnectionSuggestionMediumTests.java
@@ -31,10 +31,14 @@ import java.util.Map;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.stream.Stream;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
 import org.junit.jupiter.api.io.TempDir;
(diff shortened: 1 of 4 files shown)


medium-tests/src/test/java/mediumtest/ConnectionSuggestionMediumTests.java
```diff
@@ -383,4 +409,14 @@ void should_suggest_sonarcloud_connection_when_pascal_case(@TempDir Path tmp) th
     assertThat(connectionSuggestion.get(CONFIG_SCOPE_ID).get(0).isFromSharedConfiguration()).isTrue();
   }
 
+  private static Stream<Arguments> emptyBindingSuggestionsTestValueProvider() {
+    return Stream.of(
+      Arguments.of("", "", ""),
+      Arguments.of("", "foo", ""),
+      Arguments.of("", "", "foo"),
+      Arguments.of("key", "foo", ""),
```
vojtech-suchy-sonarsource, line 417: As discussed on our call, it would be good to have a test which asserts `fakeClient.getSuggestionsByConfigScope()` is non-empty for this case (which has a projectKey and SQ URL) +same for the two valid cased bellow

vojtech-suchy-sonarsource (approved):
The production code looks good; see minor comment 👍 

Please update the tests as discussed.

vojtech-suchy-sonarsource (approved):
Excellent ❗

PR #987, opened 2024-05-16
SLCORE-808 Create default plugin storage file if no plugin to download

serhat-yenican-sonarsource:
# For SonarSourcers:

- [ ] When changing an API:

# For external contributors:

### DIFF
--- a/backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/PluginsSynchronizer.java
+++ b/backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/PluginsSynchronizer.java
@@ -61,8 +61,13 @@ public class PluginsSynchronizer {
     var pluginsToDownload = serverPlugins.stream()
       .filter(p -> shouldDownload(p, storedPluginsByKey))
       .collect(Collectors.toList());
+
+    if (pluginsToDownload.isEmpty()) {
+      storage.plugins().storeNoPlugins();
+      return new PluginSynchronizationSummary(false);
+    }
     downloadAll(serverApi, pluginsToDownload, cancelMonitor);
-    return new PluginSynchronizationSummary(!pluginsToDownload.isEmpty());
+    return new PluginSynchronizationSummary(true);
   }
 
   private void downloadAll(ServerApi serverApi, List<ServerPlugin> pluginsToDownload, SonarLintCancelMonitor cancelMonitor) {
--- a/backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/storage/PluginsStorage.java
+++ b/backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/storage/PluginsStorage.java
@@ -128,4 +128,22 @@ public class PluginsStorage {
       return Collections.emptyList();
     }
   }
+
+  public void storeNoPlugins() {
+    if (!Files.exists(pluginReferencesFilePath)) {
+      createPluginDirectory();
+      rwLock.write(() -> {
+        var references = Sonarlint.PluginReferences.newBuilder().build();
+        ProtobufFileUtil.writeToFile(references, pluginReferencesFilePath);
+      });
+    }
+  }
+
+  private void createPluginDirectory() {
+    try {
+      Files.createDirectories(pluginReferencesFilePath.getParent());
+    } catch (IOException e) {
+      throw new StorageException(String.format("Cannot create plugin file directory: %s", rootPath), e);
+    }
+  }
 }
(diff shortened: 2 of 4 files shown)


backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/storage/PluginsStorage.java
```diff
@@ -128,4 +128,22 @@ private List<File> getUnknownFiles() {
       return Collections.emptyList();
     }
   }
+
+  public void setReady() {
```
damien-urruty-sonarsource, line 132: I suppose the `setReady` name refers to the analysis readiness? This class is about the storage itself, pretty low-level, so I would avoid a name referring to how it is used. Maybe something like `setEmpty` or `storeNoPlugins` (I'm not good at naming...)

medium-tests/src/test/java/mediumtest/synchronization/PluginSynchronizationMediumTests.java
```diff
@@ -154,7 +154,7 @@ void it_should_not_pull_plugins_that_do_not_support_sonarlint() {
       .build(client);
 
     waitAtMost(3, SECONDS).untilAsserted(() -> {
-      assertThat(getPluginsStorageFolder()).doesNotExist();
+      assertThat(getPluginsStorageFolder().toFile().listFiles()).hasSize(1);
```
damien-urruty-sonarsource, line 157: I suggest checking a little bit further to verify whether the file's content is what we expect.

backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/storage/PluginsStorage.java
```diff
@@ -128,4 +128,22 @@ private List<File> getUnknownFiles() {
...
+    if(!Files.exists(pluginReferencesFilePath)) {
```
damien-urruty-sonarsource, line 133: Nitpick: Maybe a mall formatting issue here, should be `if (` with a space

backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/PluginsSynchronizer.java
```diff
@@ -61,8 +61,13 @@ public PluginSynchronizationSummary synchronize(ServerApi serverApi, boolean sup
     var pluginsToDownload = serverPlugins.stream()
       .filter(p -> shouldDownload(p, storedPluginsByKey))
       .collect(Collectors.toList());
+
+    if(pluginsToDownload.isEmpty()) {
```
damien-urruty-sonarsource, line 65: Same here. Is it a misconfiguration of the formatter?

damien-urruty-sonarsource (approved):
Some suggestions, overall LGTM! Good catch for this bug

backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/storage/PluginsStorage.java
serhat-yenican-sonarsource: Yes, I was thinking of analysis readiness. I will go with the `storeNoPlugins` option then 👍

medium-tests/src/test/java/mediumtest/synchronization/PluginSynchronizationMediumTests.java
serhat-yenican-sonarsource: Before this change, if no plugins were downloaded, there would be no file under the `/plugins` dir. Now, we create the `pb` file in any case, but there will be no plugin jar files. That's why I checked the file count.

I can also check the file name `plugin_references.pb`. I used to add this check first, but it looked cumbersome, so I removed it. If you think it is enough, I can add it.

backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/storage/PluginsStorage.java
serhat-yenican-sonarsource: Thanks for the remainder. It seems I disabled it for some reason. I reenabled it.

backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/PluginsSynchronizer.java
serhat-yenican-sonarsource: I fixed the config, thanks

PR #999, opened 2024-05-28
Allocate more memory for the Orchestrator cache task

damien-urruty-sonarsource:


kirill-knize-sonarsource (approved):
LGTM. Let's see if it will help with OOM problem.

PR #1000, opened 2024-05-28
SLCORE-824: Remove ProtBuf in OSGi

thahnen:
Starting with SLE 10.3, issue tracking/matching is moved to SLCORE. Therefore, persistence on the IDE side based on ProtoBuf is no longer necessary.

Another benefit of dropping ProtoBuf is that we don't *leak* the ProtoBuf classes anymore. This could have caused an issue if there were another Eclipse IDE plug-in running against another incompatible ProtoBuf version (bundle).


jblievremont (approved):
LGTM 👍🏻 tested in SonarSource/sonarlint-eclipse#672

PR #1001, opened 2024-05-29
SLCORE-825 & SLCORE-672 add dbdjavafrontend plugin

sophio-japharidze-sonarsource:


nquinquenel (approved):
LGTM, maybe you can also create a similar test for DBD Java? They [require](https://github.com/SonarSource/sonar-dataflow-bug-detection/blob/fabcb2e290ebb6f214fc81cfeca37a2ca9291259/frontend/java/pom.xml#L100) to have Java

backend/plugin-commons/src/main/java/org/sonarsource/sonarlint/core/plugin/commons/loading/SonarPluginRequirementsChecker.java
```diff
@@ -173,17 +173,9 @@ private static PluginRequirementsCheckResult checkUnsatisfiedPluginDependency(Pl
...
-      // dbd and dbdpythonfrontend require Python to be working
-      if (!enableDataflowBugDetection) {
+    if (DataflowBugDetection.PLUGIN_ALLOW_LIST.contains(plugin.getKey()) && !enableDataflowBugDetection) {
         LOG.debug("DBD feature disabled. Skip loading plugin '{}'.", plugin.getName());
         return new PluginRequirementsCheckResult(plugin, SkipReason.UnsupportedFeature.INSTANCE);
-      }
-      var pythonPluginResult = currentResultsByKey.get(SonarLanguage.PYTHON.getPluginKey());
-      if (checkForPluginSkipped(pythonPluginResult)) {
-        return processUnsatisfiedDependency(currentResult.getPlugin(), SonarLanguage.PYTHON.getPluginKey());
-      }
```
damien-urruty-sonarsource, line 186: Is it safe to implement SLCORE-672 now? From what I understand a fix was made on the DBD plugin side, released in a recent SQ (10.4?), so what about older versions of SQ?

nquinquenel: The [fix](https://github.com/SonarSource/sonar-dataflow-bug-detection/blob/1.22.0.7341/frontend/python/pom.xml#L91) is part of the 1.22.0 version of DBD.

Checking SQ 10.4 (which is the minimal supported version for DBD), they [have](https://github.com/SonarSource/sonar-enterprise/blob/10.4.0.87286/build.gradle#L248) version 1.22.0 bundled.

Also, DBD is supporting SonarLint since version 1.21, and SQ 10.3 has version 1.19. So I don't think there is a risk there.

IMO that looks fine?

damien-urruty-sonarsource: Indeed :+1:

PR #1005, opened 2024-05-31
SLCORE-130 Support progress report for the analysis

damien-urruty-sonarsource:


backend/core/src/main/java/org/sonarsource/sonarlint/core/sync/SynchronizationService.java
```diff
@@ -169,6 +169,7 @@ private void synchronizeProjectsSync(Map<String, Map<String, Collection<BoundSco
         applicationEventPublisher.publishEvent(new ConfigurationScopesSynchronizedEvent(synchronizedConfScopeIds));
         client.didSynchronizeConfigurationScopes(new DidSynchronizeConfigurationScopeParams(synchronizedConfScopeIds));
       }
+      return null;
```
kirill-knize-sonarsource, line 172: What's the point of changing the task type from Consumer to Function if we return null anyway?

kirill-knize-sonarsource (approved):
LGTM with one question that is not clear.

backend/core/src/main/java/org/sonarsource/sonarlint/core/sync/SynchronizationService.java
damien-urruty-sonarsource: We return null in this case, but not in the case of the analysis which I just introduced. I could also have introduced 2 different overloads, but returning null here is not a big deal IMO

PR #1017, opened 2024-06-13
SLCORE-835 exclude git ignored files from analysis

vojtech-suchy-sonarsource:
# For SonarSourcers:

- [x] When changing an API:
    - [x] Explain in the JavaDoc the purpose of the new API
    - [x] Document the change in [API_CHANGES.md](https://github.com/SonarSource/sonarlint-core/blob/master/API_CHANGES.md)
    - [x] If the change breaks the current API, explicitly communicate those to the impacted consumers prior to merging (eg. IDE squad)


backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/gitblame/GitBlameUtils.java
```diff
@@ -79,7 +79,7 @@ private static UnaryOperator<String> adaptToPlatformBasedPath(UnaryOperator<Stri
     };
   }
 
-  private static Repository buildGitRepository(Path basedir) {
+  public static Repository buildGitRepository(Path basedir) {
```
serhat-yenican-sonarsource, line 82: Since GitUtils is more general, I think this method better fits GitUtils. Do you agree to move it there?

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/GitUtils.java
```diff
@@ -0,0 +1,89 @@
...
+    var rootDir = repository.getDirectory().getParentFile();
+    var gitIgnoreFile = new File(rootDir, GITIGNORE_FILENAME);
+    var ignoreNode = new IgnoreNode();
+    try (var inputStream = new FileInputStream(gitIgnoreFile)) {
+      ignoreNode.parse(inputStream);
+    }
+    return ignoreNode;
+  }
+
+  private static Optional<Boolean> checkIgnored(List<FastIgnoreRule> rules, String entryPath) {
```
serhat-yenican-sonarsource, line 79: I would change the return type to the boolean to make it simpler.

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/GitUtils.java
```diff
@@ -0,0 +1,89 @@
...
+      }).collect(Collectors.toList());
+    } catch (Exception e) {
+      LOG.warn("Error occurred while determining files ignored by Git: ", e);
+      LOG.warn("Considering all files as not ignored by Git");
+      return filePathsToAnalyze;
+    }
+  }
+
+  private static IgnoreNode buildIgnoreNode(Repository repository) throws IOException {
+    var rootDir = repository.getDirectory().getParentFile();
```
serhat-yenican-sonarsource, line 70: `gitRepo.getWorkTree()` is a bit simpler than `repository.getDirectory().getParentFile()`

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/GitUtils.java
```diff
@@ -0,0 +1,89 @@
...
+    // Utility class
+  }
+
+  /**
+   * Assumes the supplied {@param baseDir} or some of its parents is a git repository.
+   * The method checks {@param filePathsToAnalyze} against patterns defined in the .gitignore file and returns
+   * those that are not ignored.
+   * If error occurs during execution then all {@param filePathsToAnalyze} are considered as not ignored.
+   */
+  public static List<URI> filterOutIgnoredFiles(Path baseDir, List<URI> filePathsToAnalyze) {
```
serhat-yenican-sonarsource, line 52: I think `filterOutIgnoredFiles` is a bit too specific as a utility method. What about introducing a new class like `SonarLintGitIgnore` similar to `SonarLintBlameResult`?

```java
public class SonarLintGitIgnore {
  private final IgnoreNode ignoreNode;
  private final Path gitRepoRelativeProjectBaseDir;

  public SonarLintGitIgnore(IgnoreNode ignoreNode, Path gitRepoRelativeProjectBaseDir) {
    this.ignoreNode = ignoreNode;
    this.gitRepoRelativeProjectBaseDir = gitRepoRelativeProjectBaseDir;
  }

  public boolean isIgnored(String entryPath, boolean isDirectory) {
    List<FastIgnoreRule> rules = ignoreNode.getRules();
    // Parse rules in the reverse order that they were read because later rules have higher priority
    for (var i = rules.size() - 1; i > -1; i--) {
      var rule = rules.get(i);
      if (rule.isMatch(entryPath, isDirectory)) {
        return rule.getResult();
      }
    }
    return false;
  }
}
 ```

In this way:
- we could provide a more general git ignore utility, including the directory check
- encapsulate relative path logic
- chance to check one file
- keep the filtering business logic in the analysis service

Wdyt?

File: backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/GitUtils.java [hunk 5]
```diff
@@ -0,0 +1,89 @@
...
+    try {
+      var gitRepo = buildGitRepository(baseDir);
+      var gitRepoRelativeProjectBaseDir = gitRepo.getWorkTree().toPath().relativize(baseDir);
+      var ignoreNode = buildIgnoreNode(gitRepo);
+      return ignoreNode.getRules().isEmpty() ? filePathsToAnalyze : filePathsToAnalyze.stream().filter(uri -> {
+        var fileRelativeToGitRepoPath = gitRepoRelativeProjectBaseDir.resolve(Path.of(uri)).toUri().toString();
```
serhat-yenican-sonarsource, line 58: I think we could remove `.toUri()` part just before converting toString. I am not sure it is necessary for Windows, though.

serhat-yenican-sonarsource (commented):
Nice work!
I have some improvement suggestions.
I will review the test code in the second round once you resolve my comments.

backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisService.java
```diff
@@ -256,6 +257,7 @@ public AnalysisConfiguration getAnalysisConfigForEngine(String configScopeId, Li
     analysisProperties.putAll(inferredAnalysisProperties);
     var baseDir = fileSystemService.getBaseDir(configScopeId);
     var actualBaseDir = baseDir == null ? findCommonPrefix(filePathsToAnalyze) : baseDir;
+    filePathsToAnalyze = GitUtils.filterOutIgnoredFiles(actualBaseDir, filePathsToAnalyze);
```
serhat-yenican-sonarsource, line 260: It would be better to move this filtering to the `toInputFiles` method since serverFileExclusion filtering is already there. Better to keep them together

```java
  private List<ClientInputFile> toInputFiles(String configScopeId, List<URI> fileUrisToAnalyze) {
    return fileUrisToAnalyze.stream()
      .filter(not(fileExclusionService::isExcluded))
      .map(uri -> toInputFile(configScopeId, uri))
      .filter(Objects::nonNull)
      .collect(toList());
  }
 ```

serhat-yenican-sonarsource: Quick question. Is this filtering also necessary for auto-triggered analysis? I think yes. If so, we should filter out the files in `triggerAnalysis` method in `AnalysisService`

vojtech-suchy-sonarsource: the `triggerAnalysis` method will eventually call the `analyze()`  which in turn will call `getAnalysisConfigForEngine()` and the `toInputFiles()` - so I think the original suggestion of moving it to `toInputFiles()` is good?

serhat-yenican-sonarsource: Ah, yes. I missed that.

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/gitblame/GitBlameUtils.java
vojtech-suchy-sonarsource: Fixed in the last commit (34ae30b02d40dedea6c77724771076313aa9ef49)

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/GitUtils.java
vojtech-suchy-sonarsource: Fixed in the last commit (34ae30b02d40dedea6c77724771076313aa9ef49)

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/GitUtils.java
vojtech-suchy-sonarsource: Fixed in the last commit (34ae30b02d40dedea6c77724771076313aa9ef49)

backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisService.java
vojtech-suchy-sonarsource: Moved to `toInputFiles` in the last commit (34ae30b02d40dedea6c77724771076313aa9ef49)

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/GitUtils.java
vojtech-suchy-sonarsource: Fixed in the last commit (34ae30b02d40dedea6c77724771076313aa9ef49)

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/GitUtils.java
vojtech-suchy-sonarsource: Well spotted....yep, it's for Windows as it works fine with URI which is the same as for unix (with the forward slash as opposed to Path)

serhat-yenican-sonarsource (approved):
Nice job!
LGTM

PR #1018, opened 2024-06-13
SLCORE-827 Let clients force an analysis

kirill-knize-sonarsource:


backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisService.java
```diff
@@ -356,8 +359,9 @@ public ActiveRuleDto buildActiveRuleDto(SonarLintRuleDefinition rule) {
     return new ActiveRuleDto(rule.getKey(), rule.getLanguage().getSonarLanguageKey(), rule.getDefaultParams(), null);
   }
 
-  private boolean shouldIncludeRuleForAnalysis(String connectionId, SonarLintRuleDefinition ruleDefinition) {
-    return !ruleDefinition.getType().equals(RuleType.SECURITY_HOTSPOT) || (hotspotEnabled && isHotspotTrackingPossible(connectionId));
+  private boolean shouldIncludeRuleForAnalysis(String connectionId, SonarLintRuleDefinition ruleDefinition, boolean hotspotsOnly) {
+    var isHotspot = ruleDefinition.getType().equals(RuleType.SECURITY_HOTSPOT);
+    return (!isHotspot && !hotspotsOnly) || (isHotspot && hotspotEnabled && isHotspotTrackingPossible(connectionId));
```
damien-urruty-sonarsource, line 364: This condition is a bit hard to read but I'm not sure there is a way to make it simpler...

backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisService.java
```diff
@@ -416,7 +420,9 @@ private List<ActiveRuleDto> buildStandaloneActiveRules() {
 
     var filteredActiveRules = new ArrayList<SonarLintRuleDefinition>();
 
-    var allRulesDefinitions = rulesRepository.getEmbeddedRules();
+    var allRulesDefinitions = rulesRepository.getEmbeddedRules().stream()
+      .filter(rule -> !hotspotsOnly || !rule.getType().equals(RuleType.SECURITY_HOTSPOT))
+      .collect(toList());
```
damien-urruty-sonarsource, line 425: Hotspots are not supported in standalone mode, are they? I find it strange that the 2nd part of the condition was not already there. The first part can be removed

backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisService.java
```diff
@@ -733,15 +740,40 @@ public void didChangeAutomaticAnalysisSetting(boolean enabled) {
...
+  }
+
   private void triggerAnalysisForOpenFiles() {
     openFilesRepository.getOpenFilesByConfigScopeId()
-      .forEach(this::triggerAnalysis);
+      .forEach((configurationScopeId, files) -> triggerForcedAnalysis(configurationScopeId, files, false));
+  }
+
+  private void triggerForcedAnalysis(String configurationScopeId, List<URI> files, boolean hotspotsOnly) {
+    if (isReadyForAnalysis(configurationScopeId)) {
```
damien-urruty-sonarsource, line 768: If not the case, we should at least print a log

backend/rpc-impl/src/main/java/org/sonarsource/sonarlint/core/rpc/impl/AnalysisRpcServiceDelegate.java
```diff
@@ -138,10 +142,31 @@ public void didSetUserAnalysisProperties(DidChangeAnalysisPropertiesParams param
...
     notify(() -> getBean(AnalysisService.class).didChangeAutomaticAnalysisSetting(params.isEnabled()));
   }
 
+  @Override
+  public void analyzeFullProject(AnalyzeFullProjectParams params) {
+    notify(() -> getBean(AnalysisService.class).analyzeFullProject(params.getConfigScopeId(), params.isHotspotsOnly()));
+  }
+
+  @Override
+  public void analyzeFileList(AnalyzeFileListParams params) {
```
damien-urruty-sonarsource, line 156: Nitpick/personal taste: I prefer not to use a type in the method name, like here "list". I would suggest `analyzeFiles`

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/analysis/AnalysisRpcService.java
```diff
@@ -104,4 +104,40 @@ public interface AnalysisRpcService {
...
   void didChangeAutomaticAnalysisSetting(DidChangeAutomaticAnalysisSettingParams params);
+
+  /**
+   * Analyze all files in the project. User file exclusions and .gitignore will be respected.
+   * @param params configuration scope ID, flag to report only hotspots
+   * Issues will be reported to the client via
+   * {@link SonarLintRpcClient#raiseIssues(RaiseIssuesParams)} and {@link SonarLintRpcClient#raiseHotspots(RaiseHotspotsParams)}
+   */
+  @JsonNotification
+  void analyzeFullProject(AnalyzeFullProjectParams params);
```
damien-urruty-sonarsource, line 115: In the ticket, we mentioned all these methods should be requests, and they should return at least an ID, so that clients can connect the dots when issues are raised later

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/analysis/AnalysisRpcService.java
```diff
@@ -104,4 +104,40 @@ public interface AnalysisRpcService {
...
+
+  /**
+   * Analyze all files in the provided list. User file exclusions and .gitignore will be respected.
+   * @param params configuration scope ID, list of files to analyse
+   * Issues will be reported to the client via
+   * {@link SonarLintRpcClient#raiseIssues(RaiseIssuesParams)} and {@link SonarLintRpcClient#raiseHotspots(RaiseHotspotsParams)}
+   */
+  @JsonNotification
+  void analyzeFileList(AnalyzeFileListParams params);
```
damien-urruty-sonarsource, line 124: In SLI, there is one use case for analyzing the active file (the green triangle), but this method may be able to achieve it.
For this use case, I think exclusions are not respected, to be verified. So it might deserve a dedicated method. Or we just stop having this behavior

damien-urruty-sonarsource (approved):
Some comments, overall looks OK. I didn't find anything for clients to cancel an analysis as specified in the ticket. Is it a cut branch?

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/GitUtils.java
```diff
@@ -36,14 +39,31 @@
...
+    try {
+      var repo = buildGitRepository(baseDir);
+      var git = new Git(repo);
+      var status = git.status().call();
+      var uncommittedChanges = status.getUncommittedChanges();
+      return uncommittedChanges.stream()
+        .map(file -> baseDir.resolve(file).toUri())
+        .collect(Collectors.toList());
+    } catch (GitAPIException | IllegalStateException e) {
+      return List.of();
```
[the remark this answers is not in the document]
kirill-knize-sonarsource: Good idea, thanks.

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/GitUtils.java
```diff
@@ -36,14 +39,31 @@
...
+  public static List<URI> getVSCChangedFiles(@Nullable Path baseDir) {
+    if (baseDir == null) {
+      return List.of();
+    }
+    try {
+      var repo = buildGitRepository(baseDir);
+      var git = new Git(repo);
+      var status = git.status().call();
+      var uncommittedChanges = status.getUncommittedChanges();
+      return uncommittedChanges.stream()
```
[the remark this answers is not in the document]
kirill-knize-sonarsource: Changed behavior and test.

backend/rpc-impl/src/main/java/org/sonarsource/sonarlint/core/rpc/impl/AnalysisRpcServiceDelegate.java
kirill-knize-sonarsource: The problem is we already have the `analyzeFiles` method. And I don't have a better name. We can discuss this before merging to master.

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/analysis/AnalysisRpcService.java
kirill-knize-sonarsource: Changed as discussed.

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/analysis/AnalysisRpcService.java
kirill-knize-sonarsource: I don't think we check exclusions for the forced analysis.

PR #1020, opened 2024-06-14
SLCORE-861 Use unix separator when using `git-files-blame` or `jgit` libs

vojtech-suchy-sonarsource:
SLCORE-861 Use unix separator when using `git-files-blame` or `jgit` libs

# For SonarSourcers:

- [ ] Prefix the commit message with the ticket number, i.e. `SLCORE-XXXX` (or use `NO-JIRA` for a trivial change not tracked by a ticket)
- [ ] When changing an API:

### DIFF
--- a/backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/GitBlameUtils.java
+++ b/backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/GitBlameUtils.java
@@ -23,6 +23,7 @@ import java.io.IOException;
 import java.nio.file.Path;
 import java.util.Set;
 import java.util.stream.Collectors;
+import org.apache.commons.io.FilenameUtils;
 import org.eclipse.jgit.api.errors.GitAPIException;
 import org.eclipse.jgit.diff.RawTextComparator;
 import org.eclipse.jgit.lib.ObjectReader;
@@ -55,8 +56,9 @@ public class GitBlameUtils {
   }
 
   private static RepositoryBuilder getVerifiedRepositoryBuilder(Path basedir) {
+    var unixStyleBasedir = convertToUnixPath(basedir);
     RepositoryBuilder builder = new RepositoryBuilder()
-      .findGitDir(basedir.toFile())
+      .findGitDir(unixStyleBasedir.toFile())
       .setMustExist(true);
 
     if (builder.getGitDir() == null) {
@@ -65,8 +67,8 @@ public class GitBlameUtils {
     return builder;
   }
 
-  static SonarLintBlameResult blameWithFilesGitCommand(Repository repo, Set<Path> gitRelativePath) {
-    var pathStrings = gitRelativePath.stream().map(Path::toString).collect(Collectors.toSet());
+  private static SonarLintBlameResult blameWithFilesGitCommand(Repository repo, Set<Path> gitRelativePath) {
+    var pathStrings = gitRelativePath.stream().map(Path::toString).map(FilenameUtils::separatorsToUnix).collect(Collectors.toSet());
     RepositoryBlameCommand blameCommand = new RepositoryBlameCommand(repo)
       .setTextComparator(RawTextComparator.WS_IGNORE_ALL)
       .setMultithreading(true)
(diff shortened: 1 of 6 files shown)


backend/commons/src/test/java/org/sonarsource/sonarlint/core/commons/testutils/GitUtils.java
```diff
@@ -47,7 +48,7 @@ public static Date commit(Git git, String... paths) throws GitAPIException {
     if (paths.length > 0) {
       var add = git.add();
       for (String p : paths) {
-        add.addFilepattern(p);
+        add.addFilepattern(FilenameUtils.separatorsToUnix(p));
```
serhat-yenican-sonarsource, line 51: I am not sure we should convert the path here. Windows tests don't make sense if we convert the path here I think

backend/commons/src/test/java/org/sonarsource/sonarlint/core/commons/GitBlameUtilsTest.java
```diff
@@ -21,49 +21,62 @@
...
+    var sonarLintBlameResult = GitBlameUtils.blameWithFilesGitCommand(projectDir, Set.of(Path.of("fileA")));
     var blameResult = sonarLintBlameResult.getBlameResult();
     assertThat(blameResult.getFileBlameByPath()).hasSize(1);
     assertThat(blameResult.getFileBlameByPath().get("fileA").getCommitDates())
       .hasSize(3)
       .allMatch(date -> date.equals(c1));
   }
+
+  @Test
+  @EnabledOnOs(WINDOWS)
```
serhat-yenican-sonarsource, line 72: I am not entirely sure, but if we create a new folder under the project directory, the path should be in Windows format, right? If that's the case, we don't need to give the path in Windows format hardcoded. We could define the path as 
```java
projectDir.resolve("innerDir").resolve("fileA")
```

In that case, we don't need to use `@EnabledOnOs(WINDOWS)` and test it also in other OSs

Wdyt?

serhat-yenican-sonarsource (commented):
Nice fix!
I have some comments and will continue after your reply

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/SonarLintBlameResult.java
```diff
@@ -44,7 +45,7 @@ public SonarLintBlameResult(BlameResult blameResult) {
   public Optional<Date> getLatestChangeDateForLinesInFile(Path filePath, Collection<Integer> lineNumbers) {
     validateLineNumbersArgument(lineNumbers);
     var fileBlameByPath = blameResult.getFileBlameByPath();
-    var blameForFile = fileBlameByPath.get(filePath.toString());
+    var blameForFile = fileBlameByPath.get(FilenameUtils.separatorsToUnix(filePath.toString()));
```
serhat-yenican-sonarsource, line 48: The same question here. Is it necessary?

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/GitBlameUtils.java
```diff
@@ -65,8 +67,8 @@ private static RepositoryBuilder getVerifiedRepositoryBuilder(Path basedir) {
     return builder;
   }
 
-  static SonarLintBlameResult blameWithFilesGitCommand(Repository repo, Set<Path> gitRelativePath) {
-    var pathStrings = gitRelativePath.stream().map(Path::toString).collect(Collectors.toSet());
+  private static SonarLintBlameResult blameWithFilesGitCommand(Repository repo, Set<Path> gitRelativePath) {
+    var pathStrings = gitRelativePath.stream().map(Path::toString).map(FilenameUtils::separatorsToUnix).collect(Collectors.toSet());
```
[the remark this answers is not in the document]
vojtech-suchy-sonarsource: The relative paths come with windows-style path separators - I'm attaching screenshot from a debugging session 
image. So we need to convert those as well.

serhat-yenican-sonarsource (approved):
LGTM!

PR #1022, opened 2024-06-17

SLCORE-786 allow blame with given content for unsaved files

serhat-yenican-sonarsource:
# For SonarSourcers:

- [ ] Prefix the commit message with the ticket number, i.e. `SLCORE-XXXX` (or use `NO-JIRA` for a trivial change not tracked by a ticket)
- [ ] When changing an API:

# For external contributors:

### DIFF
--- a/backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/GitBlameUtils.java
+++ b/backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/GitBlameUtils.java
@@ -41,12 +41,44 @@ public class GitBlameUtils {
     // Utility class
   }
 
-  public static SonarLintBlameResult blameWithFilesGitCommand(Path baseDir, Set<Path> gitRelativePath, @Nullable UnaryOperator<String> fileContentProvider) {
-    var repo = buildRepository(baseDir);
-    return blameWithFilesGitCommand(repo, gitRelativePath, fileContentProvider);
+  public static SonarLintBlameResult blameWithFilesGitCommand(Path projectBaseDir, Set<Path> projectBaseRelativeFilePaths) {
+    return blameWithFilesGitCommand(projectBaseDir, projectBaseRelativeFilePaths, null);
   }
 
-  private static Repository buildRepository(Path basedir) {
+  public static SonarLintBlameResult blameWithFilesGitCommand(Path projectBaseDir, Set<Path> projectBaseRelativeFilePaths, @Nullable UnaryOperator<String> fileContentProvider) {
+    var gitRepo = buildGitRepository(projectBaseDir);
+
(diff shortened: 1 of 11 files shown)


backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisStartedEvent.java
```diff
@@ -62,4 +64,21 @@ public Set<URI> getFileUris() {
   public boolean isTrackingEnabled() {
     return enableTracking;
   }
+
+  public UnaryOperator<String> getFileContentProvider() {
+    return path -> files.stream()
```
vojtech-suchy-sonarsource, line 69: Do we need to take into account that the  filePath we set via the [RepositoryBlameCommand.seFilePath() will always have the UNIX format](https://github.com/SonarSource/sonarlint-core/blob/8dc549e82f5674928a39f2ea806e4e693a685f87/backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/GitBlameUtils.java#L79)? Do any of the paths here also need to be in UNIX format?

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/GitBlameUtils.java
```diff
@@ -67,12 +71,14 @@ private static RepositoryBuilder getVerifiedRepositoryBuilder(Path basedir) {
     return builder;
   }
 
-  private static SonarLintBlameResult blameWithFilesGitCommand(Repository repo, Set<Path> gitRelativePath) {
+  private static SonarLintBlameResult blameWithFilesGitCommand(Repository repo, Set<Path> gitRelativePath, @Nullable UnaryOperator<String> fileContentProvider) {
```
vojtech-suchy-sonarsource, line 74: This is a minor nit-pick - I usually prefer having two separate constructors so we don't need to update the existing code by passing `nulls` (such as in the tests)

vojtech-suchy-sonarsource (commented):
Over all looks good! Thanks for looking into this.

Mainly had one question on whether we need to convert any paths to be unix style (see the comment below)

backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisStartedEvent.java
serhat-yenican-sonarsource: That's a very good catch! Yes, we should definitely take this into account. Thanks.

serhat-yenican-sonarsource: I modified the `it_should_blame_with_given_contents` test to contain inner path logix to be able to test the forward-back slash scenario.

vojtech-suchy-sonarsource (approved):
Splendid 👏

PR #1025, opened 2024-06-20
SLCORE-689  Support SonarQube 9.9 as the minimal supported version

kirill-knize-sonarsource:


its/tests/src/test/java/its/SonarQubeDeveloperEditionTests.java
```diff
@@ -1183,13 +1148,12 @@ void shouldContainExtendedDescription() throws Exception {
       var ruleDetailsResponse = backend.getRulesService().getEffectiveRuleDetails(new GetEffectiveRuleDetailsParams(configScopeId,
         javaRuleKey("S106"), null)).get();
       var ruleDescription = ruleDetailsResponse.details().getDescription();
-      if (ORCHESTRATOR.getServer().version().isGreaterThan(9, 5)) {
+      if (ORCHESTRATOR.getServer().version().isGreaterThan(10, 1)) {
```
vojtech-suchy-sonarsource, line 1151: Why not 9.9?

vojtech-suchy-sonarsource (commented):
Good job, thanks for looking into this.
I've spotted few places which can be cleaned up - I'm attaching a 
[SQ_9_9_review_suggestions.patch](https://github.com/user-attachments/files/15952817/SQ_9_9_review_suggestions.patch) with the suggestions

its/tests/src/test/java/its/SonarQubeDeveloperEditionTests.java
kirill-knize-sonarsource: Ah, it's a great question. Because it wasn't implemented by 9.9 apparently. Previously we were testing for 7.9, 8,9 LTS and latest, so 9.5 was an OK check to split branches. But now we test specifically for 9.9 LTS (and we would get into this branch for LTS). Turned out description sections are not yet there, so it fails with NPE.

That's why I bumped checked version to 10.1. I'm not sure about exact version, had no time to test, but for latest version of SQ it works. So description sections were introduced somewhere between current latest and 9.9.

vojtech-suchy-sonarsource (approved):
Nice 🥇

PR #1030, opened 2024-06-23
Fix publishing previously analysed files problem

kirill-knize-sonarsource:


vojtech-suchy-sonarsource (approved):
Looks good as far as I understand it!

PR #1035, opened 2024-06-26
SLCORE-875: Move to GH action for Slack notification

thahnen:
Remove the old ("workaround") implementation and replace it with the one provided by the RE team. As the repo states, the changes will only be visible once merged to master!

The behavior will be the same, but the Slack messages might look a bit different.


kirill-knize-sonarsource (approved):
Thanks!

PR #1036, opened 2024-06-27
SLCORE-876 Plugin minimal version check prevents loading of OmniSharp plugin

kirill-knize-sonarsource:


jblievremont (approved):
LGTM 👍🏻 thanks for the quick turnaround!

PR #1053, opened 2024-07-12
SLCORE-885: Drop Burgr in Cirrus CI

thahnen:
With EOL of 31st of July, rather getting rid of Burgr early.


vojtech-suchy-sonarsource (approved):
Nice, thank you for looking into this 👍 

Please do squash with a message referencing the `SLCORE-885` as agreed :pray:

PR #1054, opened 2024-07-15
SLCORE-896 Pass analysis files to client to get inferred properties

kirill-knize-sonarsource:


damien-urruty-sonarsource (approved):
I still think the implementation is not optimal, but it should work. I would be keen to revisit this choice later

PR #1055, opened 2024-07-15
SLCORE-897 Path should be converted to server path before submitting to the web API

kirill-knize-sonarsource:


medium-tests/src/test/java/mediumtest/tracking/IssueTrackingMediumTests.java
```diff
@@ -713,6 +717,55 @@ void it_should_match_a_previously_tracked_issue_with_a_server_issue_when_binding
...
+    var connectionId = "connectionId";
+    var projectKey = "projectKey";
+    var server = newSonarCloudServer(orgKey)
+      .withProject(projectKey, project -> project
+        .withFile("server/path/prefix/pom.xml")
+          .withQualityProfile("qp")
+      )
+      .withQualityProfile("qp", qualityProfile -> qualityProfile.withLanguage("xml")
+        .withActiveRule("xml:S3421", rule -> rule.withSeverity(IssueSeverity.MAJOR)))
+      .withVersion("8.0.0.55884")
```
damien-urruty-sonarsource, line 742: Why do you need a version for SonarCloud?

medium-tests/src/test/java/mediumtest/fixtures/ServerFixture.java
```diff
@@ -158,6 +158,11 @@ public ServerBuilder withPlugin(String pluginKey, UnaryOperator<ServerPluginBuil
       return this;
     }
 
+    public ServerBuilder withVersion(String version) {
+      this.version = version;
+      return this;
+    }
+
```
damien-urruty-sonarsource, line 165: Normally this information is needed only in the SQ context, that's why it's only part of the `newSonarQubeServer` method and not a `with` as usual

damien-urruty-sonarsource (approved):
Some questions about the tests, else LGTM!

medium-tests/src/test/java/mediumtest/tracking/IssueTrackingMediumTests.java
kirill-knize-sonarsource: https://github.com/SonarSource/sonarlint-core/blob/55a175cddc96ad0cc015ec8ad13e7c102d7724b3/backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/LocalStorageSynchronizer.java#L54

This fails for the `null` version and blocks all further sync.

medium-tests/src/test/java/mediumtest/fixtures/ServerFixture.java
kirill-knize-sonarsource: It's not the case with the current flow. We read the version from the storage during sync. See other reply.

PR #1057, opened 2024-07-22
SLCORE-813 Keep track of the user telemetry in the VS context

kirill-knize-sonarsource:


API_CHANGES.md
```diff
@@ -80,6 +80,9 @@
 * Add a method to the `org.sonarsource.sonarlint.core.rpc.protocol.backend.analysis.AnalysisRpcService` to let the client notify the backend with user defined analysis properties
   * `org.sonarsource.sonarlint.core.rpc.protocol.backend.analysis.AnalysisRpcService#didSetUserAnalysisProperties` to set user defined properties
 * For analysis, both user-defined and inferred properties will be merged. If the same property is inferred by the client and provided by the user - the inferred value will be used for analysis.
+* Add a method to `org.sonarsource.sonarlint.core.rpc.protocol.SonarLintRpcClient` to allow the backend to request client-defined file
```
vojtech-suchy-sonarsource, line 83: Forgotten changes?

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/initialize/InitializeParams.java
```diff
@@ -88,6 +89,7 @@ public InitializeParams(
    * @param sonarlintUserHome           Path to SonarLint user home directory. If null, will default to ~/.sonarlint
    * @param standaloneRuleConfigByKey   Local rule configuration for standalone analysis. This configuration will override defaults rule activation and parameters.
    */
+  @Deprecated(since = "10.4")
```
vojtech-suchy-sonarsource, line 92: Should we create a ticket to go through the deprecated methods during hardening? We now have the third very large constructor in this class.

vojtech-suchy-sonarsource (approved):
Overall looks good.  See minor comments.

Now we have two mechanisms for migrating telemetry 
- one for migrating the old telemetry by parsing file
-  one for which lets clients to provide data through InitParams.

Just noting that I would prefer extending the first approach as now we will be forced to have the new TelemetryMigrationDto which is practically  used only by VS.  But regarding the current capacity and the upcoming release schedule I guess we should go with that.

API_CHANGES.md
kirill-knize-sonarsource: Yes, well spotted. Removed. Also, documentation for the new init param was added.

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/initialize/InitializeParams.java
kirill-knize-sonarsource: Good point. Created: https://sonarsource.atlassian.net/browse/SLCORE-914

PR #1058, opened 2024-07-23
SLCORE-901 Print client RPC logs in a file when debug switch enabled

damien-urruty-sonarsource:


kirill-knize-sonarsource (approved):
LGTM

PR #1059, opened 2024-07-23
SLCORE-902 Open fix suggestion in the IDE

nquinquenel:


backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryLocalStorage.java
```diff
@@ -276,6 +293,15 @@ public void incrementShowIssueRequestCount() {
     showIssueRequestsCount++;
   }
 
+  public void incrementFixSuggestionReceivedCount(String suggestionId) {
+    markSonarLintAsUsedToday();
+    this.fixSuggestionReceivedCounter.computeIfAbsent(suggestionId, k -> new TelemetryFixSuggestionReceivedCounter()).incrementFixSuggestionReceivedCount();
+  }
+
+  public void fixSuggestionResolved(String suggestionId, FixSuggestionStatus status, @Nullable Integer snippetIndex) {
```
sophio-japharidze-sonarsource, line 301: In theory, it is possible that user opens the suggestion one day, and resolves it the other day :D 
So I think we should call `markSonarLintAsUsedToday();` here as well.

backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/payload/TelemetryFixSuggestionResolvedPayload.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+package org.sonarsource.sonarlint.core.telemetry.payload;
+
+import com.google.gson.annotations.SerializedName;
+import java.util.List;
+import java.util.Map;
+
+public class TelemetryFixSuggestionResolvedPayload {
+
+  @SerializedName("status_by_suggestion_id")
```
sophio-japharidze-sonarsource, line 28: I am thinking we should name it more distinctively, like `opened_fix_status_by_suggestion_id`, because if at some point we start proposing suggestions inside the IDE directly, the naming will become confusing.

medium-tests/src/test/java/mediumtest/TelemetryMediumTests.java
```diff
@@ -334,6 +336,17 @@ void it_should_record_helpAndFeedbackLinkClicked() {
       "\"helpAndFeedbackLinkClickedCount\":{\"itemId\":{\"helpAndFeedbackLinkClickedCount\":1}}"));
   }
 
+  @Test
+  void it_should_record_fixSuggestionResolved() {
+    setupClientAndBackend();
+
+    backend.getTelemetryService().fixSuggestionResolved(new FixSuggestionResolvedParams("suggestionId", FixSuggestionStatus.ACCEPTED, 0));
+    backend.getTelemetryService().fixSuggestionResolved(new FixSuggestionResolvedParams("suggestionId", FixSuggestionStatus.DECLINED, null));
```
sophio-japharidze-sonarsource, line 344: MINOR: I think this case is a bit unrealistic since it is unlikely the same client will send `snippetIndex` once but not another time. To be more robust, we can test the same thing with 2 different suggestion IDs.

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/client/fix/FileEditDto.java
```diff
@@ -0,0 +1,43 @@
...
+import java.nio.file.Path;
+import java.util.List;
+
+public class FileEditDto {
+
+  private final Path path;
+  private final List<ChangesDto> changes;
+
+  public FileEditDto(Path path, List<ChangesDto> changes) {
+    this.path = path;
```
sophio-japharidze-sonarsource, line 31: can we have clearer name for this? Is it IDE file path? Or full file path? or server path?

sophio-japharidze-sonarsource (approved):
Great! ⭐ 

I added a few minor comments/questions.

backend/core/src/main/java/org/sonarsource/sonarlint/core/embedded/server/ShowFixSuggestionRequestHandler.java
```diff
@@ -0,0 +1,364 @@
...
+    telemetryService.fixSuggestionReceived(new FixSuggestionReceivedParams(showFixSuggestionQuery.getFixSuggestion().suggestionId));
+
+    AssistCreatingConnectionParams serverConnectionParams = createAssistServerConnectionParams(showFixSuggestionQuery);
+
+    requestHandlerBindingAssistant.assistConnectionAndBindingIfNeededAsync(
+      serverConnectionParams,
+      showFixSuggestionQuery.projectKey,
+      (connectionId, configScopeId, cancelMonitor) -> {
+        if (configScopeId != null) {
+          showFixSuggestionForScope(configScopeId, showFixSuggestionQuery.issueKey, showFixSuggestionQuery.branch, showFixSuggestionQuery.fixSuggestion);
```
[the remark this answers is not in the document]
nquinquenel: I kept the same behavior as the usual `Open in IDE` feature, I think that as long as the request was correct (feature enabled & payload correct), it should be considered as a fix suggestion received - even if it fails for any reason afterward

backend/core/src/main/java/org/sonarsource/sonarlint/core/embedded/server/ShowFixSuggestionRequestHandler.java
```diff
@@ -0,0 +1,364 @@
...
+          translation.serverToIdePath(Paths.get(fixSuggestion.fileEdit.path)),
+          fixSuggestion.fileEdit.changes.stream().map(c ->
+            new ChangesDto(
+              new LineRangeDto(c.beforeLineRange.startLine, c.beforeLineRange.endLine),
+              c.before,
+              c.after)
+          ).collect(Collectors.toList())
+        )
+      );
+      client.showFixSuggestion(new ShowFixSuggestionParams(configScopeId, issueKey, branch, fixSuggestionDto));
```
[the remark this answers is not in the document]
nquinquenel: Depending on the UI, the issue key could be used by the client to select the issue and make the link between the fix and the issue. For `branch`, I don't make use of it on the client side, but I think it was used for VS Code in the `Open in IDE` feature. If it's not relevant anymore we could remove it

backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryLocalStorage.java
nquinquenel: Right!

backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/payload/TelemetryFixSuggestionResolvedPayload.java
nquinquenel: I didn't spend too much time thinking about those names but you're right!

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/client/fix/FileEditDto.java
nquinquenel: I will rename this to `idePath` 👍🏻

backend/core/src/main/java/org/sonarsource/sonarlint/core/embedded/server/ShowFixSuggestionRequestHandler.java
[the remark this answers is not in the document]
sophio-japharidze-sonarsource: yes we don't use it on the client side either. with my PR branch will be checked on SLCORE side already

backend/core/src/main/java/org/sonarsource/sonarlint/core/embedded/server/ShowFixSuggestionRequestHandler.java
```diff
@@ -0,0 +1,376 @@
...
+    this.canOpenFixSuggestion = params.getFeatureFlags().canOpenFixSuggestion();
+    this.requestHandlerBindingAssistant = requestHandlerBindingAssistant;
+    this.pathTranslationService = pathTranslationService;
+    this.sonarCloudUrl = sonarCloudActiveEnvironment.getUri().toString();
+  }
+
+  @Override
+  public void handle(ClassicHttpRequest request, ClassicHttpResponse response, HttpContext context) throws HttpException, IOException {
+    var showFixSuggestionQuery = extractQuery(request);
+    if (!canOpenFixSuggestion || !Method.POST.isSame(request.getMethod()) || !showFixSuggestionQuery.isValid()) {
```
kirill-knize-sonarsource, line 99: Regarding the second condition `!Method.POST.isSame(request.getMethod())`, we control the server side, so I'm not sure we need to be that defensive here.

backend/core/src/main/java/org/sonarsource/sonarlint/core/embedded/server/ShowFixSuggestionRequestHandler.java
```diff
@@ -0,0 +1,376 @@
...
+
+    public FixSuggestionPayload getFixSuggestion() {
+      return fixSuggestion;
+    }
+  }
+
+  @VisibleForTesting
+  public static class FixSuggestionPayload {
+    private final FileEditPayload fileEdit;
+    private final String suggestionId;
```
kirill-knize-sonarsource, line 275: In tests, an example of a payload looks like UUID. So the field can be not String but UUID.

kirill-knize-sonarsource (approved):
Minor comments. LGTM.

backend/core/src/main/java/org/sonarsource/sonarlint/core/embedded/server/ShowFixSuggestionRequestHandler.java
nquinquenel: I'd rather keep it, I don't see a reason not to

PR #1061, opened 2024-07-25
SLCORE-909 check if branches match during open fix in IDE

sophio-japharidze-sonarsource:
# For SonarSourcers:

    - [x] Explain in the JavaDoc the purpose of the new API
    - [x] Document the change in [API_CHANGES.md](https://github.com/SonarSource/sonarlint-core/blob/master/API_CHANGES.md)
    - [x] If the change breaks the current API, explicitly communicate those to the impacted consumers prior to merging (eg. IDE squad)


nquinquenel (approved):
LGTM

PR #1069, opened 2024-08-05
NO-JIRA: Removed doublet from API_CHANGES.md

thahnen:
This content was twice in the file for the 10.4 release. The newer one is still kept above.


damien-urruty-sonarsource (approved):
Thanks!

PR #1079, opened 2024-08-19
SLCORE-922 make sure minimum plugin versions are compatible with SQ 9.9

sophio-japharidze-sonarsource:


thahnen (approved):
LGTM!

damien-urruty-sonarsource (approved):
I didn't check the individual plugin versions are correct, otherwise LGTM

PR #1085, opened 2024-08-28
SLCORE-920 Windows symbolic links are not supported

kirill-knize-sonarsource:
I decided to just ignore symlink URIs as a fix. Resolving symlink URIs to real links turned out to be a bit more involved than I expected. A summary of findings is documented in the new improvement ticket:
https://sonarsource.atlassian.net/browse/SLCORE-925


vojtech-suchy-sonarsource (approved):
Looks good 👍

PR #1086, opened 2024-08-29
SLCORE-919 Files are ignored by application of .gitignore when they should not

kirill-knize-sonarsource:
In case the git repo path is the same as the base project dir (which should be very often), `gitRepoRelativeProjectBaseDir` was resolved to an empty path, so the file path was resolved as an absolute one, not a relative one. So, matching wasn't working.
I've tested manually with the provided reproducer repo:
https://github.com/jblievremont/sample-gitignore-priority
Now it behaves as expected.

UPD:
The problem was that we were supposed to do very different things for the case when we used base project dir as a base and in the case when it's a git directory relative to the base project dir. We need to resolve the relative file path to compute if it's ignored, but we should do it in a bit different way. Previously, we had one code doing it, and it was incorrect for one case or another.


thahnen (approved):
LGTM, please squash when merging the PR or force-push earlier to get rid of the `wip` commits.

PR #1095, opened 2024-09-06
SLCORE-938 + SLCORE-940 Implement telemetry metrics system

jblievremont:
[SLCORE-938](https://sonarsource.atlassian.net/browse/SLCORE-938)

[SLCORE-938]: https://sonarsource.atlassian.net/browse/SLCORE-938?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryHttpClient.java
+++ b/backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryHttpClient.java
@@ -22,8 +22,9 @@ package org.sonarsource.sonarlint.core.telemetry;
 import com.google.common.annotations.VisibleForTesting;
 import java.time.OffsetDateTime;
 import java.time.temporal.ChronoUnit;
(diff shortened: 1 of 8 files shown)


kirill-knize-sonarsource (approved):
Looks good.

Comment:
I have noticed many values of enums that are currently not being used. 
Usually we avoid adding code "for future use" or "just in case" because there's a risk that plans will change and it will become a dead code.
I understand it's a first step of improvements for the telemetry, but the spec was not detailed enough to tell we really need all these enum values. Hopefully, these changes are well thought.
Approving PR anyway.

jblievremont:
Yup, the enum values reflect the current state of the [telemetry metrics ADR](https://xtranet-sonarsource.atlassian.net/wiki/spaces/BIZTECH/pages/3231252493/Telemetry+metrics+ADR); even though we don't need them at the moment, I consider good practice to implement the full protocol.

PR #1096, opened 2024-09-11
SLCORE-941: Don't fail on Windows Shortcuts

thahnen:
[SLCORE-941](https://sonarsource.atlassian.net/browse/SLCORE-941)

Following the Windows specification we have to exclude Windows Shortcuts from the analysis as they are no links but actual, binary files.

The exclusion checks if the file name hints to be a shortcut based on its extension. Only then do we check the magic number and decide based on it.

[SLCORE-941]: https://sonarsource.atlassian.net/browse/SLCORE-941?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kirill-knize-sonarsource (approved):
LGTM

PR #1101, opened 2024-09-16
SLCORE-937: Don't fail when checks not possible

thahnen:
[SLCORE-937](https://sonarsource.atlassian.net/browse/SLCORE-937)

Skip the check for symbolic links and Windows shortcuts when filtering URIs for the analysis for protocols / schemes that are not supporting such operations.

If not skipping but excluding these URIs then using other protocols would never be possible.

[SLCORE-937]: https://sonarsource.atlassian.net/browse/SLCORE-937?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nquinquenel (approved):
LGTM

PR #1108, opened 2024-09-25
SLCORE-800 Clean up deprecated classes and methods

kirill-knize-sonarsource:
[SLCORE-800](https://sonarsource.atlassian.net/browse/SLCORE-800)

[SLCORE-800]: https://sonarsource.atlassian.net/browse/SLCORE-800?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


API_CHANGES.md
```diff
@@ -3,6 +3,23 @@
 ## Breaking changes
 
 * New feature flag `enableMonitoring` in `org.sonarsource.sonarlint.core.rpc.protocol.backend.initialize.FeatureFlagsDto` allows clients to opt into monitoring with Sentry
+* Removed `org.sonarsource.sonarlint.core.rpc.protocol.SonarLintRpcClient#didRaiseIssue` and associated types. See `raiseIssues` and `raiseHotspots` instead.
```
kirill-knize-sonarsource, line 6: Should be part of the changes for 10.14. Right now it's 10.13.

backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisService.java
```diff
@@ -698,7 +697,6 @@ private void streamIssue(String configScopeId, UUID analysisId, Issue issue, Con
     if (activeRule != null) {
       var rawIssue = new RawIssue(issue, activeRule);
       rawIssues.add(rawIssue);
-      client.didRaiseIssue(new DidRaiseIssueParams(configScopeId, analysisId, toDto(issue, activeRule)));
```
kirill-knize-sonarsource, line 701: If I remember correctly this is about file level issues that we don't track because they don't have a location. So we decide to not raise them?

its/tests/src/test/java/its/SonarQubeCommunityEditionTests.java
```diff
@@ -98,23 +82,23 @@ static void startBackend() throws IOException {
 
     var serverToClientOutputStream = new PipedOutputStream();
     var serverToClientInputStream = new PipedInputStream(serverToClientOutputStream);
-
+    SonarLintRpcClientDelegate client = newDummySonarLintClient();
```
kirill-knize-sonarsource, line 85: var

kirill-knize-sonarsource (commented):
LGTM, minor comments

API_CHANGES.md
damien-urruty-sonarsource: Good catch! Thanks

backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisService.java
damien-urruty-sonarsource: I think you made this change originally. I don't think it's related to file-level issues, it's just the method we were using at some point to report raw issues one-by-one. It's not useful anymore. But maybe I forgot something

damien-urruty-sonarsource (approved):

PR #1121, opened 2024-10-15
SLCORE-979 Avoid keeping track of non-useful data - rules data

kirill-knize-sonarsource:
[SLCORE-979](https://sonarsource.atlassian.net/browse/SLCORE-979)

[SLCORE-979]: https://sonarsource.atlassian.net/browse/SLCORE-979?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/core/src/main/java/org/sonarsource/sonarlint/core/repository/rules/RulesRepository.java
+++ b/backend/core/src/main/java/org/sonarsource/sonarlint/core/repository/rules/RulesRepository.java
@@ -107,8 +107,9 @@ public class RulesRepository {
   @EventListener
   public void configScopeRemoved(ConfigurationScopeRemovedEvent e) {
     var removedBindingConfiguration = e.getRemovedBindingConfiguration();
-    if (removedBindingConfiguration.isBound() && hasNoMoreBindings(removedBindingConfiguration.getConnectionId())) {
-      evictAll(removedBindingConfiguration.getConnectionId());
+    var connectionId = removedBindingConfiguration.getConnectionId();
+    if (removedBindingConfiguration.isBound() && hasNoMoreBindings(connectionId)) {
+      evictAll(connectionId);
     }
   }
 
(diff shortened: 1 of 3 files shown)


backend/core/src/main/java/org/sonarsource/sonarlint/core/repository/rules/RulesRepository.java
```diff
@@ -100,6 +101,14 @@ public void connectionRemoved(ConnectionConfigurationRemovedEvent e) {
     evictAll(e.getRemovedConnectionId());
   }
 
+  @EventListener
+  public void configScopeRemoved(ConfigurationScopeRemovedEvent e) {
+    var removedBindingConfiguration = e.getRemovedBindingConfiguration();
+    if (removedBindingConfiguration.isBound()) {
+      evictAll(removedBindingConfiguration.getConnectionId());
```
vojtech-suchy-sonarsource, line 108: Is it possible that there are multiple binging configurations bound to the same connection? E.g. User has two open projects, both of which are connected to the same SQ. 
In this situation, if a user closes one project, we would evict the rules associated with the connection and there would be problems for other project using the same connection.
Do we have to check here that there are no other bindings for that connectionId before evicting?

vojtech-suchy-sonarsource (commented):
Please see the question related to a situation when there are multiple config copes using the same connectionId

backend/core/src/main/java/org/sonarsource/sonarlint/core/repository/rules/RulesRepository.java
```diff
@@ -100,6 +104,18 @@ public void connectionRemoved(ConnectionConfigurationRemovedEvent e) {
     evictAll(e.getRemovedConnectionId());
   }
 
+  @EventListener
+  public void configScopeRemoved(ConfigurationScopeRemovedEvent e) {
+    var removedBindingConfiguration = e.getRemovedBindingConfiguration();
+    if (removedBindingConfiguration.isBound() && hasNoMoreBindings(removedBindingConfiguration.getConnectionId())) {
```
vojtech-suchy-sonarsource, line 110: nit pick: we can extract `removedBindingConfiguration.getConnectionId()` to a var as it is used on two places

vojtech-suchy-sonarsource (approved):
LGTM; one nit pick

PR #1122, opened 2024-10-15
SLCORE-974 Allow clients to declare the paths to the C# analyzers

jblievremont:
SLCORE-974

PR #1123, opened 2024-10-18
SLCORE-988 allow clients to request which c# analyser to use

sophio-japharidze-sonarsource:
[SLCORE-988](https://sonarsource.atlassian.net/browse/SLCORE-988)

[SLCORE-988]: https://sonarsource.atlassian.net/browse/SLCORE-988?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/API_CHANGES.md
+++ b/API_CHANGES.md
@@ -5,6 +5,7 @@
 * Introduce a new method `org.sonarsource.sonarlint.core.rpc.protocol.backend.analysis.AnalysisRpcService.shouldUseEnterpriseCSharpAnalyzer` to allow clients to know what kind of C# analyzer should be used for the analysis
   * The method returns a boolean value indicating whether the enterprise C# analyzer should be used or not
   * The method returns `true` if a binding exists for config scope AND the related connected server has the enterprise C# plugin (`csharpenterprise`) installed
+  * The method returns `true` if binding exists with a SonarQube version < 10.8 (i.e. SQ versions that do not include repackaged dotnet analyzer)
 
 # 10.7.1
 
--- a/API_CHANGES.md
+++ b/API_CHANGES.md
@@ -6,6 +6,7 @@
   * The method returns a boolean value indicating whether the enterprise C# analyzer should be used or not
   * The method returns `true` if a binding exists for config scope AND the related connected server has the enterprise C# plugin (`csharpenterprise`) installed
   * The method returns `true` if binding exists with a SonarQube version < 10.8 (i.e. SQ versions that do not include repackaged dotnet analyzer) OR SonarCloud
+  * The method returns `false` in standalone mode or if connected to non-commercial edition of SonarQube with a version >= 10.8
 
 # 10.7.1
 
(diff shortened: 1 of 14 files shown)


backend/core/src/main/java/org/sonarsource/sonarlint/core/sync/SynchronizationService.java
```diff
@@ -336,6 +340,13 @@ private void synchronizeConnectionAndProjectsIfNeededSync(String connectionId, S
     }
   }
 
+  private void updateConnectionDetails(String connectionId, boolean hasEnterprisePlugin) {
```
damien-urruty-sonarsource, line 343: This method is called only during the sync, which means if a connection is not synchronized, the flag will not be set appropriately and will be false. Have you considered reading from the storage in `shouldUseEnterpriseCSharpAnalyzer` instead of introducing new data in the connection?

backend/core/src/main/java/org/sonarsource/sonarlint/core/repository/connection/AbstractConnectionConfiguration.java
```diff
@@ -75,6 +76,14 @@ public boolean isSameServerUrl(String otherUrl) {
     return Objects.equals(myUri, otherUri);
   }
 
+  public void setHasEnterpriseCSharpPlugin(boolean hasEnterpriseCSharpPlugin) {
+    this.hasEnterpriseCSharpPlugin = hasEnterpriseCSharpPlugin;
+  }
+
+  public boolean hasEnterpriseCSharpPlugin() {
+    return hasEnterpriseCSharpPlugin;
+  }
```
damien-urruty-sonarsource, line 85: Related to my previous comment, and maybe it's more of a personal taste. Having setters/getters makes it usually harder to follow the flow: when is it called? Are we calling the setter in all possible situations where it needs to be called? I have been trying recently to avoid setters if possible, and prefer immutable objects

damien-urruty-sonarsource (commented):
A first round of comments, we can discuss tomorrow if needed

API_CHANGES.md
```diff
@@ -1,3 +1,12 @@
+# 10.8
+
+## New features
+
+* Introduce a new method `org.sonarsource.sonarlint.core.rpc.protocol.backend.analysis.AnalysisRpcService.shouldUseEnterpriseCSharpAnalyzer` to allow clients to know what kind of C# analyzer should be used for the analysis
+  * The method returns a boolean value indicating whether the enterprise C# analyzer should be used or not
+  * The method returns `true` if a binding exists for config scope AND the related connected server has the enterprise C# plugin (`csharpenterprise`) installed
+  * The method returns `true` if binding exists with a SonarQube version < 10.8 (i.e. SQ versions that do not include repackaged dotnet analyzer) OR SonarCloud
```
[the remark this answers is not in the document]
sophio-japharidze-sonarsource: This sentence seems correct to me. For SQ versions <10.8, the analyzer used to include ALL rules, including those that are moving to the enterprise analyzer now. So in order for users not to lose those rules, we will use enterprise analyzers.

Standalone will return `false`. I will update the doc.

backend/core/src/main/java/org/sonarsource/sonarlint/core/sync/SynchronizationService.java
sophio-japharidze-sonarsource: No, I haven't thought about this possibility, and it sounds good. Let's discuss.

jblievremont:
Merged as part of #1122

PR #1128, opened 2024-10-23
SLCORE-991 Synchronize the multi-quality mode setting

nquinquenel:
[SLCORE-991](https://sonarsource.atlassian.net/browse/SLCORE-991)

[SLCORE-991]: https://sonarsource.atlassian.net/browse/SLCORE-991?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/ServerInfoSynchronizer.java
+++ b/backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/ServerInfoSynchronizer.java
@@ -22,11 +22,10 @@ package org.sonarsource.sonarlint.core.serverconnection;
 import org.sonarsource.sonarlint.core.commons.Version;
 import org.sonarsource.sonarlint.core.commons.progress.SonarLintCancelMonitor;
 import org.sonarsource.sonarlint.core.serverapi.ServerApi;
-import org.sonarsource.sonarlint.core.serverapi.settings.SettingsApi;
 
 public class ServerInfoSynchronizer {
-  private static final String MIN_MQR_MODE_SUPPORT_VERSION = "10.2";
   private static final String MQR_MODE_SETTING = "sonar.multi-quality-mode.enabled";
+  private static final String MQR_MODE_SETTING_MIN_VERSION = "10.8";
   private final ConnectionStorage storage;
 
   public ServerInfoSynchronizer(ConnectionStorage storage) {
@@ -34,10 +33,10 @@ public class ServerInfoSynchronizer {
   }
 
   public StoredServerInfo readOrSynchronizeServerInfo(ServerApi serverApi, SonarLintCancelMonitor cancelMonitor) {
-    return storage.serverInfo().read()
+    return storage.serverInfo().read(serverApi.isSonarCloud())
       .orElseGet(() -> {
         synchronize(serverApi, cancelMonitor);
-        return storage.serverInfo().read().get();
+        return storage.serverInfo().read(serverApi.isSonarCloud()).get();
       });
   }
 
@@ -45,16 +44,11 @@ public class ServerInfoSynchronizer {
     var serverStatus = serverApi.system().getStatus(cancelMonitor);
     var serverVersionAndStatusChecker = new ServerVersionAndStatusChecker(serverApi);
     serverVersionAndStatusChecker.checkVersionAndStatus(cancelMonitor);
-    var isMQRMode = serverApi.isSonarCloud() || isMQRMode(serverApi.settings(), serverStatus.getVersion(), cancelMonitor);
-    storage.serverInfo().store(serverStatus, isMQRMode);
-  }
-
-  private static boolean isMQRMode(SettingsApi settingsApi, String version, SonarLintCancelMonitor cancelMonitor) {
-    var settingResponse = settingsApi.getGlobalSetting(MQR_MODE_SETTING, cancelMonitor);
-    if (settingResponse == null) {
-      var serverVersion = Version.create(version);
-      return serverVersion.compareToIgnoreQualifier(Version.create(MIN_MQR_MODE_SUPPORT_VERSION)) >= 0;
+    var version = Version.create(serverStatus.getVersion());
+    String isMQRMode = null;
+    if (!serverApi.isSonarCloud() && version.compareToIgnoreQualifier(Version.create(MQR_MODE_SETTING_MIN_VERSION)) >= 0) {
+      isMQRMode = serverApi.settings().getGlobalSetting(MQR_MODE_SETTING, cancelMonitor);
     }
-    return Boolean.TRUE.equals(Boolean.parseBoolean(settingResponse));
+    storage.serverInfo().store(serverStatus, isMQRMode);
   }
 }
--- a/backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/ServerInfoSynchronizer.java
+++ b/backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/ServerInfoSynchronizer.java
@@ -19,10 +19,14 @@
  */
 package org.sonarsource.sonarlint.core.serverconnection;
 
+import org.sonarsource.sonarlint.core.commons.Version;
 import org.sonarsource.sonarlint.core.commons.progress.SonarLintCancelMonitor;
 import org.sonarsource.sonarlint.core.serverapi.ServerApi;
+import org.sonarsource.sonarlint.core.serverapi.settings.SettingsApi;
 
 public class ServerInfoSynchronizer {
+  private static final String MIN_MQR_MODE_SUPPORT_VERSION = "10.2";
+  private static final String MQR_MODE_SETTING = "sonar.multi-quality-mode.enabled";
   private final ConnectionStorage storage;
 
   public ServerInfoSynchronizer(ConnectionStorage storage) {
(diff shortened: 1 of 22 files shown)


backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/system/MultiQualityMode.java
```diff
@@ -0,0 +1,24 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.sonarlint.core.serverapi.system;
+
+public enum MultiQualityMode {
```
sophio-japharidze-sonarsource, line 22: I think a better name here based on the enum values would be something like `SeverityMode` ? I looked in the SonarQube MMFs and there is no general name for this concept :/ 

Just saying because `MultiQualityMode` sounds like the values should be `true` or `false`, if that makes sense 🤷

backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/ServerInfoSynchronizer.java
```diff
@@ -41,6 +43,20 @@ public void synchronize(ServerApi serverApi, SonarLintCancelMonitor cancelMonito
     var serverStatus = serverApi.system().getStatus(cancelMonitor);
     var serverVersionAndStatusChecker = new ServerVersionAndStatusChecker(serverApi);
     serverVersionAndStatusChecker.checkVersionAndStatus(cancelMonitor);
-    storage.serverInfo().store(serverStatus);
+    var multiQualityMode = getMultiQualityMode(serverApi.settings(), cancelMonitor);
+    storage.serverInfo().store(serverStatus, multiQualityMode);
+  }
+
+  private static MultiQualityMode getMultiQualityMode(SettingsApi settingsApi, SonarLintCancelMonitor cancelMonitor) {
```
sophio-japharidze-sonarsource, line 50: Could we already check the server version here and set the value only to either `MQR` or `STANDARD`? This way all the logic can be contained in one method, and there is no need for the `MultiQualityMode.DEFAULT`

backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/system/MultiQualityMode.java
nquinquenel: Ah you're probably right, multi quality mode is basically MQR mode if I understood correctly

backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/ServerInfoSynchronizer.java
nquinquenel: So in this case we only handle MQR or Standard, which means we could probably get rid of the Enum and just keep a boolean isMQRMode basically

backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/ServerInfoSynchronizer.java
```diff
@@ -41,6 +44,16 @@ public void synchronize(ServerApi serverApi, SonarLintCancelMonitor cancelMonito
...
     serverVersionAndStatusChecker.checkVersionAndStatus(cancelMonitor);
-    storage.serverInfo().store(serverStatus);
+    var isMQRMode = isMQRMode(serverApi.settings(), serverStatus.getVersion(), cancelMonitor);
+    storage.serverInfo().store(serverStatus, isMQRMode);
+  }
+
+  private static boolean isMQRMode(SettingsApi settingsApi, String version, SonarLintCancelMonitor cancelMonitor) {
+    var settingResponse = settingsApi.getGlobalSetting("sonar.multi-quality-mode.enabled", cancelMonitor);
+    if (settingResponse == null) {
+      var serverVersion = Version.create(version);
```
sophio-japharidze-sonarsource, line 54: We should not forget to take SonarCloud into account. For SC MQR mode is always `true` (for now)

nquinquenel: Good point!

backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/ServerInfoSynchronizer.java
```diff
@@ -41,6 +45,16 @@ public void synchronize(ServerApi serverApi, SonarLintCancelMonitor cancelMonito
     var serverStatus = serverApi.system().getStatus(cancelMonitor);
     var serverVersionAndStatusChecker = new ServerVersionAndStatusChecker(serverApi);
     serverVersionAndStatusChecker.checkVersionAndStatus(cancelMonitor);
-    storage.serverInfo().store(serverStatus);
+    var isMQRMode = serverApi.isSonarCloud() || isMQRMode(serverApi.settings(), serverStatus.getVersion(), cancelMonitor);
```
damien-urruty-sonarsource, line 48: Maybe not a big deal, I'm surprised you save in the storage the result of this condition and not the raw setting from the server. I was expecting the other way around, save the raw setting and have the big condition checked when reading the value from the storage. That said, I think it doesn't really matter

damien-urruty-sonarsource, line 48: Niptick: I would argue that the first part of the OR condition should be in the `isMQRMode`. Or rename the 2nd part to `isSqMQRMode`

backend/server-connection/src/test/java/org/sonarsource/sonarlint/core/serverconnection/ServerInfoSynchronizerTests.java
```diff
@@ -68,14 +70,19 @@ void it_should_read_version_from_storage_when_available() throws IOException {
...
+        .setKey("sonar.multi-quality-mode.enabled")
+        .setValue("true"))
+      .build());
 
     var storedServerInfo = synchronizer.readOrSynchronizeServerInfo(new ServerApi(mockServer.endpointParams(), HttpClientProvider.forTesting().getHttpClient()), new SonarLintCancelMonitor());
 
     assertThat(storedServerInfo)
-      .extracting("version")
-      .hasToString("9.9");
+      .extracting("version", "isMQRMode")
```
damien-urruty-sonarsource, line 84: Nitpick: you could use getters references here (e.g. `StoredServerInfo::getVersion`), this works better if a renaming occurs

medium-tests/src/test/java/mediumtest/synchronization/ServerInfoSynchronizationMediumTests.java
```diff
@@ -58,8 +58,43 @@ void it_should_pull_server_info_when_bound_configuration_scope_is_added() {
...
+      .withSonarCloudConnection("connectionId", "test")
+      .withFullSynchronization()
+      .build();
+
+    addConfigurationScope("configScopeId", "connectionId", "projectKey");
+
+    waitAtMost(3, SECONDS).untilAsserted(() -> assertThat(getServerInfoFile())
+      .exists()
+      .extracting(this::readServerMode)
+      .isEqualTo(true));
```
damien-urruty-sonarsource, line 97: Is this test realistic? Shouldn't we mimic the fact that this setting does not exist on SC and that the synchronization still works?

backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/storage/ServerInfoStorage.java
```diff
@@ -54,11 +54,11 @@ public Optional<StoredServerInfo> read() {
...
 
-  private static Sonarlint.ServerInfo adapt(ServerInfo serverInfo) {
-    return Sonarlint.ServerInfo.newBuilder().setVersion(serverInfo.getVersion()).build();
+  private static Sonarlint.ServerInfo adapt(ServerStatusInfo serverStatus, boolean isMQRMode) {
+    return Sonarlint.ServerInfo.newBuilder().setVersion(serverStatus.getVersion()).setIsMqrMode(isMQRMode).build();
   }
 
   private static StoredServerInfo adapt(Sonarlint.ServerInfo serverInfo) {
-    return new StoredServerInfo(Version.create(serverInfo.getVersion()));
+    return new StoredServerInfo(Version.create(serverInfo.getVersion()), serverInfo.getIsMqrMode());
```
damien-urruty-sonarsource, line 62: What happens if we try to read a storage that has no value for `isMqrMode`, if the storage has not yet been synchronized?

damien-urruty-sonarsource (approved):
Only nitpicks for me, good job!

backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/ServerInfoSynchronizer.java
nquinquenel: I switched to do the other way around and re-introduced the ENUM

backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/ServerInfoSynchronizer.java
```diff
@@ -19,28 +19,36 @@
...
   }
 
   public void synchronize(ServerApi serverApi, SonarLintCancelMonitor cancelMonitor) {
     var serverStatus = serverApi.system().getStatus(cancelMonitor);
     var serverVersionAndStatusChecker = new ServerVersionAndStatusChecker(serverApi);
     serverVersionAndStatusChecker.checkVersionAndStatus(cancelMonitor);
-    storage.serverInfo().store(serverStatus);
+    var version = Version.create(serverStatus.getVersion());
+    String isMQRMode = null;
+    if (!serverApi.isSonarCloud() && version.compareToIgnoreQualifier(Version.create(MQR_MODE_SETTING_MIN_VERSION)) >= 0) {
```
nquinquenel, line 49: It's to avoid calling this endpoint when we know it does not exist, otherwise 404 error is thrown, WDYT?

backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/storage/ServerInfoStorage.java
damien-urruty-sonarsource: The storage is per connection anyway, and the client should tell if a given connection is SQ or SC

PR #1132, opened 2024-10-24
SLCORE-992 Adapt RaisedFindingDto to include mode-specific data

nquinquenel:
[SLCORE-992](https://sonarsource.atlassian.net/browse/SLCORE-992)

[SLCORE-992]: https://sonarsource.atlassian.net/browse/SLCORE-992?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/core/src/main/java/org/sonarsource/sonarlint/core/DtoMapper.java
+++ b/backend/core/src/main/java/org/sonarsource/sonarlint/core/DtoMapper.java
@@ -27,6 +27,9 @@ import org.sonarsource.sonarlint.core.rpc.protocol.backend.hotspot.HotspotStatus
 import org.sonarsource.sonarlint.core.rpc.protocol.backend.rules.ImpactDto;
 import org.sonarsource.sonarlint.core.rpc.protocol.client.hotspot.RaisedHotspotDto;
 import org.sonarsource.sonarlint.core.rpc.protocol.client.issue.RaisedIssueDto;
+import org.sonarsource.sonarlint.core.rpc.protocol.common.Either;
+import org.sonarsource.sonarlint.core.rpc.protocol.common.MQRModeDetails;
+import org.sonarsource.sonarlint.core.rpc.protocol.common.StandardModeDetails;
 import org.sonarsource.sonarlint.core.rules.RuleDetailsAdapter;
 import org.sonarsource.sonarlint.core.tracking.TrackedIssue;
 
@@ -48,8 +51,10 @@ public class DtoMapper {
         .collect(Collectors.toList()), RuleDetailsAdapter.adapt(ruleDetails.getVulnerabilityProbability()));
   }
 
-  public static RaisedIssueDto toRaisedIssueDto(TrackedIssue issue, NewCodeDefinition newCodeDefinition) {
+  public static RaisedIssueDto toRaisedIssueDto(TrackedIssue issue, NewCodeDefinition newCodeDefinition, boolean isMQRMode) {
     return new RaisedIssueDto(issue.getId(), issue.getServerKey(), issue.getRuleKey(), issue.getMessage(),
+      isMQRMode ? Either.forLeft(new StandardModeDetails(RuleDetailsAdapter.adapt(issue.getSeverity()), RuleDetailsAdapter.adapt(issue.getType())))
+        : Either.forRight(new MQRModeDetails(RuleDetailsAdapter.adapt(issue.getCleanCodeAttribute()), RuleDetailsAdapter.toDto(issue.getImpacts()))),
       RuleDetailsAdapter.adapt(issue.getSeverity()),
       RuleDetailsAdapter.adapt(issue.getType()),
       RuleDetailsAdapter.adapt(issue.getCleanCodeAttribute()), RuleDetailsAdapter.toDto(issue.getImpacts()),
@@ -60,10 +65,12 @@ public class DtoMapper {
       issue.getRuleDescriptionContextKey());
   }
 
-  public static RaisedHotspotDto toRaisedHotspotDto(TrackedIssue issue, NewCodeDefinition newCodeDefinition) {
+  public static RaisedHotspotDto toRaisedHotspotDto(TrackedIssue issue, NewCodeDefinition newCodeDefinition, boolean isMQRMode) {
     var status = issue.getHotspotStatus();
     status = status != null ? status : HotspotStatus.TO_REVIEW;
     return new RaisedHotspotDto(issue.getId(), issue.getServerKey(), issue.getRuleKey(), issue.getMessage(),
+      isMQRMode ? Either.forLeft(new StandardModeDetails(RuleDetailsAdapter.adapt(issue.getSeverity()), RuleDetailsAdapter.adapt(issue.getType())))
+        : Either.forRight(new MQRModeDetails(RuleDetailsAdapter.adapt(issue.getCleanCodeAttribute()), RuleDetailsAdapter.toDto(issue.getImpacts()))),
       RuleDetailsAdapter.adapt(issue.getSeverity()),
       RuleDetailsAdapter.adapt(issue.getType()),
       RuleDetailsAdapter.adapt(issue.getCleanCodeAttribute()), RuleDetailsAdapter.toDto(issue.getImpacts()),
--- a/backend/core/src/main/java/org/sonarsource/sonarlint/core/DtoMapper.java
+++ b/backend/core/src/main/java/org/sonarsource/sonarlint/core/DtoMapper.java
@@ -53,8 +53,8 @@ public class DtoMapper {
 
   public static RaisedIssueDto toRaisedIssueDto(TrackedIssue issue, NewCodeDefinition newCodeDefinition, boolean isMQRMode) {
     return new RaisedIssueDto(issue.getId(), issue.getServerKey(), issue.getRuleKey(), issue.getMessage(),
-      isMQRMode ? Either.forLeft(new StandardModeDetails(RuleDetailsAdapter.adapt(issue.getSeverity()), RuleDetailsAdapter.adapt(issue.getType())))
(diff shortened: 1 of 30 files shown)


damien-urruty-sonarsource (approved):
Looks mostly good to me, you need to update the API_CHANGES.md file regarding the deprecations

backend/core/src/main/java/org/sonarsource/sonarlint/core/issue/IssueService.java
```diff
@@ -338,29 +339,42 @@ private void handleEvent(String connectionId, IssueChangedEvent event) {
   }
 
   private void republishPreviouslyRaisedIssues(String connectionId, IssueChangedEvent event) {
+    var isSonarCloud = serverApiProvider.getServerApiOrThrow(connectionId).isSonarCloud();
+    var isMQRMode = storageService.connection(connectionId).serverInfo().read(isSonarCloud).map(StoredServerInfo::isMQRMode).orElse(false);
```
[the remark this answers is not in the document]
nquinquenel: This is also the comment I posted on the other PR (sorry, it got merged because I saw no further comments). Since I switched to storing raw values, when reading it, I need a way to convert the protobuf enum DEFAULT/STANDARD/MQR to a simple boolean. Otherwise, the StoredServerInfo object will contain the protobuf type, which seems off. But then, the cons of computing it directly when reading is that I need the `isSonarCloud` information, perhaps it should be stored with it?

nquinquenel:
> Looks mostly good to me, you need to update the API_CHANGES.md file regarding the deprecations

Doesn't the API_CHANGES about changes to the API impact the client? The current changes are purely internal and exposing issue details is part of another ticket.

damien-urruty-sonarsource:
`RaisedFindingDto` is a client-facing API

backend/core/src/main/java/org/sonarsource/sonarlint/core/issue/IssueService.java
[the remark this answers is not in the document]
damien-urruty-sonarsource: What I would do:
* we usually try to keep in the storage an image of what's on the server. So I would store the boolean setting that you synchronized. Since it might be absent on older versions or SC, you can use an optional bool. 
* with the optional boolean it means you represent 3 states: unknown, standard or MQR. Identifying the exact mode to use is not the purpose of the storage class IMO, it's more the class that uses the storage class. So in this IssueService (or maybe you can introduce another class if you need to do it in several places) you can determine the real mode based on 3 things: the server kind, the server version and the stored setting

That's just a proposal. You can also keep an enum instead of the boolean if you prefer, I would just avoid passing isSonarCloud to the storage class

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/common/MQRModeDetails.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.util.List;
+import org.sonarsource.sonarlint.core.rpc.protocol.backend.rules.ImpactDto;
+
+public class MQRModeDetails {
+
+  private final CleanCodeAttribute cleanCodeAttribute;
+  private final List<ImpactDto> impacts;
+
+  public MQRModeDetails(CleanCodeAttribute cleanCodeAttribute, List<ImpactDto> impacts) {
```
sophio-japharidze-sonarsource, line 30: This is `org.sonarsource.sonarlint.core.rpc.protocol.common.CleanCodeAttribute`, which does not contain the clean code category. I think we need to explicitly use `org.sonarsource.sonarlint.core.commons.CleanCodeAttribute`, OR add categories to the one that is used

nquinquenel: You should be able to use this util class from the client: https://github.com/SonarSource/sonarlint-core/blob/master/client/java-client-utils/src/main/java/org/sonarsource/sonarlint/core/client/utils/CleanCodeAttribute.java#L59

backend/core/src/main/java/org/sonarsource/sonarlint/core/issue/IssueService.java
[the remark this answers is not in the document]
nquinquenel: I like your proposal, and I made the changes

backend/core/src/main/java/org/sonarsource/sonarlint/core/DtoMapper.java
```diff
@@ -60,10 +65,12 @@ public static RaisedIssueDto toRaisedIssueDto(TrackedIssue issue, NewCodeDefinit
       issue.getRuleDescriptionContextKey());
   }
 
-  public static RaisedHotspotDto toRaisedHotspotDto(TrackedIssue issue, NewCodeDefinition newCodeDefinition) {
+  public static RaisedHotspotDto toRaisedHotspotDto(TrackedIssue issue, NewCodeDefinition newCodeDefinition, boolean isMQRMode) {
     var status = issue.getHotspotStatus();
     status = status != null ? status : HotspotStatus.TO_REVIEW;
     return new RaisedHotspotDto(issue.getId(), issue.getServerKey(), issue.getRuleKey(), issue.getMessage(),
+      isMQRMode ? Either.forLeft(new StandardModeDetails(RuleDetailsAdapter.adapt(issue.getSeverity()), RuleDetailsAdapter.adapt(issue.getType())))
+        : Either.forRight(new MQRModeDetails(RuleDetailsAdapter.adapt(issue.getCleanCodeAttribute()), RuleDetailsAdapter.toDto(issue.getImpacts()))),
```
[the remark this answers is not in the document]
damien-urruty-sonarsource: I would tend to remove things if they are not required. But it might not be in the scope of this effort, so don't hesitate to create a ticket

backend/core/src/main/java/org/sonarsource/sonarlint/core/issue/IssueService.java
```diff
@@ -90,17 +91,19 @@ public class IssueService {
   private final LocalOnlyIssueRepository localOnlyIssueRepository;
   private final ApplicationEventPublisher eventPublisher;
   private final FindingReportingService findingReportingService;
+  private final SeverityModeService modeService;
```
damien-urruty-sonarsource, line 94: Nitpick: `modeService` is also vague, using `severityModeService` would make things more explicit

backend/core/src/main/java/org/sonarsource/sonarlint/core/reporting/FindingReportingService.java
```diff
@@ -60,17 +62,19 @@ public class FindingReportingService {
   private final SonarLintRpcClient client;
   private final ConfigurationRepository configurationRepository;
   private final NewCodeService newCodeService;
+  private final SeverityModeService modeService;
```
damien-urruty-sonarsource, line 65: Same here

backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/StoredServerInfo.java
```diff
@@ -19,22 +19,35 @@
...
+      this.isMQRMode = ModeDetails.STANDARD;
+    }
   }
 
   public Version getVersion() {
     return version;
   }
 
-  public boolean isMQRMode() {
+  public ModeDetails isMQRMode() {
```
damien-urruty-sonarsource, line 45: I find it strange that a `is` method does not return a boolean, maybe worth renaming the method since you rely on an enum?

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/adapter/EitherStandardOrMQRModeAdapterFactory.java
```diff
@@ -0,0 +1,37 @@
...
+import org.sonarsource.sonarlint.core.rpc.protocol.common.MQRModeDetails;
+import org.sonarsource.sonarlint.core.rpc.protocol.common.StandardModeDetails;
+
+public class EitherStandardOrMQRModeAdapterFactory extends CustomEitherAdapterFactory<StandardModeDetails, MQRModeDetails> {
+
+  private static final TypeToken<Either<StandardModeDetails, MQRModeDetails>> ELEMENT_TYPE = new TypeToken<>() {
+  };
+
+  public EitherStandardOrMQRModeAdapterFactory() {
+    super(ELEMENT_TYPE, StandardModeDetails.class, MQRModeDetails.class, new EitherTypeAdapter.PropertyChecker("details"));
```
damien-urruty-sonarsource, line 34: I'm not sure this works. The argument of `PropertyChecker` should be the name of a property present in the left type but not in the right type. Maybe a medium test is missing to verify the either works properly?

damien-urruty-sonarsource (commented):
Looks almost good, I left a few more comments, mostly minor except for the Either that I think is a bug

backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/StoredServerInfo.java
nquinquenel: Ah yes, you're right, I wanted to modify this one because there is confusion with the SeverityModeService that returns the boolean

backend/core/src/main/java/org/sonarsource/sonarlint/core/DtoMapper.java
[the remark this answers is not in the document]
nquinquenel: [Here](https://sonarsource.atlassian.net/browse/SLCORE-1019) is the ticket for reference

PR #1143, opened 2024-10-31
SLCORE-993 refactor abstract rule dto

sophio-japharidze-sonarsource:
SLCORE-993

# For SonarSourcers:

- [x] Prefix the commit message with the ticket number, i.e. `SLCORE-XXXX` if you already have a ticket in Jira
- [ ] For standalone PRs without issue in Jira:
    - [ ] Mention Epic ID in this descrition to create a new Task in Jira
    - [ ] Mention Issue ID in this descrition to create a new Sub-Task in Jira
    - [ ] Do not mention any Jira issue to create a new Task in Jira without a parent
- [x] When changing an API:

# For external contributors:

[SLCORE-1022]: https://sonarsource.atlassian.net/browse/SLCORE-1022?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/core/src/main/java/org/sonarsource/sonarlint/core/tracking/TaintVulnerabilityTrackingService.java
```diff
@@ -246,7 +246,7 @@ private static TaintVulnerabilityDto toDto(ServerTaintIssue serverTaintIssue, Pr
       serverTaintIssue.getMessage(),
       translation.serverToIdePath(serverTaintIssue.getFilePath()),
       serverTaintIssue.getCreationDate(),
-      isMQRMode && cleanCodeAttribute != null ? Either.forRight(new MQRModeDetails(cleanCodeAttribute, impacts))
+      isMQRMode && !impacts.isEmpty() ? Either.forRight(new MQRModeDetails(cleanCodeAttribute, impacts))
```
sophio-japharidze-sonarsource, line 249: When connected to SonarCloud and fetching Taint rule details, we get the Clean Code attribute but not the impacts :D So checking for impacts is a more robust way of telling which mode to use.

nquinquenel: Why are you checking for both `isMQRMode` && `impacts.isEmpty`? `isMQRMode` should be sufficient if you get it from the service

sophio-japharidze-sonarsource: Unfortunately, `isMQRMode` check does not seem sufficient. With SonarCloud, taint vulnerabilities don't seem to have `impacts` set at all.

I think there is something with protobuf mapping, did not have time to check but anyway this additional check guarantees that we get the correct data.

API_CHANGES.md
```diff
@@ -15,6 +15,7 @@
 ## Breaking changes
 
 * Remove the `org.sonarsource.sonarlint.core.serverconnection.ServerPathProvider` class.
+* Remove `severity` and `type` fields from `org.sonarsource.sonarlint.core.rpc.protocol.backend.rules.RuleDefinitionDto` as this class is only used for fetching standalone rule details, which should always have the Clean Code Attribute and Impacts
```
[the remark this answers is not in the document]
sophio-japharidze-sonarsource: We could, but from the `RuleDefinitionDto` we were always supposed to get only the CCT stuff. So, realistically, I think it's easier to just remove it. I could also add them back. WDYT @damien-urruty-sonarsource ?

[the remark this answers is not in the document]
nquinquenel: The problem is that it will break compatibility for people who did not migrate to this new stuff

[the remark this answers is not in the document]
nquinquenel: Summary of the discussion -> We go with the fields removal and do not keep backward compatibility because all flavors will work on it in the coming days

backend/core/src/main/java/org/sonarsource/sonarlint/core/tracking/TaintVulnerabilityTrackingService.java
nquinquenel: Maybe add a TODO on this line to not forget about investigating this, or just a comment. In the meantime I guess we can go with this

nquinquenel (approved):
LGTM

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/rules/EffectiveRuleDetailsDto.java
```diff
@@ -53,4 +62,48 @@ public Either<RuleMonolithicDescriptionDto, RuleSplitDescriptionDto> getDescript
...
+  @CheckForNull
+  public RuleType getType() {
+    return this.severityDetails.isLeft() ?
+      this.severityDetails.getLeft().getType() : null;
+  }
+
+  @CheckForNull
+  public List<ImpactDto> getDefaultImpacts() {
+    return this.severityDetails.isRight() ?
+      this.severityDetails.getRight().getImpacts() : null;
```
nquinquenel, line 93: Could you return an empty list and not null?

backend/core/src/main/java/org/sonarsource/sonarlint/core/tracking/TaintVulnerabilityTrackingService.java
nquinquenel: I'd also keep the `cleanCodeAttribute != null`, to avoid the warning

backend/core/src/main/java/org/sonarsource/sonarlint/core/rules/RuleDetailsAdapter.java
```diff
@@ -78,16 +80,15 @@ private RuleDetailsAdapter() {
   }
 
   public static EffectiveRuleDetailsDto transform(RuleDetails ruleDetails, @Nullable String contextKey) {
-    org.sonarsource.sonarlint.core.commons.IssueSeverity defaultSeverity = ruleDetails.getDefaultSeverity();
-    RuleType type = ruleDetails.getType();
+    var cleanCodeAttribute = ruleDetails.getCleanCodeAttribute().map(RuleDetailsAdapter::adapt).orElse(null);
+    Either<StandardModeDetails, MQRModeDetails> severityDetails = cleanCodeAttribute != null && !ruleDetails.getDefaultImpacts().isEmpty() ?
+      Either.forRight(new MQRModeDetails(cleanCodeAttribute, Objects.requireNonNull(toDto(ruleDetails.getDefaultImpacts())))) :
```
nquinquenel, line 85: you can remove the requireNonNull now that the list can't be null `Objects.requireNonNull(toDto(ruleDetails.getDefaultImpacts())))`

PR #1150, opened 2024-11-04
SLCORE-877 Remove plugin version checks

damien-urruty-sonarsource:
[SLCORE-877](https://sonarsource.atlassian.net/browse/SLCORE-877)

[SLCORE-877]: https://sonarsource.atlassian.net/browse/SLCORE-877?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/PluginsSynchronizer.java
```diff
@@ -58,15 +58,25 @@ public PluginSynchronizationSummary synchronize(ServerApi serverApi, boolean sup
...
+      .map(Map.Entry::getKey)
       .collect(Collectors.toList());
 
     if (pluginsToDownload.isEmpty()) {
       storage.plugins().storeNoPlugins();
+      storage.plugins().cleanUpUnknownPlugins(serverPluginsExpectedInStorage);
       return new PluginSynchronizationSummary(false);
     }
     downloadAll(serverApi, pluginsToDownload, cancelMonitor);
+    storage.plugins().cleanUpUnknownPlugins(serverPluginsExpectedInStorage);
```
thahnen, line 79: If I understand this correctly, it is now also cleaning up the old leftover plug-ins like older versions of a analyzer? If yes then this would be neat!

damien-urruty-sonarsource: Correct. In fact any file that does not correspond to a current plugin on the server will be removed

kirill-knize-sonarsource (approved):
LGTM!

PR #1157, opened 2024-11-11
SLCORE-1032 Wrong token error during sync

kirill-knize-sonarsource:
[SLCORE-1032](https://sonarsource.atlassian.net/browse/SLCORE-1032)

[SLCORE-1032]: https://sonarsource.atlassian.net/browse/SLCORE-1032?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/API_CHANGES.md
+++ b/API_CHANGES.md
@@ -5,6 +5,8 @@
 * Signature of `org.sonarsource.sonarlint.core.rpc.protocol.backend.file.DidUpdateFileSystemParams#DidUpdateFileSystemParams` was changed
   * Parameter `addedOrChangedFiles` was split into `addedFiles` and `changedFiles`
 * Removed parameter `branch` and `pullRequest` from `org.sonarsource.sonarlint.core.rpc.protocol.client.issue.IssueDetailsDto` as it should not be used anymore by the client.
+* Add new client method `org.sonarsource.sonarlint.core.rpc.client.SonarLintRpcClientDelegate#invalidToken`
+  * It is called when the token is invalid and the client should notify the user about it
 
 # 10.10
 
--- a/backend/core/src/main/java/org/sonarsource/sonarlint/core/ConnectionManager.java
+++ b/backend/core/src/main/java/org/sonarsource/sonarlint/core/ConnectionManager.java
@@ -19,6 +19,7 @@
  */
 package org.sonarsource.sonarlint.core;
 
+import com.google.common.annotations.VisibleForTesting;
 import java.net.URI;
 import java.util.Optional;
 import javax.annotation.Nullable;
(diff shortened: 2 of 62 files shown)


medium-tests/src/test/java/mediumtest/synchronization/ConnectionSyncMediumTests.java
```diff
@@ -137,6 +139,41 @@ void it_should_sync_when_credentials_are_updated() {
...
+
+    var introductionDate = Instant.now().truncatedTo(ChronoUnit.MILLIS);
+    var server = newSonarQubeServer()
+      .withProject("projectKey",
+        project -> project.withBranch("main",
+          branch -> branch.withTaintIssue("issueKey", "rule:key", "message", "author", "file/path", "OPEN", null, introductionDate, new TextRange(1, 2, 3, 4),
+            RuleType.VULNERABILITY)))
+      .start();
+
+    server.getMockServer().stubFor(get("/api/system/status").willReturn(aResponse().withStatus(status)));
```
damien-urruty-sonarsource, line 158: I think it would make sense to improve the server fixture instead of accessing the internals. It might be useful for other cases in the future, and it hides how we implement the fixture

backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/ServerApi.java
```diff
@@ -69,7 +77,7 @@ public OrganizationApi organization() {
   }
 
   public IssueApi issue() {
-    return new IssueApi(helper);
+    return new IssueApi(helper, wrongTokenCallbackWithConnectionIdParam);
```
damien-urruty-sonarsource, line 80: Why is the IssueApi the only one that accepts this parameter, and not other Api classes?

backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/ServerApiHelper.java
```diff
@@ -135,13 +137,15 @@ public static String concat(String baseUrl, String relativePath) {
       (relativePath.startsWith("/") ? relativePath.substring(1) : relativePath);
   }
 
-  public static RuntimeException handleError(HttpClient.Response toBeClosed) {
+  public static RuntimeException handleError(HttpClient.Response toBeClosed, Runnable invalidTokenCallback) {
     try (var failedResponse = toBeClosed) {
       if (failedResponse.code() == HttpURLConnection.HTTP_UNAUTHORIZED) {
+        invalidTokenCallback.run();
```
damien-urruty-sonarsource, line 143: I feel propagating this callback in many classes is a bit odd. As discussed maybe we could rely on exceptions and have a "middle -man" catching and acting on them

backend/core/src/main/java/org/sonarsource/sonarlint/core/usertoken/UserTokenService.java
```diff
@@ -45,4 +51,9 @@ public void revokeToken(RevokeTokenParams params, SonarLintCancelMonitor cancelM
       .revoke(params.getTokenName(), cancelMonitor);
   }
 
+  @EventListener
+  public void notifyClientAboutInvalidToken(InvalidTokenEvent event) {
+    client.invalidToken(new InvalidTokenParams(event.getConnectionId()));
```
damien-urruty-sonarsource, line 56: If there is a "connection manager" class as we discussed, maybe it can also take care of notifying the client directly

damien-urruty-sonarsource (commented):
A first round of comments

backend/core/src/main/java/org/sonarsource/sonarlint/core/ConnectionManager.java
```diff
@@ -0,0 +1,329 @@
...
+import org.sonarsource.sonarlint.core.sync.InvalidTokenEvent;
+import org.springframework.context.ApplicationEventPublisher;
+import org.springframework.context.event.EventListener;
+
+import static java.util.stream.Collectors.toList;
+import static org.apache.commons.lang.StringUtils.removeEnd;
+
+@Named
+@Singleton
+public class ConnectionManager {
```
damien-urruty-sonarsource, line 90: This `ConnectionManager` is a huge facade, and `ServerApiErrorHandlingWrapper` seems to be a huge facade too, with some overlap between the 2. It seems a bit complex to me

backend/core/src/main/java/org/sonarsource/sonarlint/core/ConnectionManager.java
```diff
@@ -0,0 +1,329 @@
...
+  }
+
+  private HttpClient getClientFor(Either<TokenDto, UsernamePasswordDto> credentials) {
+    return credentials.map(
+      tokenDto -> httpClientProvider.getHttpClientWithPreemptiveAuth(tokenDto.getToken()),
+      userPass -> httpClientProvider.getHttpClientWithPreemptiveAuth(userPass.getUsername(), userPass.getPassword()));
+  }
+
+  private void publishWrongTokenEvent(@Nullable String connectionId) {
+    eventPublisher.publishEvent(new InvalidTokenEvent(connectionId));
```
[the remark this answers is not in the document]
kirill-knize-sonarsource: Done, thanks!

backend/core/src/main/java/org/sonarsource/sonarlint/core/ConnectionManager.java
kirill-knize-sonarsource: The functionality to call APIs moved from `ConnectionManager`, so now its task is to initialize the `ServerApiErrorHandlingWrapper` and provide the notification callback.

backend/core/src/main/java/org/sonarsource/sonarlint/core/ServerApiProvider.java
```diff
@@ -67,12 +70,12 @@ public Optional<ServerApi> getServerApi(String connectionId) {
       LOG.debug("Connection '{}' is gone", connectionId);
       return Optional.empty();
     }
-    return Optional.of(new ServerApi(params.get(), awareHttpClientProvider.getHttpClient(connectionId)));
+    return Optional.of(new ServerApi(params.get(), awareHttpClientProvider.getHttpClient(connectionId), () -> publishWrongTokenEvent(connectionId), this::publishWrongTokenEvent));
```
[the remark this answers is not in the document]
kirill-knize-sonarsource: Design changed.

backend/core/src/main/java/org/sonarsource/sonarlint/core/usertoken/UserTokenService.java
kirill-knize-sonarsource: Done.

backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/ServerApiErrorHandlingWrapper.java
```diff
@@ -0,0 +1,260 @@
...
+    return callWithErrorHandleAndRethrow(() -> serverApi.system().getStatus(cancelMonitor));
+  }
+
+  public Optional<ServerRule> tryFetchRule(String ruleKey, SonarLintCancelMonitor cancelMonitor) {
+    return callWithErrorHandleAndRethrow(() -> serverApi.rules().getRule(ruleKey, cancelMonitor));
+  }
+
+  public Optional<IssueApi.ServerIssueDetails> tryFetchIssue(String issueKey, String projectKey, String branch, @Nullable String pullRequest,
+    SonarLintCancelMonitor cancelMonitor) {
+    return callWithErrorHandleAndRethrow(() -> serverApi.issue().fetchServerIssue(issueKey, projectKey, branch, pullRequest, cancelMonitor));
```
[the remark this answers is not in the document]
kirill-knize-sonarsource: It works in the sense that it is a transparent wrapper that will notify the client in case a wrapped call throws a corresponding exception. If the wrapped call doesn't throw - the wrapper won't do anything.

backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/ServerApiErrorHandlingWrapper.java
```diff
@@ -0,0 +1,260 @@
...
+import org.sonarsource.sonarlint.core.serverapi.proto.sonarqube.ws.Issues;
+import org.sonarsource.sonarlint.core.serverapi.push.SonarServerEvent;
+import org.sonarsource.sonarlint.core.serverapi.qualityprofile.QualityProfile;
+import org.sonarsource.sonarlint.core.serverapi.rules.ServerActiveRule;
+import org.sonarsource.sonarlint.core.serverapi.rules.ServerRule;
+import org.sonarsource.sonarlint.core.serverapi.stream.EventStream;
+import org.sonarsource.sonarlint.core.serverapi.system.ServerStatusInfo;
+import org.sonarsource.sonarlint.core.serverapi.system.ValidationResult;
+
+public class ServerApiErrorHandlingWrapper {
```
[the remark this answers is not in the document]
kirill-knize-sonarsource: I came up with this because it's an approach where we centralize all the API calls and handle certain situations like wrong tokens or network errors in a single place. And introducing new API calls will naturally follow the same pattern. Even if not - if someone forgets to put the new call into the wrapper - in the worst-case scenario, we will get an unhandled error for this call as it exists today.
The suggestion above to change API behaviors method per method is pretty much as error-prone as this - you can easily forget to do it. But method per method changes will be a less visible pattern than a centralized wrapper, and I think it's actually easier to forget to follow this idea in the future.

Let's discuss our options. I created a meeting.

API_CHANGES.md
```diff
@@ -5,6 +5,8 @@
 * Signature of `org.sonarsource.sonarlint.core.rpc.protocol.backend.file.DidUpdateFileSystemParams#DidUpdateFileSystemParams` was changed
   * Parameter `addedOrChangedFiles` was split into `addedFiles` and `changedFiles`
 * Removed parameter `branch` and `pullRequest` from `org.sonarsource.sonarlint.core.rpc.protocol.client.issue.IssueDetailsDto` as it should not be used anymore by the client.
+* Add new client method `org.sonarsource.sonarlint.core.rpc.client.SonarLintRpcClientDelegate#invalidToken`
+  * It is called when the token is invalid and the client should notify the user about it
```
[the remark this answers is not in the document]
kirill-knize-sonarsource: We haven't discussed in depth the scenarios you described. This feature is more about mitigating the sync errors that lead to an inconsistent state of the app in a way that is not obvious to the user.
Situations like wrong token or network error during the wizard manipulations are not in the scope, and I think notifications to clients should be disregarded or even not sent in the first place. Unfortunately, on the CORE side, we don't track if the Web API call is performed by the client request or because of the CORE internal need. So, not sending notifications is quite tricky.
Let's discus it.

PR #1159, opened 2024-11-12
SLCORE-1036 Remove duplicated dependency from ITs pom file

kirill-knize-sonarsource:


kirill-knize-sonarsource:
The same dependency is declared in line 14 of the same pom file.

damien-urruty-sonarsource (approved):

PR #1162, opened 2024-11-15
SLCORE-1026 SonarLint is aligned with the custom/new severity for issues

nquinquenel:
[SLCORE-1026](https://sonarsource.atlassian.net/browse/SLCORE-1026)

[SLCORE-1026]: https://sonarsource.atlassian.net/browse/SLCORE-1026?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/SoftwareQuality.java
+++ b/backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/SoftwareQuality.java
@@ -22,6 +22,5 @@ package org.sonarsource.sonarlint.core.commons;
 public enum SoftwareQuality {
   MAINTAINABILITY,
   RELIABILITY,
-  SECURITY;
-
+  SECURITY
 }
--- a/backend/core/src/main/java/org/sonarsource/sonarlint/core/issue/IssueService.java
+++ b/backend/core/src/main/java/org/sonarsource/sonarlint/core/issue/IssueService.java
@@ -20,6 +20,8 @@
 package org.sonarsource.sonarlint.core.issue;
 
 import java.nio.file.Path;
+import java.util.Collections;
+import java.util.EnumMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -35,6 +37,7 @@ import org.eclipse.lsp4j.jsonrpc.ResponseErrorException;
 import org.eclipse.lsp4j.jsonrpc.messages.ResponseError;
 import org.sonarsource.sonarlint.core.ServerApiProvider;
 import org.sonarsource.sonarlint.core.commons.Binding;
+import org.sonarsource.sonarlint.core.commons.ImpactSeverity;
 import org.sonarsource.sonarlint.core.commons.LocalOnlyIssue;
 import org.sonarsource.sonarlint.core.commons.NewCodeDefinition;
 import org.sonarsource.sonarlint.core.commons.Transition;
@@ -418,7 +421,7 @@ public class IssueService {
     var resolved = event.getResolved();
     var userSeverity = event.getUserSeverity();
     var userType = event.getUserType();
-    var impactedIssueKeys = Set.copyOf(event.getImpactedIssues().stream().map(IssueChangedEvent.Issue::getIssueKey).collect(Collectors.toSet()));
+    var impactedIssueKeys = event.getImpactedIssues().stream().map(IssueChangedEvent.Issue::getIssueKey).collect(Collectors.toSet());
     if (resolved != null) {
       UnaryOperator<RaisedIssueDto> issueUpdater = it -> it.builder().withResolution(resolved).buildIssue();
       updatedIssue = updateIssue(updatedIssue, impactedIssueKeys, issueUpdater);
@@ -441,12 +444,24 @@ public class IssueService {
             )
           ).collect(Collectors.toList());
         UnaryOperator<RaisedIssueDto> issueUpdater =
-          it -> it.builder().withMQRModeDetails(it.getCleanCodeAttribute(), impacts).buildIssue();
+          it -> it.builder().withMQRModeDetails(
+            it.getCleanCodeAttribute(),
+            mergeImpacts(it.getSeverityMode().getRight().getImpacts(), impacts)
+          ).buildIssue();
         updatedIssue = updateIssue(updatedIssue, impactedIssueKeys, issueUpdater);
       }
 
     }
     return updatedIssue;
+  }
+
+  private static List<ImpactDto> mergeImpacts(List<ImpactDto> currentImpacts, List<ImpactDto> overriddenImpacts) {
+    for (var impact : overriddenImpacts) {
+      currentImpacts.removeIf(i -> i.getSoftwareQuality().equals(impact.getSoftwareQuality()));
+      currentImpacts.add(new ImpactDto(impact.getSoftwareQuality(), impact.getImpactSeverity()));
+    }
+
+    return currentImpacts;
   }
 
   private static RaisedIssueDto updateIssue(RaisedIssueDto issue, Set<String> impactedIssueKeys, UnaryOperator<RaisedIssueDto> issueUpdater) {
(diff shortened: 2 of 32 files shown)


backend/core/src/main/java/org/sonarsource/sonarlint/core/issue/IssueService.java
```diff
@@ -416,7 +418,7 @@ public static RaisedIssueDto raisedIssueUpdater(RaisedIssueDto previouslyRaisedI
     var resolved = event.getResolved();
     var userSeverity = event.getUserSeverity();
     var userType = event.getUserType();
-    var impactedIssueKeys = Set.copyOf(event.getImpactedIssueKeys());
+    var impactedIssueKeys = Set.copyOf(event.getImpactedIssues().stream().map(IssueChangedEvent.Issue::getIssueKey).collect(Collectors.toSet()));
```
damien-urruty-sonarsource, line 421: You probably don't need a copy now that you are using a stream and collect

backend/core/src/main/java/org/sonarsource/sonarlint/core/issue/IssueService.java
```diff
@@ -450,18 +459,22 @@ private static RaisedIssueDto updateIssue(RaisedIssueDto issue, Set<String> impa
...
+        storedIssue.setType(userType);
       }
       var resolved = event.getResolved();
       if (resolved != null) {
-        issue.setResolved(resolved);
+        storedIssue.setResolved(resolved);
+      }
+      var impacts = issue.getImpacts();
+      if (!impacts.isEmpty()) {
+        storedIssue.setImpacts(impacts);
```
damien-urruty-sonarsource, line 477: Is `getImpacts` in this event returning the overridden impacts or the full list? This [ticket](https://sonarsource.atlassian.net/browse/SONAR-23287) suggests it's only the "change", not sure what that means. On the other hand, this [ticket](https://sonarsource.atlassian.net/browse/SONAR-23429) mentions that the `issues/pull` endpoint returns the full list of impacts. This means you are calling `setImpacts` with a different set of impacts, but these assumptions need to be verified. Then I would suggest:
* merge impacts from the events with the ones already present in the issue
* add a test to cover this scenario

One last thing: how do you distinguish between no impacts in the event (because something else changed, like the resolution) and overridden impacts being removed? This if might prevent you from clearing the overrides in the storage

backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/push/parsing/IssueChangedEventParser.java
```diff
@@ -65,6 +86,8 @@ private boolean isInvalid() {
     private static class ChangedIssuePayload {
       private String issueKey;
       private String branchName;
+      @Nullable
+      private ImpactPayload[] impacts;
```
damien-urruty-sonarsource, line 90: I know it's not on you but I find it very strange that this field is in the `issues` object in the payload and not above

backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/storage/XodusServerIssueStore.java
```diff
@@ -192,8 +192,10 @@ private static ServerIssue adapt(Entity storedIssue) {
     var creationDate = (Instant) requireNonNull(storedIssue.getProperty(CREATION_DATE_PROPERTY_NAME));
     var userSeverity = (IssueSeverity) storedIssue.getProperty(USER_SEVERITY_PROPERTY_NAME);
     var type = (RuleType) requireNonNull(storedIssue.getProperty(TYPE_PROPERTY_NAME));
+    var impacts = readImpacts(storedIssue.getBlob(IMPACTS_BLOB_NAME));
```
damien-urruty-sonarsource, line 195: Be aware that for issues we have an incremental update with the pull mechanism. It means no impacts will be returned for all issues that are already stored. For such changes, we might want at some point to force a re-download of all issues. We had a similar case with [SLCORE-441](https://sonarsource.atlassian.net/browse/SLCORE-441)

backend/core/src/main/java/org/sonarsource/sonarlint/core/issue/IssueService.java
```diff
@@ -429,14 +431,21 @@ public static RaisedIssueDto raisedIssueUpdater(RaisedIssueDto previouslyRaisedI
...
+      if (!issue.getImpacts().isEmpty() && isMQRMode) {
+        var impacts = issue.getImpacts().entrySet().stream()
+          .map(impact ->
+            new ImpactDto(
+              SoftwareQuality.valueOf(impact.getKey().name()),
+              org.sonarsource.sonarlint.core.rpc.protocol.common.ImpactSeverity.valueOf(impact.getValue().name())
+            )
+          ).collect(Collectors.toList());
+        UnaryOperator<RaisedIssueDto> issueUpdater =
+          it -> it.builder().withMQRModeDetails(it.getCleanCodeAttribute(), impacts).buildIssue();
```
damien-urruty-sonarsource, line 444: Same thing here, you might be replacing the full list of impacts with just the overridden ones from the event. You probably need to merge what comes from the event with the previously raised issue

backend/core/src/main/java/org/sonarsource/sonarlint/core/tracking/TaintVulnerabilityTrackingService.java
```diff
@@ -193,8 +193,9 @@ private void updateStorageAndNotifyClient(String connectionId, IssueChangedEvent
   }
 
   private List<ServerTaintIssue> updateTaintIssues(String connectionId, String projectKey, @Nullable org.sonarsource.sonarlint.core.commons.IssueSeverity userSeverity,
-    @Nullable org.sonarsource.sonarlint.core.commons.RuleType userType, @Nullable Boolean resolved, List<String> issueKeys) {
+    @Nullable org.sonarsource.sonarlint.core.commons.RuleType userType, @Nullable Boolean resolved, List<IssueChangedEvent.Issue> issues) {
     var findingsStorage = storageService.connection(connectionId).project(projectKey).findings();
+    var issueKeys = Set.copyOf(issues.stream().map(IssueChangedEvent.Issue::getIssueKey).collect(Collectors.toSet()));
```
damien-urruty-sonarsource, line 198: Same here with the copy

damien-urruty-sonarsource (commented):
It looks mostly good but I left some comments related to how you handle the event, this might need some more work

backend/core/src/main/java/org/sonarsource/sonarlint/core/issue/IssueService.java
nquinquenel: After confirmation, once a rule has been modified, the impact field will always be present, even when the impact is modified back to the original.

backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/push/parsing/IssueChangedEventParser.java
nquinquenel: Multiple issues may have different impacts, so it cannot be above. Whereas changing the type would modify every issues for example.

sophio-japharidze-sonarsource (approved):

backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/push/parsing/IssueChangedEventParser.java
```diff
@@ -37,19 +43,37 @@ public class IssueChangedEventParser implements EventParser<IssueChangedEvent> {
 
   @Override
   public Optional<IssueChangedEvent> parse(String jsonData) {
+    LOG.info("'IssueChangedEvent' jsonData: " + jsonData);
     var payload = gson.fromJson(jsonData, IssueChangedEventPayload.class);
+    LOG.info("'IssueChangedEvent' payload: " + payload);
+    LOG.info("'IssueChangedEvent' payload issues toString: " + payload.issues.toString());
```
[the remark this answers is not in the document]
nquinquenel: Ah yes, leftovers 👍🏻

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/SoftwareQuality.java
```diff
@@ -22,6 +22,5 @@
 public enum SoftwareQuality {
   MAINTAINABILITY,
   RELIABILITY,
-  SECURITY;
-
+  SECURITY
```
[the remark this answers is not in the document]
nquinquenel: the `;` is unnecessary

PR #1179, opened 2024-12-03
SLCORE-1064 Fix the loading of sonar-text analyzer in Connected Mode

nquinquenel:
[SLCORE-1064](https://sonarsource.atlassian.net/browse/SLCORE-1064)

[SLCORE-1064]: https://sonarsource.atlassian.net/browse/SLCORE-1064?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/core/src/main/java/org/sonarsource/sonarlint/core/plugin/PluginsService.java
+++ b/backend/core/src/main/java/org/sonarsource/sonarlint/core/plugin/PluginsService.java
@@ -179,7 +179,7 @@ public class PluginsService {
       return false;
     }
     // when storage is not present, assume that secrets are not supported by server
-    return connection.getKind() != ConnectionKind.SONARCLOUD && storageService.connection(connectionId).serverInfo().read()
+    return connection.getKind() == ConnectionKind.SONARCLOUD || storageService.connection(connectionId).serverInfo().read()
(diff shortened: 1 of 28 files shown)


sophio-japharidze-sonarsource (approved):

backend/core/src/main/java/org/sonarsource/sonarlint/core/plugin/PluginsService.java
```diff
@@ -179,7 +179,7 @@ public boolean supportsCustomSecrets(String connectionId) {
       return false;
     }
     // when storage is not present, assume that secrets are not supported by server
-    return connection.getKind() != ConnectionKind.SONARCLOUD && storageService.connection(connectionId).serverInfo().read()
+    return connection.getKind() == ConnectionKind.SONARCLOUD || storageService.connection(connectionId).serverInfo().read()
```
damien-urruty-sonarsource, line 182: It doesn't have to be done for the release but I think a test would help. It was already missing because we did not discover the bug in the first place, now that we are fixing it, I think we should add one

PR #1187, opened 2024-12-10
SLCORE-1032 Wrong token error during sync - [New design]

kirill-knize-sonarsource:
[SLCORE-1032](https://sonarsource.atlassian.net/browse/SLCORE-1032)

[SLCORE-1032]: https://sonarsource.atlassian.net/browse/SLCORE-1032?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/core/src/main/java/org/sonarsource/sonarlint/core/connection/ConnectionManager.java
```diff
@@ -0,0 +1,42 @@
...
+  ServerConnection getConnectionOrThrow(String connectionId);
+  /**
+   *   Having dedicated TransientConnection class makes sense only if we handle the connection errors from there.
+   *   Which brings up the problem of managing global state for notifications because we don't know the connection ID. <br/><br/>
+   *   On other hand providing ServerApis directly, all Web API calls from transient ServerApi are not protected by checks for connection state.
+   *   So we still can spam server with unprotected requests.
+   *   It's not a big problem because we don't use such requests during scheduled sync.
+   *   They are mostly related to setting up the connection or other user-triggered actions.
+   */
+  ServerApi getTransientConnection(String token, @Nullable String organization, String baseUrl);
```
damien-urruty-sonarsource, line 38: I wonder if this method should be located in `ConnectionManager.` As it's "transient" and by definition we won't keep state for it, and it's not (yet) a connection. I think it would simplify a bit the current `ServerApiProvider` which has a lot of entrypoints

backend/core/src/main/java/org/sonarsource/sonarlint/core/connection/ServerConnection.java
```diff
@@ -0,0 +1,103 @@
...
+    if (state != ConnectionState.INVALID_CREDENTIALS && state != ConnectionState.MISSING_PERMISSION) {
+      return false;
+    }
+    if (lastNotificationTime == null) {
+      return true;
+    }
+    return lastNotificationTime.plus(WRONG_TOKEN_NOTIFICATION_INTERVAL).isBefore(Instant.now());
+  }
+
+  private void notifyClientAboutWrongTokenIfNeeded() {
```
damien-urruty-sonarsource, line 97: I wonder if it's the responsibility of the `ServerConnection` to notify to the client. I would find it more logical to have it in the `ConnectionManager`.

backend/core/src/main/java/org/sonarsource/sonarlint/core/embedded/server/ShowHotspotRequestHandler.java
```diff
@@ -105,12 +105,7 @@ private void showHotspotForScope(String connectionId, String configurationScopeI
   }
 
   private Optional<ServerHotspotDetails> tryFetchHotspot(String connectionId, String hotspotKey, SonarLintCancelMonitor cancelMonitor) {
-    var serverApi = serverApiProvider.getServerApi(connectionId);
-    if (serverApi.isEmpty()) {
-      // should not happen since we found the connection just before, improve the design ?
-      return Optional.empty();
-    }
-    return serverApi.get().hotspot().fetch(hotspotKey, cancelMonitor);
+    return serverApiProvider.withValidConnectionFlatMapOptionalAndReturn(connectionId,api -> api.hotspot().fetch(hotspotKey, cancelMonitor));
```
damien-urruty-sonarsource, line 108: The name of this method is a bit long. In fact this makes me wonder if `Api` classes should return Optionals at all, maybe they should instead let the errors bubble up?

backend/core/src/main/java/org/sonarsource/sonarlint/core/ServerApiProvider.java
```diff
@@ -137,4 +145,44 @@ private HttpClient getClientFor(EndpointParams params, Either<TokenDto, Username
       userPass -> httpClientProvider.getHttpClientWithPreemptiveAuth(userPass.getUsername(), userPass.getPassword()));
   }
 
+  @Override
+  public ServerConnection getConnectionOrThrow(String connectionId) {
```
damien-urruty-sonarsource, line 149: As shared earlier, I find the surface of methods quite large, so for people not familiar it might be complex to know which one to use. Do you think there is a way to simplify? Maybe we should at least document the different usages

backend/core/src/main/java/org/sonarsource/sonarlint/core/ServerApiProvider.java
```diff
@@ -137,4 +145,44 @@ private HttpClient getClientFor(EndpointParams params, Either<TokenDto, Username
...
+
+  @Override
+  public <T> Optional<T> withValidConnectionFlatMapOptionalAndReturn(String connectionId, Function<ServerApi, Optional<T>> serverApiConsumer) {
+    return getValidConnection(connectionId).map(connection -> connection.withClientApiAndReturn(serverApiConsumer)).flatMap(Function.identity());
+  }
+
+  @VisibleForTesting
+  public Optional<ServerConnection> getValidConnection(String connectionId) {
+    return tryGetConnection(connectionId).filter(ServerConnection::isValid);
+  }
```
damien-urruty-sonarsource, line 187: This is used only in a single test, couldn't we get rid of it and use another method?

backend/core/src/main/java/org/sonarsource/sonarlint/core/ServerApiProvider.java
```diff
@@ -137,4 +145,44 @@ private HttpClient getClientFor(EndpointParams params, Either<TokenDto, Username
...
+
+  @Override
+  public <T> Optional<T> withValidConnectionFlatMapOptionalAndReturn(String connectionId, Function<ServerApi, Optional<T>> serverApiConsumer) {
+    return getValidConnection(connectionId).map(connection -> connection.withClientApiAndReturn(serverApiConsumer)).flatMap(Function.identity());
+  }
+
+  @VisibleForTesting
+  public Optional<ServerConnection> getValidConnection(String connectionId) {
+    return tryGetConnection(connectionId).filter(ServerConnection::isValid);
```
damien-urruty-sonarsource, line 186: A last thing, should we log that the connection is invalid? Maybe too noisy?

PR #1216, opened 2025-01-22
SLCORE-716: Mend / Promote task to use Java 17

thahnen:
[SLCORE-716](https://sonarsource.atlassian.net/browse/SLCORE-716)

Because the Docker image was setting the Java version conditionally on Cirrus CI.

[SLCORE-716]: https://sonarsource.atlassian.net/browse/SLCORE-716?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -155,7 +155,7 @@ mend_scan_task:
     <<: *CONTAINER_DEFINITION
   env:
     WS_APIKEY: VAULT[development/kv/data/mend data.apikey]
-    JDK_VERSION: "11"
+    JDK_VERSION: "17"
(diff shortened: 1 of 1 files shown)


damien-urruty-sonarsource (approved):
LGTM. Was it needed to bump the version for the promote task too? Anyway, it doesn't hurt.
As discussed yesterday, if ou didn't notice the Mend task was failing, I would never have. After thinking a bit more about it, I think the promotion should fail if Mend fails, WDYT? We could remove line 166

PR #1217, opened 2025-01-22
SLCORE-716 Fix code smells

damien-urruty-sonarsource:


damien-urruty-sonarsource:
There are a bit more cleanups than the issues found by Next, because I tried the "Code cleanup" feature from IntelliJ. I double checked and it fixed a handful more smells, I think it doesn't hurt

thahnen (approved):
LGTM!

PR #1192, opened 2024-12-17
SLCORE-1086 Investigate git blame slowness

nquinquenel:
[SLCORE-1086](https://sonarsource.atlassian.net/browse/SLCORE-1086)

[SLCORE-1086]: https://sonarsource.atlassian.net/browse/SLCORE-1086?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1228, opened 2025-01-27
SLCORE-1133: Rely on native Git if possible for blaming

thahnen:
[SLCORE-1133](https://sonarsource.atlassian.net/browse/SLCORE-1133)

Cache the Git executable and leverage it if possible first for blaming. If there is an issue, fallback to the implementation based on `git-files-blame` / `JGit`.

Supersedes #1192 

[SLCORE-1133]: https://sonarsource.atlassian.net/browse/SLCORE-1133?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/BlameParser.java
+++ b/backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/BlameParser.java
@@ -23,21 +23,29 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.StringReader;
 import java.nio.file.Path;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.Date;
 import java.util.regex.Pattern;
 import org.sonar.scm.git.blame.BlameResult;
 import org.sonarsource.sonarlint.core.commons.SonarLintBlameResult;
 
 public class BlameParser {
+
+  private static final String FILENAME = "filename ";
+  private static final String AUTHOR_MAIL = "author-mail ";
+  private static final String COMMITTER_TIME = "committer-time ";
+  // if this text change between different git versions it will break the implementation
+  private static final String NOT_COMMITTED = "<not.committed.yet>";
   private BlameParser() {
     // Utility class
   }
-  
+
   public static SonarLintBlameResult parseBlameOutput(String blameOutput, String currentFilePath, Path projectBaseDir) throws IOException {
     var blameResult = new BlameResult();
     var currentFileBlame = new BlameResult.FileBlame(currentFilePath, countCommitterTimeOccurrences(blameOutput));
     blameResult.getFileBlameByPath().put(currentFilePath, currentFileBlame);
-    var fileSections = blameOutput.split("filename ");
+    var fileSections = blameOutput.split(FILENAME);
     var currentLineNumber = 0;
 
     for (var fileSection : fileSections) {
@@ -47,11 +55,14 @@ public class BlameParser {
 
       try (var reader = new BufferedReader(new StringReader(fileSection))) {
         String line;
-
         while ((line = reader.readLine()) != null) {
-          if (line.startsWith("author-time ")) {
-            var epochSeconds = Long.parseLong(line.substring(12));
-            var commitDate = new Date(epochSeconds * 1000);
+          if (line.startsWith(AUTHOR_MAIL)) {
+            var authorEmail = line.substring(AUTHOR_MAIL.length());
+            currentFileBlame.getAuthorEmails()[currentLineNumber] = authorEmail;
+          }
+          if (line.startsWith(COMMITTER_TIME) && !currentFileBlame.getAuthorEmails()[currentLineNumber].equals(NOT_COMMITTED)) {
+            var committerTime = line.substring(COMMITTER_TIME.length());
+            var commitDate = Date.from(Instant.ofEpochSecond(Long.parseLong(committerTime)).truncatedTo(ChronoUnit.SECONDS));
             currentFileBlame.getCommitDates()[currentLineNumber] = commitDate;
           }
         }
@@ -63,7 +74,7 @@ public class BlameParser {
   }
 
   public static int countCommitterTimeOccurrences(String blameOutput) {
-    var pattern = Pattern.compile("^committer-time ", Pattern.MULTILINE);
+    var pattern = Pattern.compile("^" + COMMITTER_TIME, Pattern.MULTILINE);
     var matcher = pattern.matcher(blameOutput);
     var count = 0;
     while (matcher.find()) {
--- a/backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/BlameParser.java
+++ b/backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/BlameParser.java
@@ -1,0 +1,84 @@
+/*
+ * SonarLint Core - Commons
+ * Copyright (C) 2016-2025 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.sonarlint.core.commons.util.git;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.StringReader;
+import java.nio.file.Path;
+import java.text.ParseException;
+import java.util.Date;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.sonar.scm.git.blame.BlameResult;
+import org.sonarsource.sonarlint.core.commons.SonarLintBlameResult;
+
+public class BlameParser {
+
+  public static SonarLintBlameResult parseBlameOutput(String blameOutput, String currentFilePath, Path projectBaseDir) throws IOException, ParseException {
+    BlameResult blameResult = new BlameResult();
+    BlameResult.FileBlame currentFileBlame = new BlameResult.FileBlame(currentFilePath, countCommitterTimeOccurrences(blameOutput));
+    blameResult.getFileBlameByPath().put(currentFilePath, currentFileBlame);
+    String[] fileSections = blameOutput.split("filename ");
+    int currentLineNumber = 0;
+
+    for (String fileSection : fileSections) {
+      if (fileSection.isBlank()) {
+        continue;
+      }
+
+      try (BufferedReader reader = new BufferedReader(new StringReader(fileSection))) {
+        String line;
+
+        while ((line = reader.readLine()) != null) {
+          if (line.startsWith("author ")) {
+            String author = line.substring(7);
+            currentFileBlame.getAuthorEmails()[currentLineNumber] = author;
+          } else if (line.startsWith("author-time ")) {
+            long epochSeconds = Long.parseLong(line.substring(12));
+            Date commitDate = new Date(epochSeconds * 1000);
+            currentFileBlame.getCommitDates()[currentLineNumber] = commitDate;
+          } else if (line.startsWith("commit ")) {
+            String commitHash = line.substring(7);
+            currentFileBlame.getCommitHashes()[currentLineNumber] = commitHash;
+          }
+        }
+      }
+      currentLineNumber++;
+    }
+
+    return new SonarLintBlameResult(blameResult, projectBaseDir);
+  }
+
+  public static int countCommitterTimeOccurrences(String blameOutput) {
+    Pattern pattern = Pattern.compile("^committer-time ", Pattern.MULTILINE);
+    Matcher matcher = pattern.matcher(blameOutput);
+    int count = 0;
+    while (matcher.find()) {
+      count++;
+    }
+    return count;
+  }
+
+  private static int countLines(String blameOutput) {
+    return blameOutput.split("\n").length;
+  }
+
+}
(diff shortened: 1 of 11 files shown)


backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/GitUtils.java
```diff
@@ -84,6 +98,21 @@ public static List<URI> getVSCChangedFiles(@Nullable Path baseDir) {
...
 
+  public static SonarLintBlameResult getBlameResult(Path projectBaseDir, Set<Path> projectBaseRelativeFilePaths, @Nullable UnaryOperator<String> fileContentProvider) {
+    return getBlameResult(projectBaseDir, projectBaseRelativeFilePaths, fileContentProvider, GitUtils::checkIfEnabled);
+  }
+
+  public static SonarLintBlameResult getBlameResult(Path projectBaseDir, Set<Path> projectBaseRelativeFilePaths, @Nullable UnaryOperator<String> fileContentProvider,
+    Predicate<Path> isEnabled) {
+    if (isEnabled.test(projectBaseDir)) {
+      LOG.debug("Using native git blame");
+      return blameFromNativeCommand(projectBaseDir, projectBaseRelativeFilePaths);
```
thahnen, line 109: If the Git command throws an error instead of failing we should then fall back to JGit. This is currently not the case here (anymore).

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/BlameParser.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.regex.Pattern;
+import org.sonar.scm.git.blame.BlameResult;
+import org.sonarsource.sonarlint.core.commons.SonarLintBlameResult;
+
+public class BlameParser {
+
+  private static final String FILENAME = "filename ";
+  private static final String AUTHOR_MAIL = "author-mail ";
+  private static final String COMMITTER_TIME = "committer-time ";
+  // if this text change between different git versions it will break the implementation
```
thahnen, line 38: Git is always trying to preserve compatibility. I guess this will never change, but glad you added the comment here for people who have to come back to this later.

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/GitUtils.java
kirill-knize-sonarsource: Yes, but normally it should not because we test if git is installed and correct version with `checkIfEnabled`. We run native git and parse version output. It looks like an adequate check. So we assume blame command will also work.

One case when it won't - if there's no git repo, but in this case falling back to jgit also won't help.

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/GitUtils.java
```diff
@@ -110,6 +139,110 @@ public static SonarLintBlameResult blameWithFilesGitCommand(Path projectBaseDir,
     }
   }
 
+  /**
+   * Get the native Git executable by checking for the version of both `git` and `git.exe`. We cache this information
+   * to run these expensive processes more than once (or twice in case of Windows).
+   */
+  private static String getNativeGitExecutable() {
+    if (!checkedForNativeGitExecutable) {
+      for (var executable : List.of("git", "git.exe")) {
```
nquinquenel, line 148: `git.exe` should be checked only if OS is Windows

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/GitUtils.java
```diff
@@ -110,6 +139,110 @@ public static SonarLintBlameResult blameWithFilesGitCommand(Path projectBaseDir,
...
+          var process = new ProcessBuilder(executable, "--version").start();
+          var exitCode = process.waitFor();
+          if (exitCode == 0) {
+            nativeGitExecutable = executable;
+            break;
+          }
+        } catch (IOException e) {
+          // Do log the exception stacktrace here as it might not be due to the command not being found but rather something like the
+          // Windows Execution Policy preventing it or something else. In this case this is extremely useful for debugging!
+          LOG.debug("Checking for native Git executable '" + executable + "' failed with an exception", e);
```
nquinquenel, line 159:
```suggestion
          LOG.debug("Checking for native Git executable '" + executable + "' failed", e);
```

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/GitUtils.java
nquinquenel, line 148: On Windows it will locate a git.exe inside the current directory, which may be insecure. I think we could do like SQ:S and use where.exe and the $PATH to look for git.exe

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/GitUtils.java
```diff
@@ -110,6 +139,110 @@ public static SonarLintBlameResult blameWithFilesGitCommand(Path projectBaseDir,
...
+    return nativeGitExecutable;
+  }
+
+  private static String executeGitCommand(Path workingDir, String... command) throws IOException {
+    var processBuilder = new ProcessBuilder(command)
+      .directory(workingDir.toFile())
+      .redirectErrorStream(true);
+    var process = processBuilder.start();
+    var output = new StringBuilder();
+    try (var reader = new BufferedReader(new InputStreamReader(process.getInputStream(), StandardCharsets.UTF_8))) {
```
nquinquenel, line 175: Might be a detail but we use a BufferedReader to parse the output to a String, and then right after (in the parent method) we re-use a BufferedReader to parse the String. I think the [ProcessWrapper](https://github.com/SonarSource/sonar-enterprise/blob/master/sonar-scanner-engine/src/main/java/org/sonar/scm/git/ProcessWrapperFactory.java) from SQ:S is elegant

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/GitUtils.java
```diff
@@ -110,6 +139,110 @@ public static SonarLintBlameResult blameWithFilesGitCommand(Path projectBaseDir,
...
+   * Get the native Git executable by checking for the version of both `git` and `git.exe`. We cache this information
+   * to run these expensive processes more than once (or twice in case of Windows).
+   */
+  private static String getNativeGitExecutable() {
+    if (!checkedForNativeGitExecutable) {
+      for (var executable : List.of("git", "git.exe")) {
+        try {
+          var process = new ProcessBuilder(executable, "--version").start();
+          var exitCode = process.waitFor();
+          if (exitCode == 0) {
```
nquinquenel, line 152: If we want to be picky, we should check the output makes sense (i.e., starting by `git version`)

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/GitUtils.java
```diff
@@ -110,6 +139,110 @@ public static SonarLintBlameResult blameWithFilesGitCommand(Path projectBaseDir,
...
+  private static boolean isCompatibleGitVersion(String gitVersionCommandOutput) {
+    // Due to the danger of argument injection on git blame the use of `--end-of-options` flag is necessary
+    // The flag is available only on git versions >= 2.24.0
+    String gitVersion = whitespaceRegex
+      .splitAsStream(gitVersionCommandOutput)
+      .skip(2)
+      .findFirst()
+      .orElse("");
+
+    String formattedGitVersion = formatGitSemanticVersion(gitVersion);
```
nquinquenel, line 235:
```suggestion
    var formattedGitVersion = formatGitSemanticVersion(gitVersion);
```

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/GitUtils.java
```diff
@@ -110,6 +139,110 @@ public static SonarLintBlameResult blameWithFilesGitCommand(Path projectBaseDir,
...
+      return output.contains("git version") && isCompatibleGitVersion(output);
+    } catch (IOException e) {
+      return false;
+    }
+  }
+
+  private static boolean isCompatibleGitVersion(String gitVersionCommandOutput) {
+    // Due to the danger of argument injection on git blame the use of `--end-of-options` flag is necessary
+    // The flag is available only on git versions >= 2.24.0
+    String gitVersion = whitespaceRegex
```
nquinquenel, line 229:
```suggestion
    var gitVersion = whitespaceRegex
```

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/GitUtils.java
thahnen: This is a fair point! I was more concerned that `git blame` itself fails maybe due to incorrect inputs or so.

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/BlameParser.java
```diff
@@ -0,0 +1,85 @@
...
+
+    for (var fileSection : fileSections) {
+      if (fileSection.isBlank()) {
+        continue;
+      }
+
+      try (var reader = new BufferedReader(new StringReader(fileSection))) {
+        String line;
+        while ((line = reader.readLine()) != null) {
+          if (line.startsWith(AUTHOR_MAIL)) {
```
nquinquenel, line 59: Do we even care about author mail?

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/BlameParser.java
```diff
@@ -0,0 +1,85 @@
...
+          }
+        }
+      }
+      currentLineNumber++;
+    }
+
+    return new SonarLintBlameResult(blameResult, projectBaseDir);
+  }
+
+  public static int countCommitterTimeOccurrences(String blameOutput) {
```
nquinquenel, line 76: What's the point of counting this?

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/GitUtils.java
```diff
@@ -84,6 +98,21 @@ public static List<URI> getVSCChangedFiles(@Nullable Path baseDir) {
...
+    return getBlameResult(projectBaseDir, projectBaseRelativeFilePaths, fileContentProvider, GitUtils::checkIfEnabled);
+  }
+
+  public static SonarLintBlameResult getBlameResult(Path projectBaseDir, Set<Path> projectBaseRelativeFilePaths, @Nullable UnaryOperator<String> fileContentProvider,
+    Predicate<Path> isEnabled) {
+    if (isEnabled.test(projectBaseDir)) {
+      LOG.debug("Using native git blame");
+      return blameFromNativeCommand(projectBaseDir, projectBaseRelativeFilePaths);
+    } else {
+      LOG.debug("Falling back to jgit git blame");
```
nquinquenel, line 111:
```suggestion
      LOG.debug("Falling back to JGit git blame");
```

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/BlameParser.java
kirill-knize-sonarsource: Yes we do because committer time is provided even for not committed lines, so the author email is the way to understand if line was committed. If not it will be `<not.committed.yet>`.

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/BlameParser.java
kirill-knize-sonarsource: To initialize the BlameResult underlying arrays with correct size.

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/GitUtils.java
kirill-knize-sonarsource: That's a very good catch. I will address it today. Thanks!

kirill-knize-sonarsource: True.

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/GitUtils.java
kirill-knize-sonarsource: Borrowed the ProcessWrapper from SQ:S. Thanks.

kirill-knize-sonarsource:
Coverage is below 85% because of `locateGitOnWindows`. We don't measure coverage on windows unfortunately. But code is being executed on Windows UTs.

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/GitUtils.java
kirill-knize-sonarsource: But it's we who provide the inputs. We fully in control. I don't see how it can be incorrect in normal scenario.

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/GitUtils.java
nquinquenel: I think the ProcessWrapper could be used here, I believe they do something similar on SQ:S. Because the output is generated via the BufferedReader, and then you re-create a BufferedReader later to parse the output. You could maybe do the logic directly here? Not sure if I'm clear but anyway, it's maybe just an optimization for later

kirill-knize-sonarsource: Done.

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/ProcessWrapperFactory.java
```diff
@@ -0,0 +1,121 @@
...
+        });
+
+        processInputStream(p.getErrorStream(), line -> {
+          if (!line.isBlank()) {
+            output.add(line);
+            LOG.debug(line);
+          }
+        });
+
+        int exit = p.waitFor();
```
nquinquenel, line 104:
```suggestion
        var exit = p.waitFor();
```

nquinquenel (approved):

PR #1230, opened 2025-01-27
SLCORE-1144 Fix SSF-699

nquinquenel:
[SLCORE-1144](https://sonarsource.atlassian.net/browse/SLCORE-1144)

[SLCORE-1144]: https://sonarsource.atlassian.net/browse/SLCORE-1144?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/commons/src/test/java/org/sonarsource/sonarlint/core/commons/StringUtilsTests.java
+++ b/backend/commons/src/test/java/org/sonarsource/sonarlint/core/commons/StringUtilsTests.java
@@ -20,17 +20,28 @@
 package org.sonarsource.sonarlint.core.commons;
 
 import org.junit.jupiter.api.Test;
-import org.sonarsource.sonarlint.core.commons.util.StringUtils;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.sonarsource.sonarlint.core.commons.util.StringUtils.pluralize;
+import static org.sonarsource.sonarlint.core.commons.util.StringUtils.sanitizeAgainstRTLO;
 
 class StringUtilsTests {
 
   @Test
   void should_pluralize_words() {
-    assertThat(StringUtils.pluralize(0, "word")).isEqualTo("0 word");
-    assertThat(StringUtils.pluralize(1, "word")).isEqualTo("1 word");
-    assertThat(StringUtils.pluralize(2, "word")).isEqualTo("2 words");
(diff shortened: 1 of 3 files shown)


backend/commons/src/test/java/org/sonarsource/sonarlint/core/commons/StringUtilsTests.java
```diff
@@ -20,17 +20,28 @@
...
+import static org.sonarsource.sonarlint.core.commons.util.StringUtils.sanitizeAgainstRTLO;
 
 class StringUtilsTests {
 
   @Test
   void should_pluralize_words() {
-    assertThat(StringUtils.pluralize(0, "word")).isEqualTo("0 word");
-    assertThat(StringUtils.pluralize(1, "word")).isEqualTo("1 word");
-    assertThat(StringUtils.pluralize(2, "word")).isEqualTo("2 words");
+    assertThat(pluralize(0, "word")).isEqualTo("0 word");
```
damien-urruty-sonarsource, line 32: Have you seen the post from JB? Apparently `0 words` should be plural :exploding_head:

damien-urruty-sonarsource (approved):
It would be interesting to check if there are other Unicode problematic characters. Also should the fix be applied to other inputs received by the embedded server?

nquinquenel:
> It would be interesting to check if there are other Unicode problematic characters. Also should the fix be applied to other inputs received by the embedded server?

There are many documentation ([OWASP](https://cheatsheetseries.owasp.org/cheatsheets/Input_Validation_Cheat_Sheet.html#validating-free-form-unicode-text), [unicode.org](https://unicode.org/reports/tr36/)), on this subject, and I feel like there is no easy and straightforward solution regarding Unicode validation. Depending on what kind of text we expect, we could allow or disallow many things; the risk is that you could have false positives that would break the AI-generated code. I'm not strongly opinionated, but I'd tend to go easy for now on the validation.

nquinquenel:
> Also should the fix be applied to other inputs received by the embedded server?

The only inputs received by the embedded server that are shown to the client are the fix suggestions and the branch names in some notification titles. I think the only attack vector is when showing and applying a fix suggestion.

PR #1247, opened 2025-02-14
SLCORE-1142 Improve Git blaming for the analysis

kirill-knize-sonarsource:


backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/NewCodeDefinition.java
```diff
@@ -62,6 +62,8 @@ static NewCodeDefinition withNumberOfDaysWithDate(int days, long thresholdDate)
     return new NewCodeNumberOfDaysWithDate(days, thresholdDate);
   }
 
+  long getThresholdDate();
```
damien-urruty-sonarsource, line 65: Why not returning an Instant from here? You can make the conversion later when you use it

damien-urruty-sonarsource: I think threshold date is maybe not precise enough. I am bad at naming, but what about something like "period start date"?

damien-urruty-sonarsource: Nitpick: I would move the method somewhere else, to not be in the middle of `with*` methods.

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/NewCodeDefinition.java
```diff
@@ -116,6 +118,11 @@ public boolean isSupported() {
       return true;
     }
 
+    @Override
+    public long getThresholdDate() {
```
damien-urruty-sonarsource, line 122: Shouldn't this method also be implemented in `NewCodeDefinitionWithDate`? Maybe it's the case.
EDIT: I see this abstract class is not used at all, maybe we could remove it?

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/GitUtils.java
```diff
@@ -189,6 +192,11 @@ public static SonarLintBlameResult blameFromNativeCommand(Path projectBaseDir, S
     throw new IllegalStateException("There is no native Git available");
   }
 
+  private static String getBlameHistoryWindow(long thresholdDate) {
+    var blameLimit = Instant.ofEpochMilli(thresholdDate);
```
damien-urruty-sonarsource, line 196: Related to what I said earlier, we create Instants that we convert to a timestamp, then back to an Instant

damien-urruty-sonarsource (approved):
Some minor comments, overall LGTM.
Maybe it would deserve some medium tests to make sure everything is plugged together

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/GitUtils.java
kirill-knize-sonarsource: Fixed.

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/NewCodeDefinition.java
kirill-knize-sonarsource: Fixed the problem with Instant and long, now returning Instant.
Regarding the method location - I'm not sure where to move it, so I left it here for now.

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/NewCodeDefinition.java
kirill-knize-sonarsource: This method is implemented in `NewCodeDefinitionWithDate`. Not sure what you mean exactly.
Also abstract class is holding the common state - `thresholdDate` for children: `NewCodeNumberOfDaysWithDate`, `NewCodePreviousVersion`, `NewCodeSpecificAnalysis`.
By removing the class we will have to re-define the field many times.

If you see how to improve the design, let's discuss or make a PR.
For now I merge as is.

PR #1248, opened 2025-02-14
SLCORE-1158 Expose a way to request a fix suggestion

damien-urruty-sonarsource:
[SLCORE-1158](https://sonarsource.atlassian.net/browse/SLCORE-1158)

[SLCORE-1158]: https://sonarsource.atlassian.net/browse/SLCORE-1158?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/core/src/main/java/org/sonarsource/sonarlint/core/remediation/aicodefix/AiCodeFixService.java
+++ b/backend/core/src/main/java/org/sonarsource/sonarlint/core/remediation/aicodefix/AiCodeFixService.java
@@ -1,0 +1,119 @@
+/*
+ * SonarLint Core - Implementation
+ * Copyright (C) 2016-2025 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.sonarlint.core.remediation.aicodefix;
+
+import java.util.UUID;
+import org.eclipse.lsp4j.jsonrpc.ResponseErrorException;
+import org.eclipse.lsp4j.jsonrpc.messages.ResponseError;
+import org.eclipse.lsp4j.jsonrpc.messages.ResponseErrorCode;
+import org.sonarsource.sonarlint.core.ConnectionManager;
+import org.sonarsource.sonarlint.core.commons.Binding;
+import org.sonarsource.sonarlint.core.commons.progress.SonarLintCancelMonitor;
+import org.sonarsource.sonarlint.core.fs.ClientFileSystemService;
+import org.sonarsource.sonarlint.core.repository.config.ConfigurationRepository;
+import org.sonarsource.sonarlint.core.repository.connection.ConnectionConfigurationRepository;
+import org.sonarsource.sonarlint.core.repository.connection.SonarCloudConnectionConfiguration;
+import org.sonarsource.sonarlint.core.repository.reporting.PreviouslyRaisedFindingsRepository;
+import org.sonarsource.sonarlint.core.repository.reporting.RaisedIssue;
+import org.sonarsource.sonarlint.core.rpc.protocol.backend.remediation.aicodefix.SuggestFixChangeDto;
+import org.sonarsource.sonarlint.core.rpc.protocol.backend.remediation.aicodefix.SuggestFixResponse;
+import org.sonarsource.sonarlint.core.serverapi.fixsuggestions.AiSuggestionRequestBodyDto;
+import org.sonarsource.sonarlint.core.serverapi.fixsuggestions.AiSuggestionResponseBodyDto;
+
+import static java.util.Objects.requireNonNull;
+import static org.sonarsource.sonarlint.core.rpc.protocol.SonarLintRpcErrorCode.CONFIG_SCOPE_NOT_BOUND;
+import static org.sonarsource.sonarlint.core.rpc.protocol.SonarLintRpcErrorCode.CONNECTION_KIND_NOT_SUPPORTED;
+import static org.sonarsource.sonarlint.core.rpc.protocol.SonarLintRpcErrorCode.CONNECTION_NOT_FOUND;
+import static org.sonarsource.sonarlint.core.rpc.protocol.SonarLintRpcErrorCode.FILE_NOT_FOUND;
+import static org.sonarsource.sonarlint.core.rpc.protocol.SonarLintRpcErrorCode.ISSUE_NOT_FOUND;
+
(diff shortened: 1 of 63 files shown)


backend/core/src/main/java/org/sonarsource/sonarlint/core/repository/reporting/PreviouslyRaisedFindingsRepository.java
```diff
@@ -84,4 +84,10 @@ public Optional<RaisedHotspotDto> getRaisedHotspotWithScopeAndId(String scopeId,
       .findFirst();
   }
 
+  public Optional<RaisedIssue> findRaisedIssueById(String configurationScopeId, UUID issueId) {
+    return Optional.ofNullable(previouslyRaisedIssuesByScopeId.get(configurationScopeId))
+      .flatMap(issuesByUri -> issuesByUri.entrySet().stream()
+        .flatMap(entry -> entry.getValue().stream().filter(issue -> issue.getId().equals(issueId)).findFirst().map(issue -> new RaisedIssue(entry.getKey(), issue)).stream())
+        .findFirst());
```
nquinquenel, line 91:
```suggestion
return Optional.ofNullable(previouslyRaisedIssuesByScopeId.get(configurationScopeId))
    .flatMap(issuesByUri -> issuesByUri.entrySet().stream()
        .flatMap(entry -> entry.getValue().stream()
            .filter(issue -> issue.getId().equals(issueId))
            .map(issue -> new RaisedIssue(entry.getKey(), issue)))
        .findFirst());
```
We could use better indentation for readability, and I think we could eliminate the first `.findFirst()` and `.stream()` for simplicity?

backend/core/src/main/java/org/sonarsource/sonarlint/core/remediation/aicodefix/AiCodeFixService.java
```diff
@@ -0,0 +1,119 @@
...
+
+  private static SuggestFixResponse adapt(AiSuggestionResponseBodyDto responseBodyDto) {
+    return new SuggestFixResponse(responseBodyDto.id(), responseBodyDto.explanation(),
+      responseBodyDto.changes().stream().map(change -> new SuggestFixChangeDto(change.startLine(), change.endLine(), change.newCode())).toList());
+  }
+
+  private SonarQubeCloudBinding ensureBoundToSonarQubeCloud(String configurationScopeId) {
+    var effectiveBinding = configurationRepository.getEffectiveBinding(configurationScopeId);
+    if (effectiveBinding.isEmpty()) {
+      throw new ResponseErrorException(new ResponseError(CONFIG_SCOPE_NOT_BOUND, "The provided configuration scope is not bound", configurationScopeId));
```
nquinquenel, line 90: If these messages are meant to be displayed on the UI by clients, I'm not sure we want to talk about `configuration scope`, it may be too abstract? But then I'm not sure we have a generic way for all clients implementation

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/remediation/aicodefix/SuggestFixChangeDto.java
```diff
@@ -0,0 +1,44 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.sonarlint.core.rpc.protocol.backend.remediation.aicodefix;
+
+public class SuggestFixChangeDto {
+  private final int startLine;
```
nquinquenel, line 23: In `ChangesDto` (open fix suggestion in IDE), we use `LineRangeDto` instead of startline/endline, WDYT in terms of consistency?

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/remediation/aicodefix/SuggestFixResponse.java
```diff
@@ -0,0 +1,47 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.sonarlint.core.rpc.protocol.backend.remediation.aicodefix;
+
+import java.util.List;
+import java.util.UUID;
+
+public class SuggestFixResponse {
```
nquinquenel, line 25: I'm afraid there might be some confusion with the existing `FixSuggestion*` and this `SuggestFix*`, do you think we could make the difference clearer?

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/remediation/aicodefix/SuggestFixResponse.java
```diff
@@ -0,0 +1,47 @@
...
+  private final UUID id;
```
nquinquenel, line 26: Here we have a `UUID id`, in FixSuggestionDto we have `String suggestionId`, could we be more consistent here as well?

backend/core/src/main/java/org/sonarsource/sonarlint/core/repository/reporting/PreviouslyRaisedFindingsRepository.java
```diff
@@ -84,4 +84,10 @@ public Optional<RaisedHotspotDto> getRaisedHotspotWithScopeAndId(String scopeId,
       .findFirst();
   }
 
+  public Optional<RaisedIssue> findRaisedIssueById(String configurationScopeId, UUID issueId) {
```
nquinquenel, line 87: Do we really have to look per config scope ID? Couldn't we just find the issue key from what is globally stored?

nquinquenel (commented):
Looks good, just have a few comments for discussion

backend/core/src/main/java/org/sonarsource/sonarlint/core/repository/reporting/PreviouslyRaisedFindingsRepository.java
damien-urruty-sonarsource: I think we can simplify yes, chances to have a UUID collision across different config scopes is close to zero!

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/remediation/aicodefix/SuggestFixChangeDto.java
damien-urruty-sonarsource: I didn't consider at all what was done for the "Open in IDE" feature and the corresponding code. The data types look mostly similar but there are small differences (e.g. `FileEditDto` has the `idePath`), so I prefer to keep them separate for now. For your specific point, I could reuse `LineRangeDto` but it's in the `client` package so I would need to move it, and I prefer not to introduce breaking changes at this point.

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/remediation/aicodefix/SuggestFixResponse.java
damien-urruty-sonarsource: This class is called *Response so it shows it's the consequence of a request. The other doesn't represent that it's used in the context of Open in IDE. I don't know, do you have any suggestion?

nquinquenel (approved):

PR #1252, opened 2025-02-19
SLCORE-1161 Expose if issue is fixable when rule is supported

damien-urruty-sonarsource:
[SLCORE-1161](https://sonarsource.atlassian.net/browse/SLCORE-1161)

[SLCORE-1161]: https://sonarsource.atlassian.net/browse/SLCORE-1161?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/ServerApiHelper.java
```diff
@@ -80,6 +80,14 @@ public HttpClient.Response get(String path, SonarLintCancelMonitor cancelMonitor
     return response;
   }
 
+  public HttpClient.Response apiGet(String path, SonarLintCancelMonitor cancelMonitor) {
```
nquinquenel, line 83: I know it's already like this in all the other methods, but I think it could be named `relativeUrl` and not `path` or `relativePath`

backend/core/src/main/java/org/sonarsource/sonarlint/core/remediation/aicodefix/AiCodeFixFeature.java
```diff
@@ -0,0 +1,35 @@
...
+
+public record AiCodeFixFeature(AiCodeFixSettings settings) {
+
+  public boolean isFixable(TrackedIssue issue) {
+    return settings.supportedRules().contains(issue.getRuleKey()) && issue.getTextRangeWithHash() != null;
+  }
+
+  public boolean isFixable(RaisedIssue issue) {
+    return settings.supportedRules().contains(issue.issueDto().getRuleKey()) && issue.issueDto().getTextRange() != null;
+  }
```
nquinquenel, line 34:
```suggestion
  }
  
```

damien-urruty-sonarsource:
@nquinquenel I think there is no supported taint rule at the moment, but in theory later it could. I will make the change

backend/core/src/main/java/org/sonarsource/sonarlint/core/remediation/aicodefix/AiCodeFixFeature.java
damien-urruty-sonarsource: Is it about leaving an empty line between the end of the method and the end of the class? I don't think there is any convention settled on this, and probably our formatting tool doesn't enforce anything on that. My personal practice is to not have an extra line here, but I can change that, it's not very important

backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/ServerApiHelper.java
damien-urruty-sonarsource: I think both work. You could see it as a relative URL or as an API path, I can see the two terminologies in use on the web. Same, no strong opinion here. As you said, it's the same in other methods, I suggest we try to align that in a separate effort

backend/core/src/main/java/org/sonarsource/sonarlint/core/remediation/aicodefix/AiCodeFixFeature.java
nquinquenel: Yes, it's my OCD speaking there. We don't have any convention on this in our projects, but as a rule of thumb, I try to stay consistent: if there's an empty line after the first class' bracket, I keep an empty line before the ending class' bracket, but it's not important 😅

damien-urruty-sonarsource: Is it better if I remove the first empty line at the beginning of the class?

nquinquenel: `Is it better if I remove the first empty line at the beginning of the class?`

We do both so no strong opinion, I believe in most Java code convention (Google/Oracle) there's no empty line

nquinquenel (approved):

PR #1253, opened 2025-02-20
SLCORE-1186 Fix array out bounds error for native git blame

kirill-knize-sonarsource:
[SLCORE-1186](https://sonarsource.atlassian.net/browse/SLCORE-1186)

[SLCORE-1186]: https://sonarsource.atlassian.net/browse/SLCORE-1186?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/BlameParser.java
+++ b/backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/BlameParser.java
@@ -29,9 +29,10 @@ import java.util.Date;
 import java.util.regex.Pattern;
 import org.sonar.scm.git.blame.BlameResult;
 import org.sonarsource.sonarlint.core.commons.SonarLintBlameResult;
+import org.sonarsource.sonarlint.core.commons.log.SonarLintLogger;
 
 public class BlameParser {
-
+  private static final SonarLintLogger LOG = SonarLintLogger.get();
   private static final String FILENAME = "filename ";
   private static final String AUTHOR_MAIL = "author-mail ";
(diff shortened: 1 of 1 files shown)


backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/BlameParser.java
```diff
@@ -73,8 +75,19 @@ public static SonarLintBlameResult parseBlameOutput(String blameOutput, String c
...
   }
 
-  public static int countCommitterTimeOccurrences(String blameOutput) {
-    var pattern = Pattern.compile("^" + COMMITTER_TIME, Pattern.MULTILINE);
+  private static boolean shouldSkipSection(String fileSection, boolean lineNumberIsOff) {
+    if (fileSection.isBlank()) {
+      return true;
+    }
+    if (lineNumberIsOff) {
+      LOG.warn("Number of blame output sections is higher than expected number of lines. Section content: {}", fileSection);
```
kirill-knize-sonarsource, line 83: We could manually create an exception here to log it, so we had a visibility of such events on Sentry. It could help to understand better what's going on.

nquinquenel (commented):
How can this happen? Are there lines at the end not starting with `filename `?

kirill-knize-sonarsource:
> How can this happen? Are there lines at the end not starting with `filename `?

Yes, it's not clear because the output I seen should have only one block per line. And `filename` should be present exactly once in every block. That's why I suggest to log exception to keep observability on sentry, to get data from the real cases.

PR #1259, opened 2025-02-27
Mute issue fix for VS

kirill-knize-sonarsource:


backend/core/src/main/java/org/sonarsource/sonarlint/core/issue/IssueService.java
```diff
@@ -138,10 +140,18 @@ public void changeStatus(String configurationScopeId, String issueKey, Resolutio
...
-        .flatMap(localOnlyIssueRepository::findByKey);
+      var localIssueOpt = asUUID(issueKey).flatMap(localOnlyIssueRepository::findByKey);
       if (localIssueOpt.isEmpty()) {
-        throw issueNotFoundException(issueKey);
+        // this happens in case if VS client trying to change issue of the issue for Roslyn analysed language
+        // since analysis was executed outside the backend on the client side we trust client to provide valid issue key and send telemetry
+        try {
+          serverConnection.withClientApi(serverApi -> serverApi.issue().changeStatus(issueKey, reviewStatus, cancelMonitor));
+          ServerFinding serverFinding = () -> "unknownRuleKey";
+          eventPublisher.publishEvent(new ServerIssueStatusChangedEvent(binding.connectionId(), binding.sonarProjectKey(), serverFinding));
```
kirill-knize-sonarsource, line 150: Not sure there's a value in such telemetry. Maybe we should omit it at all. I don't think it's reasonable to ask client to provide a ruleKey via RPC. It's a bit of a corner case.

vnaskos-sonar:
I've tested muting a Roslyn issue from SQVS on SonarQube Cloud with SLCore `10.16.0.80376` from this PR and is working 🚀

backend/core/src/main/java/org/sonarsource/sonarlint/core/issue/IssueService.java
```diff
@@ -138,10 +140,18 @@ public void changeStatus(String configurationScopeId, String issueKey, Resolutio
       projectServerIssueStore.updateIssueResolutionStatus(issueKey, isTaintIssue, true)
         .ifPresent(issue -> eventPublisher.publishEvent(new ServerIssueStatusChangedEvent(binding.connectionId(), binding.sonarProjectKey(), issue)));
     } else {
-      var localIssueOpt = asUUID(issueKey)
-        .flatMap(localOnlyIssueRepository::findByKey);
+      var localIssueOpt = asUUID(issueKey).flatMap(localOnlyIssueRepository::findByKey);
       if (localIssueOpt.isEmpty()) {
-        throw issueNotFoundException(issueKey);
+        // this happens in case if VS client trying to change issue of the issue for Roslyn analysed language
```
damien-urruty-sonarsource, line 145: > to change issue of the issue

to change the status of the issue?

backend/core/src/main/java/org/sonarsource/sonarlint/core/issue/IssueService.java
damien-urruty-sonarsource: Let's remove the publishing of the event then

damien-urruty-sonarsource (approved):
LGTM if you remove the event publishing

backend/core/src/main/java/org/sonarsource/sonarlint/core/issue/IssueService.java
kirill-knize-sonarsource: Done

PR #1260, opened 2025-02-27
SLCORE-1159 Fix NPE when storing enablement settings

damien-urruty-sonarsource:


nquinquenel (approved):
LGTM but wouldn't it be better not to allow passing `null` value from the cloud service?

damien-urruty-sonarsource:
@nquinquenel no strong opinion, but project enablement was not implemented by me, it was done a while ago by the Code Generation squad and I didn't touch anything on this endpoint. So we would need to discuss with them. For now let's do the check on our side

PR #1264, opened 2025-02-28
DO NOT MERGE Experiment to run medium tests in parallel

damien-urruty-sonarsource:


damien-urruty-sonarsource:
There seems to be a problem with Xodus, that is not thread-safe. Closing for now

PR #1269, opened 2025-03-04
SLCORE-1206 Fix add issue comment for VS client

kirill-knize-sonarsource:
[SLCORE-1206](https://sonarsource.atlassian.net/browse/SLCORE-1206)

[SLCORE-1206]: https://sonarsource.atlassian.net/browse/SLCORE-1206?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test-utils/src/main/java/org/sonarsource/sonarlint/core/test/utils/server/ServerFixture.java
+++ b/test-utils/src/main/java/org/sonarsource/sonarlint/core/test/utils/server/ServerFixture.java
@@ -19,6 +19,23 @@
  */
 package org.sonarsource.sonarlint.core.test.utils.server;
 
+import static com.github.tomakehurst.wiremock.client.WireMock.aResponse;
+import static com.github.tomakehurst.wiremock.client.WireMock.equalTo;
+import static com.github.tomakehurst.wiremock.client.WireMock.get;
+import static com.github.tomakehurst.wiremock.client.WireMock.jsonResponse;
+import static com.github.tomakehurst.wiremock.client.WireMock.post;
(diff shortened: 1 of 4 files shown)


test-utils/src/main/java/org/sonarsource/sonarlint/core/test/utils/server/ServerFixture.java
```diff
@@ -19,6 +19,23 @@
  */
 package org.sonarsource.sonarlint.core.test.utils.server;
 
+import static com.github.tomakehurst.wiremock.client.WireMock.aResponse;
```
damien-urruty-sonarsource, line 22: I think your formatting settings might be broken, static imports normally come after the normal ones

damien-urruty-sonarsource (approved):
One minor comment, else LGTM.
Please fill the ticket description to give details about why/what/how, it will help our future selves. The ticket title (and commit message) could also be more precise, it is a bit vague here (it is about falling back on server update when an issue key is not known locally). 
I know the automation is handy to create tickets on our behalf, but we should still keep them in a good shape

kirill-knize-sonarsource:
> One minor comment, else LGTM. Please fill the ticket description to give details about why/what/how, it will help our future selves. The ticket title (and commit message) could also be more precise, it is a bit vague here (it is about falling back on server update when an issue key is not known locally). I know the automation is handy to create tickets on our behalf, but we should still keep them in a good shape

Jira ticket is updated.

PR #1272, opened 2025-03-05
SLCORE-1201 Fix SSF-736

nquinquenel:
[SLCORE-1201](https://sonarsource.atlassian.net/browse/SLCORE-1201)

[SLCORE-1201]: https://sonarsource.atlassian.net/browse/SLCORE-1201?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


damien-urruty-sonarsource (approved):
LGTM. It's the 2nd time we fix a similar SSF, is there a way to check if other places are impacted?

nquinquenel:
> LGTM. It's the 2nd time we fix a similar SSF, is there a way to check if other places are impacted?

For now I just manually checked all the requests going to the client, having some inputs from an external requests. I missed this one so it's definitely not the best way of doing 🤷🏻

PR #1273, opened 2025-03-05
SLCORE-1154: Clients can specify an alternative US region environment

thahnen:
[SLCORE-1154](https://sonarsource.atlassian.net/browse/SLCORE-1154)

This extends the already existing alternative SonarQube Cloud environment (the EU region). This has to be implemented on the client side and can/will be re-used for the tests.

I think with that I found a good balance with not changing too much but relying on `DTO`s instead of adding just 3 more strings (`URI`s) to the `SonarCloudAlternativeEnvironmentDto`. The good thing is that the end user does not always have to have both `DTO`s or for one all `URI`s configured but can just change one specific for testing.

[SLCORE-1154]: https://sonarsource.atlassian.net/browse/SLCORE-1154?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/API_CHANGES.md
+++ b/API_CHANGES.md
@@ -2,8 +2,8 @@
 
 ## New features
 
-* Add a new constructor to `org.sonarsource.sonarlint.core.rpc.protocol.backend.initialize.SonarCloudAlternativeEnvironmentDto` to accept region specific URIs (in a DTO each).
-    * Per region a `org.sonarsource.sonarlint.core.rpc.protocol.backend.initialize.SonarQubeCloudRegionDto` must be provided that contains the *base*, *API* and *WebSocket* URIs.
+* Add a new constructor to `org.sonarsource.sonarlint.core.rpc.protocol.backend.initialize.SonarCloudAlternativeEnvironmentDto` to accept a map of `SonarCloudRegion` to `SonarQubeCloudRegionDto`.
+    *  Per region a `org.sonarsource.sonarlint.core.rpc.protocol.backend.initialize.SonarQubeCloudRegionDto` must be provided that contains the *base*, *API* and *WebSocket* URIs.
     * `null` values are accepted for every URI - it will internally fallback to the actual region URIs for a `null` value encountered.
 
 ## Deprecation
--- a/API_CHANGES.md
+++ b/API_CHANGES.md
@@ -1,5 +1,11 @@
 # 10.17
 
+## Breaking Changes
+
+* Change `org.sonarsource.sonarlint.core.rpc.protocol.backend.initialize.SonarCloudAlternativeEnvironmentDto` to accept region specific URIs (in a DTO each) in the constructor.
+    * Per region a `org.sonarsource.sonarlint.core.rpc.protocol.backend.initialize.SonarQubeCloudRegionDto` must be provided that contains the *base*, *API* and *WebSocket* URIs.
+    * `null` values are accepted for every URI as well for the whole region DTO object - it will internally fallback to the actual region URIs for a `null` value encountered.
+
 ## Deprecation
 
 * Deprecate the `org.sonarsource.sonarlint.core.rpc.protocol.backend.connection.ConnectionRpcService.checkSmartNotificationsSupported` method. It always returns that notifications are supported.
--- a/API_CHANGES.md
+++ b/API_CHANGES.md
@@ -1,10 +1,10 @@
 # 10.17
 
-## Breaking Changes
+## New features
 
-* Change `org.sonarsource.sonarlint.core.rpc.protocol.backend.initialize.SonarCloudAlternativeEnvironmentDto` to accept region specific URIs (in a DTO each) in the constructor.
+* Add a new constructor to `org.sonarsource.sonarlint.core.rpc.protocol.backend.initialize.SonarCloudAlternativeEnvironmentDto` to accept region specific URIs (in a DTO each).
     * Per region a `org.sonarsource.sonarlint.core.rpc.protocol.backend.initialize.SonarQubeCloudRegionDto` must be provided that contains the *base*, *API* and *WebSocket* URIs.
-    * `null` values are accepted for every URI as well for the whole region DTO object - it will internally fallback to the actual region URIs for a `null` value encountered.
+    * `null` values are accepted for every URI - it will internally fallback to the actual region URIs for a `null` value encountered.
 
 ## Deprecation
 
--- a/backend/core/src/main/java/org/sonarsource/sonarlint/core/SonarCloudActiveEnvironment.java
+++ b/backend/core/src/main/java/org/sonarsource/sonarlint/core/SonarCloudActiveEnvironment.java
@@ -20,90 +20,77 @@
 package org.sonarsource.sonarlint.core;
 
 import java.net.URI;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
 import org.sonarsource.sonarlint.core.rpc.protocol.backend.initialize.SonarQubeCloudRegionDto;
 
 import static org.apache.commons.lang.StringUtils.removeEnd;
 
 public class SonarCloudActiveEnvironment {
-  private SonarQubeCloudRegionDto alternativeEuRegion;
-  private SonarQubeCloudRegionDto alternativeUsRegion;
+  private final Map<SonarCloudRegion, SonarQubeCloudRegionDto> alternativeRegionUris;
 
   public static SonarCloudActiveEnvironment prod() {
-    return new SonarCloudActiveEnvironment(new SonarQubeCloudRegionDto(null, null, null),
-      new SonarQubeCloudRegionDto(null, null, null));
+    return new SonarCloudActiveEnvironment(Map.of());
   }
-  
-  public SonarCloudActiveEnvironment(SonarQubeCloudRegionDto alternativeEuRegion, SonarQubeCloudRegionDto alternativeUsRegion) {
-    this.alternativeEuRegion = alternativeEuRegion;
-    this.alternativeUsRegion = alternativeUsRegion;
+
+  public SonarCloudActiveEnvironment(Map<org.sonarsource.sonarlint.core.rpc.protocol.common.SonarCloudRegion, SonarQubeCloudRegionDto> alternativeRegionUris) {
+    this.alternativeRegionUris = alternativeRegionUris.entrySet().stream()
+      .collect(Collectors.toMap(entry -> SonarCloudRegion.valueOf(entry.getKey().name()), Map.Entry::getValue));
   }
 
   public URI getUri(SonarCloudRegion region) {
-    if (region.equals(SonarCloudRegion.US)) {
-      return alternativeUsRegion.getUri() != null
-        ? alternativeUsRegion.getUri()
-        : SonarCloudRegion.US.getProductionUri();
+    if (alternativeRegionUris.containsKey(region) && alternativeRegionUris.get(region).getUri() != null) {
+      return alternativeRegionUris.get(region).getUri();
     }
-    return alternativeEuRegion.getUri() != null
-      ? alternativeEuRegion.getUri()
-      : SonarCloudRegion.EU.getProductionUri();
+    return region.getProductionUri();
   }
 
   public URI getApiUri(SonarCloudRegion region) {
-    if (region.equals(SonarCloudRegion.US)) {
-      return alternativeUsRegion.getApiUri() != null
-        ? alternativeUsRegion.getApiUri()
-        : SonarCloudRegion.US.getApiProductionUri();
+    if (alternativeRegionUris.containsKey(region) && alternativeRegionUris.get(region).getApiUri() != null) {
+      return alternativeRegionUris.get(region).getApiUri();
     }
-    return alternativeEuRegion.getApiUri() != null
-      ? alternativeEuRegion.getApiUri()
-      : SonarCloudRegion.EU.getApiProductionUri();
+    return region.getApiProductionUri();
   }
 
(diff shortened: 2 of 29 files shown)


rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/initialize/SonarCloudAlternativeEnvironmentDto.java
```diff
@@ -19,41 +19,29 @@
...
-  public URI getUri() {
-    return uri;
+/** This contains the alternative environment information for SonarQube Cloud with both the EU and US region */
+public class SonarCloudAlternativeEnvironmentDto {
+  @Nullable
+  private final SonarQubeCloudRegionDto euRegion;
+  @Nullable
+  private final SonarQubeCloudRegionDto usRegion;
+  
+  public SonarCloudAlternativeEnvironmentDto(@Nullable SonarQubeCloudRegionDto euRegion, @Nullable SonarQubeCloudRegionDto usRegion) {
```
sophio-japharidze-sonarsource, line 33: I would suggest not to introduce a breaking change and follow the pattern of deprecating old constructors. Namely, introduce a new one (this one with SonarQubeCloudRegionDtos), keep the original one that would call this new one with production US region URLs.

backend/core/src/main/java/org/sonarsource/sonarlint/core/SonarCloudActiveEnvironment.java
```diff
@@ -20,61 +20,97 @@
...
-    this.alternativeUris = new SonarQubeCloudUris(uri, apiUri, webSocketsEndpointUri);
+  
+  public SonarCloudActiveEnvironment(@Nullable SonarQubeCloudRegionDto alternativeEuRegion, @Nullable SonarQubeCloudRegionDto alternativeUsRegion) {
+    this.alternativeEuRegion = alternativeEuRegion;
+    this.alternativeUsRegion = alternativeUsRegion;
   }
 
   public URI getUri(SonarCloudRegion region) {
-    return alternativeUris != null ? alternativeUris.productionUri : region.getProductionUri();
+    if (region.equals(SonarCloudRegion.EU)) {
```
sophio-japharidze-sonarsource, line 47: I would revert the condition just in case - check if it equals US, and in all other cases default to EU

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/initialize/SonarQubeCloudRegionDto.java
```diff
@@ -0,0 +1,61 @@
...
+package org.sonarsource.sonarlint.core.rpc.protocol.backend.initialize;
+
+import java.net.URI;
+import javax.annotation.CheckForNull;
+import javax.annotation.Nullable;
+
+/** This is used to configure the required URIs for a SonarQube Cloud Region */
+public class SonarQubeCloudRegionDto {
+  @Nullable
+  private final URI uri;
```
sophio-japharidze-sonarsource, line 29: Why are the fields Nullable? I think it should not be possible to provide null values. In the `SonarCloudAlternativeEnvironmentDto`, the "top-level" fields are already Nullable anyway.

API_CHANGES.md
```diff
@@ -1,5 +1,11 @@
 # 10.17
 
+## Breaking Changes
+
+* Change `org.sonarsource.sonarlint.core.rpc.protocol.backend.initialize.SonarCloudAlternativeEnvironmentDto` to accept region specific URIs (in a DTO each) in the constructor.
```
sophio-japharidze-sonarsource, line 5: Same suggestion - maybe we can avoid a breaking change.

medium-tests/src/test/java/mediumtest/SharedConnectedModeSettingsMediumTests.java
```diff
@@ -76,6 +78,41 @@ void should_return_sc_config_when_bound_to_sonarcloud(SonarLintTestHarness harne
...
+      .withBoundConfigScope(configScopeId, connectionId, projectKey)
+      .withTelemetryEnabled()
+      .start();
+
+    var result = getFileContents(backend, configScopeId);
+
+    assertThat(result).succeedsWithin(3, TimeUnit.SECONDS);
+    
+    // This currently returns "EU" instead of "US" but this is what is tested here
+    assertThat(result.get().getJsonFileContent()).isNotEqualTo(expectedFileContent);
```
sophio-japharidze-sonarsource, line 109: Why does it return EU? 🧐

medium-tests/src/test/java/mediumtest/websockets/WebSocketMediumTests.java
```diff
@@ -1448,7 +1450,8 @@ void should_send_one_subscribe_message_per_project_key_when_reopening_connection
   public SonarLintBackendFixture.SonarLintBackendBuilder newBackendWithWebSockets(SonarLintTestHarness harness) {
     return harness.newBackend()
       .withServerSentEventsEnabled()
-      .withSonarCloudWebSocketsUrl(webSocketServer.getUrl());
+      .withSonarQubeCloudEuRegionDto(new SonarQubeCloudRegionDto(null, null, URI.create(webSocketServer.getUrl())))
+      .withSonarQubeCloudUsRegionDto(null);
```
sophio-japharidze-sonarsource, line 1454: wouldn't dropping line 1454 achieve the same result?

sophio-japharidze-sonarsource (commented):
Please see my comments. Overall, looks like it will do the job!

medium-tests/src/test/java/mediumtest/websockets/WebSocketMediumTests.java
thahnen: Yes it would, it is there because the `SonarLintBackendFixure` requires coverage :face_exhaling: And until we introduce proper unit/medium/integration tests for the US region, this is okey to please SonarQube :laughing:

backend/core/src/main/java/org/sonarsource/sonarlint/core/SonarCloudActiveEnvironment.java
thahnen: It used to be only EU related URIs, so I kept it, but I can change it around, no problem.

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/initialize/SonarQubeCloudRegionDto.java
thahnen: They have to accept `null` due to how it will be implemented in the client. The values *can* come from system properties or however it is done on each client, but it can be that only parts of the URIs are provided - see the medium tests for example.
I replace it on the `SonarCloudAlternativeEnvironmentDto` to not have `null` values allowed there by following the approach to not introduce breaking changes.

medium-tests/src/test/java/mediumtest/SharedConnectedModeSettingsMediumTests.java
thahnen: I have no idea, I accepted it (it is for coverage again) until we implement the actual proper unit/integration/medium tests in the following days.

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/initialize/SonarCloudAlternativeEnvironmentDto.java
thahnen: Done

backend/core/src/main/java/org/sonarsource/sonarlint/core/SonarCloudActiveEnvironment.java
thahnen: Done

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/initialize/SonarQubeCloudRegionDto.java
thahnen: Done

backend/core/src/main/java/org/sonarsource/sonarlint/core/SonarCloudActiveEnvironment.java
```diff
@@ -20,61 +20,90 @@
...
 
 public class SonarCloudActiveEnvironment {
-
-  private SonarQubeCloudUris alternativeUris;
+  private SonarQubeCloudRegionDto alternativeEuRegion;
+  private SonarQubeCloudRegionDto alternativeUsRegion;
 
   public static SonarCloudActiveEnvironment prod() {
-    return new SonarCloudActiveEnvironment();
+    return new SonarCloudActiveEnvironment(new SonarQubeCloudRegionDto(null, null, null),
```
sophio-japharidze-sonarsource, line 32: This looks very hacky and hard to follow. Let's keep the empty constructor as it was before and check for "top-level" nulls in the `getUri` methods. WDYT?

backend/core/src/main/java/org/sonarsource/sonarlint/core/SonarCloudActiveEnvironment.java
```diff
@@ -20,61 +20,83 @@
...
+import java.util.Map;
+import java.util.stream.Collectors;
+import org.sonarsource.sonarlint.core.rpc.protocol.backend.initialize.SonarQubeCloudRegionDto;
 
 import static org.apache.commons.lang.StringUtils.removeEnd;
 
 public class SonarCloudActiveEnvironment {
-
-  private SonarQubeCloudUris alternativeUris;
+  private final Map<SonarCloudRegion, SonarQubeCloudRegionDto> alternateRegionUris;
```
sophio-japharidze-sonarsource, line 30:
```suggestion
  private final Map<SonarCloudRegion, SonarQubeCloudRegionDto> alternativeRegionUris;
```

thahnen: Done

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/initialize/SonarCloudAlternativeEnvironmentDto.java
```diff
@@ -36,24 +37,20 @@ public SonarCloudAlternativeEnvironmentDto(URI uri, URI webSocketsEndpointUri) {
...
-
-  public URI getUri() {
-    return uri;
+    this(Map.of(SonarCloudRegion.EU, new SonarQubeCloudRegionDto(uri, apiUri, webSocketsEndpointUri)));
   }
-
-  public URI getApiUri() {
-    return apiUri;
+  
+  public SonarCloudAlternativeEnvironmentDto(Map<SonarCloudRegion, SonarQubeCloudRegionDto> alternateRegionUris) {
```
sophio-japharidze-sonarsource, line 49: `alternativeRegionUris`

thahnen: Done

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/initialize/SonarCloudAlternativeEnvironmentDto.java
```diff
@@ -20,11 +20,12 @@
...
 import java.net.URI;
+import java.util.Map;
+import org.sonarsource.sonarlint.core.rpc.protocol.common.SonarCloudRegion;
 
+/** This contains the alternative environment information for SonarQube Cloud with both the EU and US region */
 public class SonarCloudAlternativeEnvironmentDto {
-  private final URI uri;
-  private final URI apiUri;
-  private final URI webSocketsEndpointUri;
+  private final Map<SonarCloudRegion, SonarQubeCloudRegionDto> alternateRegionUris;
```
sophio-japharidze-sonarsource, line 28:
```suggestion
  private final Map<SonarCloudRegion, SonarQubeCloudRegionDto> alternativeRegionUris;
```

backend/core/src/main/java/org/sonarsource/sonarlint/core/SonarCloudActiveEnvironment.java
```diff
@@ -20,61 +20,77 @@
...
+    for (var entry : alternativeRegionUris.entrySet()) {
+      var regionDto = entry.getValue();
+      if (regionDto.getUri() != null && removeEnd(regionDto.getUri().toString(), "/").equals(cleanedUri)) {
+        return Optional.of(entry.getKey());
+      }
+    }
 
-  private record SonarQubeCloudUris(URI productionUri, URI productionApiUri, URI wsUri) {
+    for (var region : SonarCloudRegion.values()) {
+      if (removeEnd(region.getProductionUri().toString(), "/").equals(uri)) {
```
sophio-japharidze-sonarsource, line 89:
```suggestion
      if (removeEnd(region.getProductionUri().toString(), "/").equals(cleanedUri)) {
```

sophio-japharidze-sonarsource (approved):
LGTM! See minor leftover comments

PR #1274, opened 2025-03-06
SLCORE-1074 Introduce an analysis scheduler component

kirill-knize-sonarsource:
[SLCORE-1211](https://sonarsource.atlassian.net/browse/SLCORE-1211)


damien-urruty-sonarsource (approved):
Let's keep the duplication and coverage issues for a bit later in the iteration

PR #1275, opened 2025-03-06
SLCORE-1212 Don't fail on git blaming newly created file

kirill-knize-sonarsource:
[SLCORE-1212](https://sonarsource.atlassian.net/browse/SLCORE-1212)

[SLCORE-1212]: https://sonarsource.atlassian.net/browse/SLCORE-1212?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/GitUtils.java
+++ b/backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/GitUtils.java
@@ -154,10 +154,10 @@ public class GitUtils {
         if (exitCode == 0) {
           nativeGitExecutable = executable;
         }
-      } catch (IOException e) {
-        LOG.debug("Checking for native Git executable failed", e);
       } catch (InterruptedException e) {
         Thread.currentThread().interrupt();
+      } catch (Exception e) {
+        LOG.debug("Checking for native Git executable failed", e);
       }
       checkedForNativeGitExecutable = true;
     }
--- a/backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/GitUtils.java
+++ b/backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/GitUtils.java
@@ -55,6 +55,7 @@ import org.sonarsource.sonarlint.core.commons.SonarLintBlameResult;
 import org.sonarsource.sonarlint.core.commons.SonarLintGitIgnore;
 import org.sonarsource.sonarlint.core.commons.Version;
 import org.sonarsource.sonarlint.core.commons.log.SonarLintLogger;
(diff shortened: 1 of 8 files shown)


backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/GitUtils.java
```diff
@@ -153,7 +154,7 @@ private static String getNativeGitExecutable() {
         if (exitCode == 0) {
           nativeGitExecutable = executable;
         }
-      } catch (IOException e) {
+      } catch (IOException | IllegalStateException e) {
```
damien-urruty-sonarsource, line 157: Wouldn't it be safer to catch Exception here?

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/exceptions/GitException.java
```diff
@@ -0,0 +1,32 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.sonarlint.core.commons.util.git.exceptions;
+
+public class GitException extends IllegalStateException {
```
damien-urruty-sonarsource, line 22: Nitpick: I think it's uncommon to extend `IllegalStateException`, I don't see what benefit it brings. Usually we directly extend `RuntimeException`

damien-urruty-sonarsource (approved):
Some suggestions for improvement, else LGTM

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/exceptions/GitException.java
kirill-knize-sonarsource: Because in many places we throw IllegalStateException like in ProcessWrapperFactory or in WinGitUtils, and in corresponding catch blocks we expect the IllegalStateException itself, or something that extends it.

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/GitUtils.java
kirill-knize-sonarsource: Good point, changed to exception.

PR #1280, opened 2025-03-10
SLCORE-1078  Start executing analyses only when preconditions are met

kirill-knize-sonarsource:
[SLCORE-1218](https://sonarsource.atlassian.net/browse/SLCORE-1218)

[SLCORE-1218]: https://sonarsource.atlassian.net/browse/SLCORE-1218?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisService.java
```diff
@@ -498,15 +474,19 @@ public void onStandaloneRulesConfigurationChanged(StandaloneRulesConfigurationCh
     }
   }
 
-  public UUID analyzeOpenFiles(String configScopeId) {
+  @CheckForNull
```
kirill-knize-sonarsource, line 477: RPC client is not one of the callers now, it's the single caller, so we might want to review what's going on here.

backend/core/src/main/java/org/sonarsource/sonarlint/core/fs/FileExclusionService.java
```diff
@@ -185,7 +192,120 @@ public boolean isExcluded(URI fileUri) {
...
+    var filteredURIsNotUserDefined = new ArrayList<URI>();
+    var filteredURIsFromSymbolicLink = new ArrayList<URI>();
+    var filteredURIsFromWindowsShortcut = new ArrayList<URI>();
+    var filteredURIsNoFile = new ArrayList<URI>();
+
+    // Do the actual filtering and in case of a filtered out URI, save them for later logging!
+    var actualFilesToAnalyze = filterOutClientExcludedFiles(configurationScopeId, files)
+      .stream()
+      .map(uri -> {
+        var file = findFile(configurationScopeId, uri);
```
kirill-knize-sonarsource, line 229: Note for future:
This is a potential performance problem when we will switch to lazy virtual file system initialization. For large list of files we might spend a lot of time here.

backend/core/src/main/java/org/sonarsource/sonarlint/core/fs/FileExclusionService.java
```diff
@@ -185,7 +192,120 @@ public boolean isExcluded(URI fileUri) {
...
+        if (file == null) {
+          filteredURIsNoFile.add(uri);
```
kirill-knize-sonarsource, line 231: Important part of functionality is not tested, to be tested later..

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/analysis/ForceAnalyzeResponse.java
```diff
@@ -21,11 +21,12 @@
 
 import java.util.UUID;
 import javax.annotation.CheckForNull;
+import javax.annotation.Nullable;
 
 public class ForceAnalyzeResponse {
   UUID analysisId;
 
-  public ForceAnalyzeResponse(UUID analysisId) {
+  public ForceAnalyzeResponse(@Nullable UUID analysisId) {
```
kirill-knize-sonarsource, line 29: This nullable analysis ID comes from a single place which does not look like a strong reason. Maybe we should revisit it. See other comment on AnalysisService.

backend/core/src/main/java/org/sonarsource/sonarlint/core/fs/FileExclusionService.java
damien-urruty-sonarsource: Ah yes, indeed there is a lack of coverage here. I will cover it later

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/analysis/ForceAnalyzeResponse.java
damien-urruty-sonarsource: We can revisit yes in the current iteration yes. FYI it's not something that I changed, the analysisId was nullable before, I just fixed the annotation

backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisService.java
damien-urruty-sonarsource: It is. `forceAnalyzeOpenFiles` is called by `org.sonarsource.sonarlint.core.rpc.impl.AnalysisRpcServiceDelegate#analyzeOpenFiles`

PR #1293, opened 2025-03-20
SLCORE-1234 Keep track of the connections state

kirill-knize-sonarsource:
[SLCORE-1234](https://sonarsource.atlassian.net/browse/SLCORE-1234)

Solution is to introduce the cache of ServerConnection objects on the level of ConnectionManager. 

Getting connection without credentials is a creation of one-time use adhoc object to send a request that doesn't require authorization. So connections without credentials are not being stored in cache. And they never send connection error notifications. 

[SLCORE-1234]: https://sonarsource.atlassian.net/browse/SLCORE-1234?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/core/src/main/java/org/sonarsource/sonarlint/core/ConnectionManager.java
+++ b/backend/core/src/main/java/org/sonarsource/sonarlint/core/ConnectionManager.java
@@ -19,7 +19,9 @@
  */
 package org.sonarsource.sonarlint.core;
 
+import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import javax.annotation.Nullable;
@@ -52,6 +54,7 @@ public class ConnectionManager {
   private final HttpClientProvider httpClientProvider;
   private final SonarLintRpcClient client;
   private final SonarCloudActiveEnvironment sonarCloudActiveEnvironment;
+  private final Map<String, ServerConnection> connectionCache = new ConcurrentHashMap<>();
 
   public ConnectionManager(ConnectionConfigurationRepository connectionRepository, ConnectionAwareHttpClientProvider awareHttpClientProvider, HttpClientProvider httpClientProvider,
     SonarCloudActiveEnvironment sonarCloudActiveEnvironment, SonarLintRpcClient client) {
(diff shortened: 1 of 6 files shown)


backend/core/src/main/java/org/sonarsource/sonarlint/core/ConnectionManager.java
```diff
@@ -145,24 +148,26 @@ private HttpClient getClientFor(EndpointParams params, Either<TokenDto, Username
...
+    return connectionCache.computeIfAbsent(connectionId, connId -> {
+      var serverApi = getServerApiOrThrow(connId);
+      return new ServerConnection(connId, serverApi, client);
+    });
   }
 
   /**
    * Returns empty Optional if connection with provided ID is not found in ConnectionConfigurationRepository
    */
   public Optional<ServerConnection> tryGetConnection(String connectionId) {
```
damien-urruty-sonarsource, line 160: Not directly related to your change, but as we discussed before the API surface for this manager is huge. I think this method could be removed. It's used in a single place, where I guess we could use getConnectionOrThrow

backend/core/src/main/java/org/sonarsource/sonarlint/core/ConnectionManager.java
```diff
@@ -145,24 +148,26 @@ private HttpClient getClientFor(EndpointParams params, Either<TokenDto, Username
...
+      getServerApi(connId).map(serverApi -> new ServerConnection(connId, serverApi, client)).orElse(null)));
   }
 
   /**
    * Should be used for WebAPI requests without an authentication
    */
   public Optional<ServerConnection> tryGetConnectionWithoutCredentials(String connectionId) {
     return getServerApiWithoutCredentials(connectionId)
-      .map(serverApi -> new ServerConnection(connectionId, serverApi, client));
+      .map(serverApi -> new ServerConnection(connectionId, serverApi, client, true));
```
damien-urruty-sonarsource, line 170: Similarly, I think this method should be removed. The only use case is when we want to fetch the server version to check if it is soon unsupported. I don't see why we wouldn't make an authenticated call. If this specific endpoint does not require authentication, then it's more at the `SystemApi` level that we should not pass the credentials when calling `get`, maybe we need a `getAnonymous` in the ServerApiHelper. This would simplify the new constructor you added

damien-urruty-sonarsource (approved):
LGTM, I left 2 suggestions for further improvements, not directly related to your change. If you don't plan to implement them, feel free to create tickets

backend/core/src/main/java/org/sonarsource/sonarlint/core/ConnectionManager.java
kirill-knize-sonarsource: Unfortunately, it's not a trivial change. I will address it in a separate PR later.

backend/core/src/main/java/org/sonarsource/sonarlint/core/ConnectionManager.java
kirill-knize-sonarsource: Same. See other comment.

PR #1296, opened 2025-03-24
SLCORE-1081 Batch automatic analyses

damien-urruty-sonarsource:
[SLCORE-1081](https://sonarsource.atlassian.net/browse/SLCORE-1081)

[SLCORE-1081]: https://sonarsource.atlassian.net/browse/SLCORE-1081?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/api/TriggerType.java
+++ b/backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/api/TriggerType.java
@@ -20,5 +20,23 @@
 package org.sonarsource.sonarlint.core.analysis.api;
 
 public enum TriggerType {
-  AUTO, FORCED
+  AUTO(true, true),
+  FORCED(false, false),
+  FORCED_WITH_EXCLUSIONS(true, false);
+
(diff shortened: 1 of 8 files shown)


backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/api/TriggerType.java
```diff
@@ -20,5 +20,5 @@
 package org.sonarsource.sonarlint.core.analysis.api;
 
 public enum TriggerType {
-  AUTO, FORCED
+  AUTO, FORCED, ANALYZE_FILES_AND_TRACK
```
kirill-knize-sonarsource, line 23: We can change the naming or document in code what is the rationale behind these 3 types.

damien-urruty-sonarsource: I change the names of the constants, and also tried to make things more explicit in the enum. Let me know what you think

PR #1297, opened 2025-03-25
SLCORE-1234 Evict connections cache when needed

damien-urruty-sonarsource:
[SLCORE-1234](https://sonarsource.atlassian.net/browse/SLCORE-1234)

[SLCORE-1234]: https://sonarsource.atlassian.net/browse/SLCORE-1234?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/medium-tests/src/test/java/mediumtest/synchronization/ConnectionSyncMediumTests.java
+++ b/medium-tests/src/test/java/mediumtest/synchronization/ConnectionSyncMediumTests.java
@@ -26,8 +26,11 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import org.sonarsource.sonarlint.core.commons.RuleType;
 import org.sonarsource.sonarlint.core.commons.api.TextRange;
+import org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding.BindingConfigurationDto;
+import org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding.DidUpdateBindingParams;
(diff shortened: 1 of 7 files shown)


medium-tests/src/test/java/mediumtest/synchronization/ConnectionSyncMediumTests.java
```diff
@@ -145,19 +148,71 @@ void it_should_notify_client_on_invalid_token_exactly_once() {
...
     await().untilAsserted(() -> assertThat(client.getLogMessages()).contains("Error during synchronization"));
+    await().during(5, TimeUnit.SECONDS).untilAsserted(() -> assertThat(client.getConnectionIdsWithInvalidToken(CONNECTION_ID)).isEqualTo(1));
+  }
 
-    backend.getConnectionService().didChangeCredentials(new DidChangeCredentialsParams(CONNECTION_ID));
+  @SonarLintTest
+  void it_should_renotify_client_on_invalid_token_after_connection_is_recreated() {
+    var status = 401;
+    var client = newFakeClient()
+      .withCredentials(CONNECTION_ID, "user", "pw")
```
kirill-knize-sonarsource, line 167: I think there's a long term plan to remove user-password auth from SQC. Not sure about SQS, but can be as well. Maybe we should rely mostly on token auth in tests. Not urgent to fix now.

damien-urruty-sonarsource: Good point, I replaced with tokens in this class to make it easier later

PR #1302, opened 2025-03-26
SLCORE-1240 take hotspot resolution into account

sophio-japharidze-sonarsource:
I should have done this change from the very beginning.
In any case, once https://github.com/SonarSource/sonar-enterprise/pull/13291 is released, we will have access to hotspot resolution.


damien-urruty-sonarsource (approved):
LGTM but I would suggest to add a test to verify that this new resolution is well taken into account

PR #1310, opened 2025-04-02
SLCORE-1261: Warn users that 9.9 LTA will soon be unsupported

thahnen:
[SLCORE-1261](https://sonarsource.atlassian.net/browse/SLCORE-1261)

With the latest LTA being 2025.1 we should warn every between 9.9 LTA and the new, latest LTA, that they will soon run out of support.

[SLCORE-1261]: https://sonarsource.atlassian.net/browse/SLCORE-1261?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/commons/src/test/java/org/sonarsource/sonarlint/core/commons/VersionTests.java
+++ b/backend/commons/src/test/java/org/sonarsource/sonarlint/core/commons/VersionTests.java
@@ -24,6 +24,12 @@ import org.junit.jupiter.api.Test;
 import static org.assertj.core.api.Assertions.assertThat;
 
 class VersionTests {
+  @Test
+  void test_versioning_changes_handled_correctly() {
+    var newVersion = Version.create("2025.1.0.102418");
+    var oldVersion = Version.create("9.9.0.65466");
+    assertThat(newVersion).isGreaterThan(oldVersion);
+  }
 
   @Test
   void test_fields_of_snapshot_versions() {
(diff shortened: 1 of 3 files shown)


backend/commons/src/test/java/org/sonarsource/sonarlint/core/commons/VersionTests.java
```diff
@@ -24,6 +24,12 @@
 import static org.assertj.core.api.Assertions.assertThat;
 
 class VersionTests {
+  @Test
+  void test_versioning_changes_handled_correctly() {
```
sophio-japharidze-sonarsource, line 28: Let's add another test also for Community Build versioning, just in case - it would be `25.1`

backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/VersionUtils.java
```diff
@@ -23,15 +23,16 @@
 
 public class VersionUtils {
 
-  private static final Version CURRENT_LTS = Version.create("9.9");
+  private static final Version CURRENT_LTS = Version.create("2025.1");
```
sophio-japharidze-sonarsource, line 26:
```suggestion
  private static final Version CURRENT_LTA = Version.create("2025.1");
```

sophio-japharidze-sonarsource (commented):
LGTM; Need to test manually before I approve

thahnen:
We will do better :smile:

PR #1311, opened 2025-04-02
SLCORE-1236 Let clients cancel automatic analyses

damien-urruty-sonarsource:
[SLCORE-1236](https://sonarsource.atlassian.net/browse/SLCORE-1236)

[SLCORE-1236]: https://sonarsource.atlassian.net/browse/SLCORE-1236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/container/global/ModuleRegistry.java
```diff
@@ -34,37 +33,29 @@
...
-    var moduleContainer = new ModuleContainer(parent, false);
+    var moduleContainer = new ModuleContainer(parent);
     if (clientFileSystem != null) {
       moduleContainer.add(clientFileSystem);
     }
     moduleContainer.startComponents();
     return moduleContainer;
   }
 
-  public ModuleContainer createTransientContainer(Iterable<ClientInputFile> filesToAnalyze) {
```
kirill-knize-sonarsource, line 59: Why do we remove notion of the transient container? It is needed for single file analysis for files outside of the config scope. TransientModuleFileSystem.class left unused after this change. We need to remove it as well if it's intentional change.

damien-urruty-sonarsource: > It is needed for single file analysis for files outside of the config scope

This is already not supported anymore, all files are part of a config scope. For the LS for example, they create an "artificial" root configuration scope that serves for files outside folders.

Good catch for the TransientModuleFileSystem, removing it now

kirill-knize-sonarsource (approved):
LGTM

PR #1313, opened 2025-04-03
SLCORE-1264 Ignore Maven Wrapper During SCA Step

mikeyoung85:
[SLCORE-1264](https://sonarsource.atlassian.net/browse/SLCORE-1264)

This PR adds the option in the `test_linux` build step to ignore the Maven Wrapper committed to this project and instead use the provided Maven binary that is being used to run the SQ analysis. This should help prevent the dependency resolution steps during the SCA step in the SQ scanner process from failing to resolve dependency information due to the `MAVEN_CONFIG` option that seems to break the Maven Wrapper calls.

[SLCORE-1264]: https://sonarsource.atlassian.net/browse/SLCORE-1264?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -103,7 +103,7 @@ test_linux_task:
   <<: *MAVEN_CACHE_DEFINITION
   script:
     - source cirrus-env QA
-    - PULL_REQUEST_SHA=$GIT_SHA1 regular_mvn_build_deploy_analyze -P-deploy-sonarsource,-release,-sign -Dcommercial -Dmaven.install.skip=true -Dmaven.deploy.skip=true -Dsonar.coverage.jacoco.xmlReportPaths=$CIRRUS_WORKING_DIR/report-aggregate/target/site/jacoco-aggregate/jacoco.xml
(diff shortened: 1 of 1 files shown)


mikeyoung85:
It looks like disabling the Maven Wrapper allowed the dependencies to be discovered correctly. The analysis for this PR shows quite a few more dependencies and the logs do not show any errors anymore.

damien-urruty-sonarsource:
@mikeyoung85 thanks for the in-depth investigation!

If upgrading the Maven wrapper to the latest version fixes the issue, I would rather do that than tweaking some configuration. There shouldn't be any blocker for us to upgrade. I will take care of that next week and close this PR when done

damien-urruty-sonarsource:
@mikeyoung85 I confirm bumping the Maven wrapper version works (#1317), I'm now able to see Dependency risks on Next. I am closing this PR. Thanks a lot for the investigation!

PR #1321, opened 2025-04-09
SLCORE-1274 Fix SonarQubeDeveloperEditonTests

eray-felek-sonarsource:
[SLCORE-1274](https://sonarsource.atlassian.net/browse/SLCORE-1274)

[SLCORE-1274]: https://sonarsource.atlassian.net/browse/SLCORE-1274?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/its/tests/src/test/java/its/SonarQubeDeveloperEditionTests.java
+++ b/its/tests/src/test/java/its/SonarQubeDeveloperEditionTests.java
@@ -366,7 +366,10 @@ class SonarQubeDeveloperEditionTests extends AbstractConnectedTests {
 
       var rawIssues = analyzeFile(configScopeId, "sample-cloudformation", "src/sample.yaml");
 
-      assertThat(rawIssues).hasSize(2);
+      assertThat(rawIssues).satisfiesAnyOf(
+              issues -> assertThat(issues).hasSize(1),
+              issues -> assertThat(issues).hasSize(2)
+      );
     }
 
     @Test


thahnen (approved):
_90 Degree crew_ assemble :speaking_head: :loudspeaker:

⌐
ᘏ 
⊣
ᴟ

damien-urruty-sonarsource:
Thanks for fixing the issue @eray-felek-sonarsource.

I'm sorry to intervene here, but there are lot of things that caught my attention on this PR. I wouldn't have approved it myself, and I would like to give some feedback:

* I have no clue what the cause of the test failure was. It's not documented anywhere. Why is the analyzer finding only one issue now compared to 2 before? Is it expected? Was a rule removed from the default QP?
* The Jira ticket is blank. Related to my first point, there is nothing there, we should at least document our findings, so we can remember why some changes were made
* The Jira ticket is outside of any hardening or any version. I think the automatic ticket creation is being more and more abused, and it leads to many ticket remaining blank and disconnected from the sprint and releases
* As a consequence, the ticket number is not part of the commit message
* I find the actual fix a bit strange. Now if for an unexpected reason we start raising 2 issues on an older SQS version, the test would not catch it. There are other places where we check the server version to determine how many issues should be raised, and that's how we should do it here

I encourage you to fix what you can, maybe even the test code, and be a bit more mindful next time, specially around documenting what you did

PR #1326, opened 2025-04-11
SLCORE-1260 Report all applied fixes in a single per-install counter

jblievremont:
[SLCORE-1260](https://sonarsource.atlassian.net/browse/SLCORE-1260)

[SLCORE-1260]: https://sonarsource.atlassian.net/browse/SLCORE-1260?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryHttpClient.java
+++ b/backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryHttpClient.java
@@ -166,14 +166,17 @@ public class TelemetryHttpClient {
       ))
       .forEach(values::add);
 
-    data.getQuickFixCountByRuleKey().entrySet().stream()
-      .map(e -> {
-        var ruleKey = e.getKey();
-        var ruleRepoAndId = ruleKey.split(":");
-        var measureKey = String.format("quick_fix.applied_count.%s.%s", ruleRepoAndId[0], ruleRepoAndId[1]);
-        return new TelemetryMeasuresValue(measureKey, String.valueOf(e.getValue()), INTEGER, DAILY);
-      })
-      .forEach(values::add);
+    var allQuickFixCount = data.getQuickFixCountByRuleKey().values().stream()
+      .mapToInt(Integer::intValue)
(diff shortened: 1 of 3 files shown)


backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryHttpClient.java
```diff
@@ -166,14 +166,17 @@ private TelemetryMeasuresPayload createMeasuresPayload(TelemetryLocalStorage dat
...
-        var ruleKey = e.getKey();
-        var ruleRepoAndId = ruleKey.split(":");
-        var measureKey = String.format("quick_fix.applied_count.%s.%s", ruleRepoAndId[0], ruleRepoAndId[1]);
-        return new TelemetryMeasuresValue(measureKey, String.valueOf(e.getValue()), INTEGER, DAILY);
-      })
-      .forEach(values::add);
+    var allQuickFixCount = data.getQuickFixCountByRuleKey().values().stream()
+      .mapToInt(Integer::intValue)
+      .sum();
+    if (allQuickFixCount > 0) {
```
damien-urruty-sonarsource, line 172: Don't we also want to collect the measure if the count is 0?

damien-urruty-sonarsource (approved):
It's a bit annoying that we cannot collect the rule key, we are a bit limited by the framework. LGTM, I just left a question for an edge case

backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryHttpClient.java
jblievremont: As far as I can tell, the goal is to count quick fixes applied on a given period, so reporting 0 as a quick fix count for a given period has the same effect as not reporting a value, with an additional row in the generated measures table. My guess is that this will be the case for most installs, so a lot of not-so-useful lines with a value of 0 in the table.

PR #1336, opened 2025-04-17
Remove excessive logging for native git blame

kirill-knize-sonarsource:


backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/BlameParser.java
```diff
@@ -48,13 +48,12 @@ public static void parseBlameOutput(String blameOutput, String currentFilePath,
...
+    var lineBlameSnippets = blameOutput.split(FILENAME);
     var currentLineNumber = 0;
 
-    for (var fileSection : fileSections) {
-      if (shouldSkipSection(fileSection, currentLineNumber >= numberOfLines)) {
-        continue;
-      }
+    // because of the way we split the blame output, the last snippet should be skipped
+    for (var i = 0; i < lineBlameSnippets.length - 1; i++) {
+      var fileSection = lineBlameSnippets[i];
```
damien-urruty-sonarsource, line 56: I have the impression it's not only a change about logging but also a change of behavior. There was some kind of filtering before and it'snot the case anymore. Is it expected? If so, could you update the ticket to describe that?

kirill-knize-sonarsource: Updated the ticket description.

damien-urruty-sonarsource (approved):

PR #1342, opened 2025-04-28
SLCORE-1272 Rollback update jgit version

kirill-knize-sonarsource:


vnaskos-sonar:
Context: The rollback is necessary because Eclipse cannot be built using the jgit version 7.2. There was a conflict with the minimum java version.

vnaskos-sonar (approved):
👍

PR #1344, opened 2025-05-01
SLCORE-1325 Windows where.exe tool output is being parsed incorrectly for two git installations case

kirill-knize-sonarsource:
[SLCORE-1325](https://sonarsource.atlassian.net/browse/SLCORE-1325)

[SLCORE-1325]: https://sonarsource.atlassian.net/browse/SLCORE-1325?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/WinGitUtils.java
+++ b/backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/WinGitUtils.java
@@ -19,7 +19,9 @@
  */
 package org.sonarsource.sonarlint.core.commons.util.git;
 
+import java.util.Arrays;
 import java.util.Optional;
+import java.util.function.Supplier;
 import org.sonarsource.sonarlint.core.commons.log.SonarLintLogger;
 
 public class WinGitUtils {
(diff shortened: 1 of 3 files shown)


backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/WinGitUtils.java
```diff
@@ -30,19 +32,28 @@ private WinGitUtils() {
...
-    var result = new ProcessWrapperFactory()
-      .create(null, "C:\\Windows\\System32\\where.exe", "$PATH:git.exe")
-      .execute();
+    var result = gitExeLocationSupplier.get();
 
-    if (result.exitCode() == 0) {
-      var out = result.output().trim();
+    var output = result.output();
+    if (result.exitCode() == 0 && output.contains("git.exe")) {
+      var out = Arrays.stream(output.split(System.lineSeparator())).map(String::trim).findFirst();
```
damien-urruty-sonarsource, line 45: It might not happen often, but as we randomly pick the first found Git, we might end up using an older version than what we need, while another Git would have suited better. But the minimal version we require is 6y old, so maybe it wouldn't be a common issue

damien-urruty-sonarsource (approved):
I have one small concern about picking the first result randomly, the rest looks good

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/WinGitUtils.java
kirill-knize-sonarsource: Concern is fair, but I suggest not to write the version picking algorithm now.
Ideally yes, we should call the `git --version` here and make sure it's good for us. If not, check another version and fallback to JGit if none is OK.
I suggest to proceed this way now.

PR #1347, opened 2025-05-06
SLCORE-1341 Fix NPE when an analysis command is canceled

kirill-knize-sonarsource:
[SLCORE-1341](https://sonarsource.atlassian.net/browse/SLCORE-1341)

[SLCORE-1341]: https://sonarsource.atlassian.net/browse/SLCORE-1341?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisScheduler.java
+++ b/backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisScheduler.java
@@ -64,12 +64,14 @@ public class AnalysisScheduler {
   }
 
   private void executeQueuedCommands() {
-    while (termination.get() == null) {
+    if (termination.get() == null) {
       SonarLintLogger.get().setTarget(logOutput);
+    }
+    while (termination.get() == null) {
       try {
         executingCommand.set(analysisQueue.takeNextCommand());
         if (termination.get() == CANCELING_TERMINATION) {
-          executingCommand.getAndSet(null).cancel();
+          executingCommand.get().cancel();
           break;
         }
         executingCommand.get().execute(globalAnalysisContainer.get().getModuleRegistry());
@@ -78,6 +80,8 @@ public class AnalysisScheduler {
         if (termination.get() != CANCELING_TERMINATION) {
           LOG.error("Analysis engine interrupted", e);
         }
+      } catch (Exception e) {
+        LOG.debug("Analysis command failed", e);
       }
     }
     termination.get().run();
--- a/backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisScheduler.java
+++ b/backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisScheduler.java
@@ -64,8 +64,8 @@ public class AnalysisScheduler {
   }
 
   private void executeQueuedCommands() {
-    SonarLintLogger.get().setTarget(logOutput);
     while (termination.get() == null) {
+      SonarLintLogger.get().setTarget(logOutput);
       try {
         var command = analysisQueue.takeNextCommand();
         executingCommand.set(command);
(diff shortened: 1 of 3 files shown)


backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisScheduler.java
```diff
@@ -64,12 +64,14 @@ public void wakeUp() {
...
-    while (termination.get() == null) {
+    if (termination.get() == null) {
       SonarLintLogger.get().setTarget(logOutput);
+    }
+    while (termination.get() == null) {
       try {
         executingCommand.set(analysisQueue.takeNextCommand());
         if (termination.get() == CANCELING_TERMINATION) {
-          executingCommand.getAndSet(null).cancel();
+          executingCommand.get().cancel();
```
damien-urruty-sonarsource, line 74: In fact, if I'm not wrong, when the termination is set, the executing command is canceled in the `stop` method. So maybe the only thing to do here is to break?

kirill-knize-sonarsource: No, it's not enough because it will not work on windows. I don't have the specific details what exactly happens in different order. Maybe we should spend time investigating and documenting it because it's not the first time I see discrepancies with windows threading.
But long story short what you written above is only correct for mac/linux.

backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisScheduler.java
```diff
@@ -64,12 +64,14 @@ public void wakeUp() {
   }
 
   private void executeQueuedCommands() {
-    while (termination.get() == null) {
+    if (termination.get() == null) {
       SonarLintLogger.get().setTarget(logOutput);
+    }
```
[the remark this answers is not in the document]
kirill-knize-sonarsource: Same reason, setting log output used to sit inside while loop protected by loop condition. To make it once before the loop and not break on windows this condition is necessary to keep logic the same. Otherwise we starting to have flaky behavior on Windows.

backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisScheduler.java
damien-urruty-sonarsource: It was already there, so we can leave it like this. But it would be great to understand what we are doing :)

backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisScheduler.java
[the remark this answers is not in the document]
damien-urruty-sonarsource: Same answer from me, keeping things just to avoid flaky tests and without really understanding what we are doing seems a bit fragile and not good on the long run.

damien-urruty-sonarsource (approved):
The fix should be functionally correct now. There is still code that might not be necessary, but it doesn't complicate the logic too much, so we can keep it like this

backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisScheduler.java
kirill-knize-sonarsource: For the history nature of the problem:

We start, we wake up executing thread. It start to wait for command. Then we post command from main thread, then we stop scheduler from main thread, executing command atomic was never set yet because executing thread is still sleeping.
So we cleanup the queue during stop, finalize everything. And only then executing thread wakes up and getting new command from the queue. By this moment it sets the command to the Atomic, but stop() already long gone, so there's no one to cancel it.
Then we notice that everything is canceled and break the executing loop.

Solution is to add Thread.sleep(100) to affected tests in order to let scheduler to pickup and properly process the posted command from the queue.

PR #1348, opened 2025-05-09
Use latest Plugin API

kirill-knize-sonarsource:


damien-urruty-sonarsource (approved):
LGTM! Could you also make sure to update the [README](https://github.com/SonarSource/sonar-plugin-api/?tab=readme-ov-file#sonarqube-for-ide) of the sonar-plugin-api repo with the right version after the release?

PR #1358, opened 2025-05-19
SLCORE-1375 introduce getConnectionSuggestions method

sophio-japharidze-sonarsource:


rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/connection/ConnectionRpcService.java
```diff
@@ -126,4 +129,11 @@ public interface ConnectionRpcService {
    */
   @JsonRequest
   CompletableFuture<GetProjectNamesByKeyResponse> getProjectNamesByKey(GetProjectNamesByKeyParams params);
+
```
nquinquenel, line 132:
```suggestion
```

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/client/connection/GetConnectionSuggestionsParams.java
```diff
@@ -0,0 +1,33 @@
...
+
+  private final String configurationScopeId;
+
+  public GetConnectionSuggestionsParams(String configurationScopeId) {
+    this.configurationScopeId = configurationScopeId;
+  }
+
+  public String getConfigurationScopeId() {
+    return configurationScopeId;
+  }
```
nquinquenel, line 32:
```suggestion
  }
  
```

For consistency, I always add a new line before/after the class (or none)

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/connection/ConnectionRpcService.java
```diff
@@ -126,4 +129,11 @@ public interface ConnectionRpcService {
...
+
+  /*
+   * Returns a list of connection suggestions for the given configuration scope. To be used when neither connection nor binding exists.
+   */
+  @JsonRequest
+  CompletableFuture<List<ConnectionSuggestionDto>> getConnectionSuggestions(GetConnectionSuggestionsParams params);
```
nquinquenel, line 138: We usually define a Response object instead of sending raw types such as List, this is usually better to structure the code, and in case of refactoring. Furthermore, I don't think we have the practice of returning a DTO. A DTO is supposed to represent some data that we transmit over the different services of our application. As a response of our APIs, we usually define Response object that defines what we exactly want to return from this endpoint. Currently, the DTO and the Response object might be similar, but in the future, the DTO might grow while the response should not.

backend/rpc-impl/src/main/java/org/sonarsource/sonarlint/core/rpc/impl/ConnectionRpcServiceDelegate.java
```diff
@@ -104,4 +108,12 @@ public CompletableFuture<GetProjectNamesByKeyResponse> getProjectNamesByKey(GetP
     return requestAsync(cancelMonitor -> new GetProjectNamesByKeyResponse(getBean(ConnectionService.class)
       .getProjectNamesByKey(params.getTransientConnection(), params.getProjectKeys(), cancelMonitor)));
   }
+
+  @Override
+  public CompletableFuture<List<ConnectionSuggestionDto>>getConnectionSuggestions(GetConnectionSuggestionsParams params) {
```
nquinquenel, line 113:
```suggestion
  public CompletableFuture<List<ConnectionSuggestionDto>> getConnectionSuggestions(GetConnectionSuggestionsParams params) {
```

backend/core/src/main/java/org/sonarsource/sonarlint/core/ConnectionSuggestionProvider.java
```diff
@@ -107,7 +108,19 @@ private void queueConnectionSuggestion(Set<String> listConfigScopeIds) {
     }
   }
 
+  record ConnectionAndBindingSuggestions(
```
nquinquenel, line 111: I would avoid putting `record` in the middle of the class, could we move it to the end?
Also, should it really have `package` visibility?

backend/core/src/main/java/org/sonarsource/sonarlint/core/ConnectionSuggestionProvider.java
```diff
@@ -107,7 +108,19 @@ private void queueConnectionSuggestion(Set<String> listConfigScopeIds) {
...
+    Map<String, List<ConnectionSuggestionDto>> connectionSuggestionsByConfigScopeIds,
+    Set<String> bindingSuggestionsForConfigScopeIds
+  ) {}
+
   private void suggestConnectionForGivenScopes(Set<String> listOfFilesPerConfigScopeIds, SonarLintCancelMonitor cancelMonitor) {
```
nquinquenel, line 116: I know it was already like this before, but the name `listOfFilesPerConfigScopeIds` doesn't fit the type `Set<String>`. From what I understand, it's just a list of config scope IDs.

backend/core/src/main/java/org/sonarsource/sonarlint/core/ConnectionSuggestionProvider.java
```diff
@@ -107,7 +108,19 @@ private void queueConnectionSuggestion(Set<String> listConfigScopeIds) {
...
+    var connectionAndBindingSuggestions = computeConnectionAndBindingSuggestions(listOfFilesPerConfigScopeIds, cancelMonitor);
+
+    suggestConnectionToClientIfAny(connectionAndBindingSuggestions.connectionSuggestionsByConfigScopeIds());
+    computeBindingSuggestionIfAny(connectionAndBindingSuggestions.bindingSuggestionsForConfigScopeIds());
```
nquinquenel, line 120: So the method name is "suggest connections", but the actual content will trigger suggestions for both connections and bindings. I know it was like this before but probably we could update the name for clarity?

backend/core/src/main/java/org/sonarsource/sonarlint/core/ConnectionSuggestionProvider.java
```diff
@@ -176,4 +187,10 @@ private void computeBindingSuggestionIfAny(Set<String> bindingSuggestionsForConf
     }
   }
 
+  public List<ConnectionSuggestionDto> getConnectionSuggestions(String configScopeId, SonarLintCancelMonitor cancelMonitor) {
+    var connectionAndBindigSuggestions = computeConnectionAndBindingSuggestions(Set.of(configScopeId), cancelMonitor);
```
nquinquenel, line 191:
```suggestion
    var connectionAndBindingSuggestions = computeConnectionAndBindingSuggestions(Set.of(configScopeId), cancelMonitor);
```

nquinquenel (approved):
LGTM

PR #1370, opened 2025-06-03
SLCORE-1372 Drop branch check in open in IDE scenarios

eray-felek-sonarsource:
[SLCORE-1372](https://sonarsource.atlassian.net/browse/SLCORE-1372)

[SLCORE-1372]: https://sonarsource.atlassian.net/browse/SLCORE-1372?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/core/src/test/java/org/sonarsource/sonarlint/core/embedded/server/ShowFixSuggestionRequestHandlerTests.java
```diff
@@ -234,90 +234,6 @@ void should_validate_fix_suggestion_query_for_sc() {
...
-    await().atMost(10, TimeUnit.SECONDS).untilAsserted(() -> verify(sonarLintRpcClient).showMessage(showMessageArgumentCaptor.capture()));
-    assertThat(showMessageArgumentCaptor.getValue().getType()).isEqualTo(MessageType.ERROR);
-    assertThat(showMessageArgumentCaptor.getValue().getText()).isEqualTo("Attempted to show a fix suggestion from branch 'branch', " +
-      "which is different from the currently checked-out branch.\nPlease switch to the correct branch and try again.");
-    verify(sonarLintRpcClient).matchProjectBranch(any());
-    verifyNoMoreInteractions(sonarLintRpcClient);
-  }
-
-  @Test
-  void should_find_main_branch_when_not_provided_and_not_stored() throws HttpException, IOException {
```
sophio-japharidze-sonarsource, line 282: I think this test should be kept?

sophio-japharidze-sonarsource (approved):
LGTM, please see my minor comment

nquinquenel:
I don't think the client method `matchProjectBranch` has any other purpose now. I think it should be marked as `Deprecated` and removed at some point in the future.

Also, some class attributes should be removed, I don't think `sonarProjectBranchesSynchronizationService` has any utility.

eray-felek-sonarsource:
> I don't think the client method `matchProjectBranch` has any other purpose now. I think it should be marked as `Deprecated` and removed at some point in the future.
> 
> Also, some class attributes should be removed, I don't think `sonarProjectBranchesSynchronizationService` has any utility.

Not sure if we can remove sonarProjectBranchesSynchronizationService on ShowIssueRequestHandler. fetchServerIssue still relies on the branch information provided on the API call for ShowIssueRequestHandler. However we can remove the attribute for ShowFixSuggestionRequestHandler

Yes I think we should deprecate matchProjectBranch

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/SonarLintRpcClient.java
```diff
@@ -210,9 +210,10 @@ default void showFixSuggestion(ShowFixSuggestionParams params) {
   CompletableFuture<MatchSonarProjectBranchResponse> matchSonarProjectBranch(MatchSonarProjectBranchParams params);
 
   /**
-   * Used for checking whether a locally checked out branch matches a candidate branch name (not necessarily a Sonar branch).
-   * For example, in "show fix suggestion" use-case, to match a local branch with a PR branch that originated a fix suggestion
+   * @deprecated avoid calling this method if possible, branch matching is not done anymore
+   * and client should rely on notifications when there is code mismatch
    */
+  @Deprecated
```
nquinquenel: Clients can't call this method, it's SLCORE who triggers it.

Actually, I think it would even make sense to simply remove this method. There's no replacement needed so it should be a no-brainer for the clients.

Should be documented in the API CHANGES file

nquinquenel (approved):
LGTM

PR #1371, opened 2025-06-05
SLCORE-1406 Consider AnalyzerConfigurationStorage invalid if file doesn't exist

kirill-knize-sonarsource:
[SLCORE-1312](https://sonarsource.atlassian.net/browse/SLCORE-1312)

Proposal is to not log full exception since it ends up being captured by sentry. To reduce the noise we keep logging the fact storage was not available, but without reporting to sentry.
For more generic exception we still report to sentry to be aware about other types of issues here.

[SLCORE-1312]: https://sonarsource.atlassian.net/browse/SLCORE-1312?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/AnalyzerConfigurationStorage.java
+++ b/backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/AnalyzerConfigurationStorage.java
@@ -19,6 +19,7 @@
  */
 package org.sonarsource.sonarlint.core.serverconnection;
 
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Map;
 import java.util.Optional;
@@ -45,6 +46,10 @@ public class AnalyzerConfigurationStorage {
   }
 
   public boolean isValid() {
+    if (!Files.exists(storageFilePath)) {
+      LOG.debug("Analyzer configuration storage doesn't exist: {}", storageFilePath);
+      return false;
+    }
     return tryRead().isPresent();
   }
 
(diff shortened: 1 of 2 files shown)


damien-urruty-sonarsource:
@kirill-knize-sonarsource I am not sure I understand the point of the changes:
* the commit message says "investigate" but we are taking some other action
* the ticket description says we are logging at error level, but it doesn't seem true, just after the code you added there was a catch for Exceptions, logging at the debug level
* the change you bring does not change a lot of things, except that we now log the file name, and not the complete error. But we still log at the debug level.

Could you give me some clarification so I can make the review?

kirill-knize-sonarsource:
https://github.com/SonarSource/sonarlint-core/blob/e28ab4daebb6645d1e118529d171d99e88f6f554/backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/log/SonarLintLogger.java#L152

It's how our sentry reporting works. If we pass non-null exception object to the logger (for both Warn and Error log levels), it will be captured by sentry client and reported to the server.
So we have three options:
1. Change the log level to say Debug, where we are not supposed to pass the exception.
2. Keep current log level but don't pass the exception to avoid sentry reporting - it's the proposed option.
3. Also we can exclude this event by configuring sentry client by exception type or by the string pattern as it was done for parsing exception coming from Python analyzer:
https://github.com/SonarSource/sonarlint-core/blob/e28ab4daebb6645d1e118529d171d99e88f6f554/backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/monitoring/MonitoringService.java#L75

Please tell me if some other context is needed.

PR #1375, opened 2025-06-07
SLCORE-1242 Trace "overhead" operations on analysis requests

nquinquenel:
[SLCORE-1242](https://sonarsource.atlassian.net/browse/SLCORE-1242)

[SLCORE-1242]: https://sonarsource.atlassian.net/browse/SLCORE-1242?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/command/AnalyzeCommand.java
+++ b/backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/command/AnalyzeCommand.java
@@ -36,6 +36,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
+import javax.annotation.CheckForNull;
 import javax.annotation.Nullable;
 import org.sonarsource.sonarlint.core.analysis.api.AnalysisConfiguration;
 import org.sonarsource.sonarlint.core.analysis.api.AnalysisResults;
@@ -50,6 +51,7 @@ import org.sonarsource.sonarlint.core.commons.progress.ProgressIndicator;
 import org.sonarsource.sonarlint.core.commons.progress.SonarLintCancelMonitor;
 import org.sonarsource.sonarlint.core.commons.progress.TaskManager;
 
+import static org.sonarsource.sonarlint.core.commons.monitoring.Trace.startChild;
 import static org.sonarsource.sonarlint.core.commons.util.StringUtils.pluralize;
 
 public class AnalyzeCommand extends Command {
@@ -167,8 +169,8 @@ public class AnalyzeCommand extends Command {
     analysisStarted.accept(configuration.inputFiles());
     var moduleContainer = moduleRegistry.getContainerFor(moduleKey);
     if (moduleContainer == null) {
-      // if not found, means we are outside any module (e.g. single file analysis on VSCode)
-      moduleContainer = moduleRegistry.createTransientContainer(configuration.inputFiles());
+      moduleContainer = startChild(trace, "createTransientContainer", null,
+        () -> moduleRegistry.createTransientContainer(configuration.inputFiles()));
     }
     Throwable originalException = null;
     doIfTraceIsSet(t -> {
@@ -205,7 +207,7 @@ public class AnalyzeCommand extends Command {
       });
       result.setDuration(Duration.ofMillis(System.currentTimeMillis() - startTime));
       return result;
-    } catch (Throwable e) {
+    } catch (Exception e) {
       originalException = e;
       doIfTraceIsSet(t -> t.finishExceptionally(e));
       throw e;
@@ -265,5 +267,10 @@ public class AnalyzeCommand extends Command {
     var filesMatch = Objects.equals(getFiles(), analyzeCommand.getFiles());
     var extraPropertiesMatch = Objects.equals(getExtraProperties(), analyzeCommand.getExtraProperties());
     return triggerTypesMatch && filesMatch && extraPropertiesMatch;
+  }
+
+  @CheckForNull
+  public Trace getTrace() {
+    return trace;
   }
 }
--- a/backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/command/AnalyzeCommand.java
+++ b/backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/command/AnalyzeCommand.java
@@ -207,7 +207,7 @@ public class AnalyzeCommand extends Command {
       });
       result.setDuration(Duration.ofMillis(System.currentTimeMillis() - startTime));
       return result;
-    } catch (Exception e) {
+    } catch (Throwable e) {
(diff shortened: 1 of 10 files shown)


backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/command/AnalyzeCommand.java
```diff
@@ -205,7 +207,7 @@ private AnalysisResults doRunAnalysis(ModuleRegistry moduleRegistry, ProgressInd
       });
       result.setDuration(Duration.ofMillis(System.currentTimeMillis() - startTime));
       return result;
-    } catch (Throwable e) {
+    } catch (Exception e) {
```
damien-urruty-sonarsource: I know this is raised by SQ but it was done on purpose. The nature of plugins is that they might try to load things at runtime, and some `Error`s can occur. I think it would be better to keep catching Throwables

backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/container/module/ModuleContainer.java
```diff
@@ -63,16 +65,17 @@ public boolean isTransient() {
   }
 
   public AnalysisResults analyze(AnalysisConfiguration configuration, Consumer<Issue> issueListener, ProgressIndicator progressIndicator, @Nullable Trace trace) {
-    var analysisContainer = new AnalysisContainer(this, progressIndicator);
+    var analysisContainer = startChild(trace, "newAnalysisContainer", "analyze", () -> new AnalysisContainer(this, progressIndicator));
     analysisContainer.add(configuration);
     analysisContainer.add(new IssueListenerHolder(issueListener));
-    analysisContainer.add(new ActiveRulesAdapter(configuration.activeRules().stream().map(ActiveRuleAdapter::new).toList()));
+    analysisContainer.add(startChild(trace, "newActiveRulesAdapter", "analyze", () ->
+      new ActiveRulesAdapter(configuration.activeRules().stream().map(ActiveRuleAdapter::new).toList())));
```
damien-urruty-sonarsource, line 72: I'm not sure measuring this is very useful, but it's just a gut feeling, let's see the figures

backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisSchedulerCache.java
```diff
@@ -116,12 +123,16 @@ private synchronized AnalysisScheduler getStandaloneSchedulerIfStarted() {
...
+    return new AnalysisScheduler(createSchedulerConfiguration(actualCsharpAnalyzerPath, trace), plugins, SonarLintLogger.get().getTargetForCopy());
   }
 
   private AnalysisSchedulerConfiguration createSchedulerConfiguration(@Nullable Path actualCsharpAnalyzerPath) {
-    var activeNodeJs = nodeJsService.getActiveNodeJs();
+    return createSchedulerConfiguration(actualCsharpAnalyzerPath, null);
+  }
+
+  private AnalysisSchedulerConfiguration createSchedulerConfiguration(@Nullable Path actualCsharpAnalyzerPath, @Nullable Trace trace) {
+    var activeNodeJs = startChild(trace, "getActiveNodeJs", "createSchedulerConfiguration", nodeJsService::getActiveNodeJs);
```
damien-urruty-sonarsource, line 135: I am not sure this measure would always be relevant because there is caching in `getActiveNodeJs`. I guess the goal is to measure how much time it takes to detect Node.js, could we measure that at another location?

backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisService.java
```diff
@@ -287,14 +294,25 @@ public GetAnalysisConfigResponse getAnalysisConfig(String configScopeId, boolean
...
     })
-      .orElseGet(() -> new GetAnalysisConfigResponse(buildStandaloneActiveRules(), userAnalysisProperties, nodeJsDetailsDto,
-        Set.copyOf(pluginsService.getEmbeddedPluginPaths())));
+      .orElseGet(() -> {
+        var standaloneActiveRules = startChild(trace, "buildStandaloneActiveRules", GET_ANALYSIS_CFG,
+          this::buildStandaloneActiveRules);
+        var embeddedPluginPaths = startChild(trace, "getEmbeddedPluginPaths", GET_ANALYSIS_CFG,
+          pluginsService::getEmbeddedPluginPaths);
+        return new GetAnalysisConfigResponse(standaloneActiveRules, userAnalysisProperties, nodeJsDetailsDto,
+          Set.copyOf(embeddedPluginPaths));
```
damien-urruty-sonarsource: This method has become quite verbose and hard to read, but I don't have any idea to simplify it. The `GET_ANALYSIS_CFG` repeats a lot of time so maybe we could introduce an intermediate object

backend/plugin-commons/src/main/java/org/sonarsource/sonarlint/core/plugin/commons/container/SpringComponentContainer.java
```diff
@@ -175,9 +182,9 @@ public void execute() {
 
   @Override
   public SpringComponentContainer startComponents() {
-    doBeforeStart();
-    context.refresh();
-    doAfterStart();
+    startChild(trace, "doBeforeStart", START_COMPONENTS, this::doBeforeStart);
+    startChild(trace, "refresh", START_COMPONENTS, context::refresh);
+    startChild(trace, "doAfterStart", START_COMPONENTS, this::doAfterStart);
     return this;
```
damien-urruty-sonarsource: What about something like:

```
trace.newChild("startComponents", 
  step("doBeforeStart", this::doBeforeStart),
  step("refresh", context::refresh),
  step("doAfterStart", this::doAfterStart));
```

damien-urruty-sonarsource (approved):
Overall I find it introduces a lot of verbosity. I'm not sure how we can do shorter though. Otherwise LGTM, I think we could go further and compute the overhead of issue tracking and streaming

backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/command/AnalyzeCommand.java
nquinquenel: Good to know, thanks!

backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisSchedulerCache.java
nquinquenel: I'd like to also avoid passing this `Trace` in every single method, I think as a first step it's fine 🤷🏻 
When I checked this metric, it was something very fast (caching used I guess), and sometimes slower. We should be able to manually differentiate the two for the time being

PR #1376, opened 2025-06-07
SLCORE-1401 Fix SSF-808

nquinquenel:


kirill-knize-sonarsource (approved):
It's a bit sad we don't test delays here, but it will add too much tests execution time.

PR #1377, opened 2025-06-07
SLCORE-1403 Avoid NPE when repository has no commits

nquinquenel:


kirill-knize-sonarsource:
New code is not covered. It looks like we need to setup an empty bare repo to test it.

kirill-knize-sonarsource (changes requested):
New code is not covered. It looks like we need to setup an empty bare repo to test it.

kirill-knize-sonarsource (approved):
LGTM!

PR #1378, opened 2025-06-10
SLCORE-1406 Consider AnalyzerConfigurationStorage invalid if file doesn't exist

kirill-knize-sonarsource:
[SLCORE-1406](https://sonarsource.atlassian.net/browse/SLCORE-1406)


damien-urruty-sonarsource:
@kirill-knize-sonarsource I just noticed the same logic exists in `PluginsStorage`, could you apply the same fix?

PR #1382, opened 2025-06-17
SLCORE-1397 Give clients possibility to record "Hotspot investigated" as telemetry measure

kirill-knize-sonarsource:
[SLCORE-1397](https://sonarsource.atlassian.net/browse/SLCORE-1397)

[SLCORE-1397]: https://sonarsource.atlassian.net/browse/SLCORE-1397?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/measures/payload/TelemetryMeasuresBuilder.java
+++ b/backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/measures/payload/TelemetryMeasuresBuilder.java
@@ -60,6 +60,8 @@ public class TelemetryMeasuresBuilder {
 
     addPerformanceMeasures(values);
 
+    addIssueInvestigationMeasures(values);
+
     return new TelemetryMeasuresPayload(UUID.randomUUID().toString(), platform, storage.installTime(), product, TelemetryMeasuresDimension.INSTALLATION, values);
   }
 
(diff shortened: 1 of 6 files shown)


backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/measures/payload/TelemetryMeasuresBuilder.java
```diff
@@ -60,6 +60,8 @@ public TelemetryMeasuresPayload build() {
 
     addPerformanceMeasures(values);
 
+    addIssueInvestigationMeasures(values);
```
sophio-japharidze-sonarsource: Let's call it `addFindingInvestigationMeasures`

sophio-japharidze-sonarsource (commented):
Added one small naming suggestion. AND, more importantly we should update API_CHANGES doc

backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/measures/payload/TelemetryMeasuresBuilder.java
kirill-knize-sonarsource: Changed.

PR #1383, opened 2025-06-18
SLCORE-1429 Ensure open file references are not duplicated

damien-urruty-sonarsource:
[SLCORE-1429](https://sonarsource.atlassian.net/browse/SLCORE-1429)

[SLCORE-1429]: https://sonarsource.atlassian.net/browse/SLCORE-1429?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/rpc-impl/src/main/java/org/sonarsource/sonarlint/core/rpc/impl/AnalysisRpcServiceDelegate.java
+++ b/backend/rpc-impl/src/main/java/org/sonarsource/sonarlint/core/rpc/impl/AnalysisRpcServiceDelegate.java
@@ -20,6 +20,7 @@
 package org.sonarsource.sonarlint.core.rpc.impl;
 
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import org.eclipse.lsp4j.jsonrpc.ResponseErrorException;
 import org.eclipse.lsp4j.jsonrpc.messages.ResponseError;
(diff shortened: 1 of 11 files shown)


medium-tests/src/test/java/mediumtest/analysis/AnalysisTriggeringMediumTests.java
```diff
@@ -264,6 +264,39 @@ void it_should_analyze_open_files_when_re_enabling_automatic_analysis(SonarLintT
         .containsExactly("Replace \"pom.version\" with \"project.version\"."));
   }
 
+  @SonarLintTest
+  void it_should_analyze_open_files_when_re_enabling_automatic_analysis_when_same_file_opened_twice(SonarLintTestHarness harness, @TempDir Path baseDir) {
```
georgii-borovinskikh-sonarsource, line 269: Not sure if this should be a medium test or a unit test, but one additional scenario we could cover is when the 2nd didOpenFile happens after the analysis (to test that it doesn't produce the 2nd event)

backend/rpc-impl/src/main/java/org/sonarsource/sonarlint/core/rpc/impl/AnalysisRpcServiceDelegate.java
```diff
@@ -137,8 +138,9 @@ public CompletableFuture<GetAutoDetectedNodeJsResponse> getAutoDetectedNodeJs()
   @Override
   public CompletableFuture<AnalyzeFilesResponse> analyzeFilesAndTrack(AnalyzeFilesAndTrackParams params) {
     var configurationScopeId = params.getConfigurationScopeId();
-    return requestFutureAsync(cancelChecker -> getBean(AnalysisService.class).scheduleAnalysis(params.getConfigurationScopeId(), params.getAnalysisId(), params.getFilesToAnalyze(),
-      params.getExtraProperties(), params.isShouldFetchServerIssues(), TriggerType.FORCED_WITH_EXCLUSIONS, cancelChecker)
+    return requestFutureAsync(cancelChecker -> getBean(AnalysisService.class)
+      .scheduleAnalysis(params.getConfigurationScopeId(), params.getAnalysisId(), Set.copyOf(params.getFilesToAnalyze()),
```
georgii-borovinskikh-sonarsource, line 142: I notice there are calls to `Set.copyOf` in some places where originally the list was not copied. Can you explain the reason why? I don't know the specifics of java collections too well

georgii-borovinskikh-sonarsource (approved):

backend/rpc-impl/src/main/java/org/sonarsource/sonarlint/core/rpc/impl/AnalysisRpcServiceDelegate.java
damien-urruty-sonarsource: It's one concise way to convert a List to a Set in Java. The other ways I can think of are more verbose

medium-tests/src/test/java/mediumtest/analysis/AnalysisTriggeringMediumTests.java
damien-urruty-sonarsource: Good idea! I added a test. In case you want to have another look, I will merge the PR at the beginning of the afternoon

georgii-borovinskikh-sonarsource: All good from my side

PR #1385, opened 2025-06-19
SLCORE-907 Skip BOM when reading file from disk

damien-urruty-sonarsource:
[SLCORE-907](https://sonarsource.atlassian.net/browse/SLCORE-907)

[SLCORE-907]: https://sonarsource.atlassian.net/browse/SLCORE-907?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/core/src/main/java/org/sonarsource/sonarlint/core/fs/ClientFile.java
```diff
@@ -112,14 +117,33 @@ public String getContent() {
...
       throw new IllegalStateException("Unable to read file " + fsPath + "content with charset " + charsetToUse, e);
     }
   }
 
+  public InputStream inputStream() {
+    if (isDirty && clientProvidedContent != null) {
+      return new ByteArrayInputStream(clientProvidedContent.getBytes(getCharset()));
+    }
+    if (fsPath == null) {
+      throw new IllegalStateException("File " + uri + " is not dirty but has no OS Path defined");
```
kirill-knize-sonarsource: This message is not exactly true, because we can get here when file may be dirty but clientProvidedContent is null.

backend/core/src/main/java/org/sonarsource/sonarlint/core/fs/ClientFile.java
```diff
@@ -112,14 +117,33 @@ public String getContent() {
...
+    }
+    try {
+      return BOMInputStream.builder().setInputStream(Files.newInputStream(fsPath))
+        .setByteOrderMarks(ByteOrderMark.UTF_8, ByteOrderMark.UTF_16LE, ByteOrderMark.UTF_16BE, ByteOrderMark.UTF_32LE, ByteOrderMark.UTF_32BE).get();
```
kirill-knize-sonarsource: Constructor enforces the sorting in descending order (by byte length). So if you want to save some CPU clock cycles for every creation of such object, you can provide them in a reversed order right away.

backend/core/src/main/java/org/sonarsource/sonarlint/core/fs/ClientFile.java
damien-urruty-sonarsource: True. What about:
> throw new IllegalStateException("File " + uri + " is not dirty or does not have content but has no OS Path defined");

backend/core/src/main/java/org/sonarsource/sonarlint/core/fs/ClientFile.java
damien-urruty-sonarsource: Why not yes, we assume things about the underlying implementation (which might change), but it doesn't cost much to adjust later if needed

PR #1386, opened 2025-06-20
SLCORE-1431 Send the count of sub-projects bound for telemetry

nquinquenel:
[SLCORE-1431](https://sonarsource.atlassian.net/browse/SLCORE-1431)

[SLCORE-1431]: https://sonarsource.atlassian.net/browse/SLCORE-1431?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/core/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryServerAttributesProvider.java
+++ b/backend/core/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryServerAttributesProvider.java
@@ -58,7 +58,7 @@ public class TelemetryServerAttributesProvider {
     var usesConnectedMode = !allBindings.isEmpty();
     var usesSonarCloud = allBindings.stream().anyMatch(isSonarCloudConnectionConfiguration());
 
-    var childBindingCount = countChildBindings(allBindings);
+    var childBindingCount = countChildBindings();
     var sonarQubeServerBindingCount = countSonarQubeServerBindings(allBindings);
     var sonarQubeCloudEUBindingCount = countSonarQubeCloudBindings(allBindings, SonarCloudRegion.EU);
     var sonarQubeCloudUSBindingCount = countSonarQubeCloudBindings(allBindings, SonarCloudRegion.US);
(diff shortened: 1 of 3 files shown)


backend/core/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryServerAttributesProvider.java
```diff
@@ -108,6 +105,12 @@ private int countSonarQubeServerBindings(Collection<BoundScope> allBindings) {
       .count();
   }
 
+  private int countChildBindings(Collection<BoundScope> allBindings) {
+    return (int) allBindings.stream()
+      .filter(scope -> configurationRepository.getParentId(scope.getConfigScopeId()).isPresent())
+      .count();
+  }
```
damien-urruty-sonarsource: Just a note, In VSCode a root config scope is created, and then a config scope for each folder is created as a child of the root. The count might be non-zero, but I don't think it's a problem

backend/core/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryServerAttributesProvider.java
```diff
@@ -108,6 +105,12 @@ private int countSonarQubeServerBindings(Collection<BoundScope> allBindings) {
       .count();
   }
 
+  private int countChildBindings(Collection<BoundScope> allBindings) {
+    return (int) allBindings.stream()
+      .filter(scope -> configurationRepository.getParentId(scope.getConfigScopeId()).isPresent())
```
damien-urruty-sonarsource: Actually, I think this is not sufficient. If you check the implementation of `ConfigurationRepository#getAllBoundScopes`, you will see that all config scopes having an "effective binding" will be returned. And the effective binding is something you can inherit from the parent, so I think you would also count modules not having a override

backend/core/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryServerAttributesProvider.java
nquinquenel: Ah in this case that's an issue. We want to know whether people override bindings in their projects, one easy way is to check whether the value is >= 1.  Getting the actual count is just a "plus" to assess how many overrides people do.

PR #1387, opened 2025-06-20
SLCORE-1391 Gather speed-related metrics on analysis traces

jblievremont:


jblievremont:
Oops sorry, asked for review a bit too soon - force-pushed after manual testing. Should be alright now!

backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/command/AnalyzeCommand.java
```diff
@@ -167,27 +170,43 @@ private AnalysisResults doRunAnalysis(ModuleRegistry moduleRegistry, ProgressInd
...
     Throwable originalException = null;
-    if (trace != null) {
-      trace.setData("filesCount", configuration.inputFiles().size());
-      trace.setData("languages", configuration.inputFiles().stream()
-        .map(ClientInputFile::language)
-        .filter(Objects::nonNull)
-        .map(SonarLanguage::getSonarLanguageKey)
-        .toList());
-    }
+    doIfTraceIsSet(t -> {
```
damien-urruty-sonarsource, line 174: Having a method does not bring much, but it's more of a personal taste

backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/command/AnalyzeCommand.java
```diff
@@ -167,27 +170,43 @@ private AnalysisResults doRunAnalysis(ModuleRegistry moduleRegistry, ProgressInd
...
+      int filesCount = configuration.inputFiles().size();
+      t.setData("filesCount", filesCount);
+      var fileSizes = new ArrayList<>(filesCount);
+      var languages = new  ArrayList<>(filesCount);
+      configuration.inputFiles().forEach(f -> {
+        try {
+          fileSizes.add(f.contents().length());
```
damien-urruty-sonarsource, line 181: This is something I noticed when fixing the BOM issue, we may end up reading the same file many times, there is no caching. I think there is room for improvement, but I will create a separate ticket for that.

backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/command/AnalyzeCommand.java
```diff
@@ -167,27 +170,43 @@ private AnalysisResults doRunAnalysis(ModuleRegistry moduleRegistry, ProgressInd
...
-        .map(ClientInputFile::language)
-        .filter(Objects::nonNull)
-        .map(SonarLanguage::getSonarLanguageKey)
-        .toList());
-    }
+    doIfTraceIsSet(t -> {
+      int filesCount = configuration.inputFiles().size();
+      t.setData("filesCount", filesCount);
+      var fileSizes = new ArrayList<>(filesCount);
+      var languages = new  ArrayList<>(filesCount);
```
damien-urruty-sonarsource: Could this be a Set instead?

backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/command/AnalyzeCommand.java
jblievremont: I see it as a way to avoid sprinkling `null`-checks across the code.

backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/command/AnalyzeCommand.java
jblievremont: Yup; this being said, this should be called only when tracing is enabled, so it may not be too much of a performance hit.

PR #1392, opened 2025-06-27
SLCORE-1504 Track SQC account creation through token generation

jblievremont:


rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/connection/auth/HelpGenerateUserTokenParams.java
```diff
@@ -20,14 +20,59 @@
...
+import javax.annotation.CheckForNull;
+import javax.annotation.Nullable;
+
 public class HelpGenerateUserTokenParams {
   private final String serverUrl;
 
+  private final Utm utm;
+
   public HelpGenerateUserTokenParams(String serverUrl) {
+    this(serverUrl, null);
```
damien-urruty-sonarsource, line 33: Is the goal that all clients implement the "feature"? If so, you could deprecate this constructor

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/connection/auth/HelpGenerateUserTokenParams.java
```diff
@@ -20,14 +20,59 @@
...
   }
+
+  @CheckForNull
+  public Utm getUtm() {
+    return utm;
+  }
+
+  public static class Utm {
+    private final String type;
+    private final String source;
```
damien-urruty-sonarsource, line 52: We have this information in many different forms in the backend, for telemetry or for the embedded server's status. But maybe it's safer to have it here too if they need to differ at some point

jblievremont: Yes, the values passed here will be used for marketing purposes, so I believe it is preferable to isolate them from accidental changes.

PR #1393, opened 2025-06-30
SLCORE-1505 Do not refresh path translations when config scope is removed

damien-urruty-sonarsource:
[SLCORE-1505](https://sonarsource.atlassian.net/browse/SLCORE-1505)

[SLCORE-1505]: https://sonarsource.atlassian.net/browse/SLCORE-1505?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/core/src/main/java/org/sonarsource/sonarlint/core/file/PathTranslationService.java
+++ b/backend/core/src/main/java/org/sonarsource/sonarlint/core/file/PathTranslationService.java
@@ -88,7 +88,7 @@ public class PathTranslationService {
 
   @EventListener
   public void onConfigurationScopeRemoved(ConfigurationScopeRemovedEvent event) {
-    cachedPathsTranslationByConfigScope.refreshAsync(event.getRemovedConfigurationScopeId());
+    cachedPathsTranslationByConfigScope.clear(event.getRemovedConfigurationScopeId());
   }
 
   @EventListener


backend/core/src/main/java/org/sonarsource/sonarlint/core/file/PathTranslationService.java
```diff
@@ -88,7 +88,7 @@ private FilePathTranslation matchPaths(String configScopeId, FileTreeMatcher fil
 
   @EventListener
   public void onConfigurationScopeRemoved(ConfigurationScopeRemovedEvent event) {
-    cachedPathsTranslationByConfigScope.refreshAsync(event.getRemovedConfigurationScopeId());
+    cachedPathsTranslationByConfigScope.clear(event.getRemovedConfigurationScopeId());
```
kirill-knize-sonarsource, line 91: Do we want to test it? There's no test that fails without this change.

damien-urruty-sonarsource: It's not so easy to test, even with a unit test. It would force me to modify the implementation further by exposing things that would be used from tests only. I'm confident that the change won't break anything, and also given the short time frame until the release, would you consider merging it without test? There are other things I would like to focus on

kirill-knize-sonarsource (approved):

PR #1394, opened 2025-06-30
SLCORE-1424 Trigger analysis of open files on User Analysis Properties change

kirill-knize-sonarsource:


backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisService.java
```diff
@@ -327,6 +327,7 @@ private static boolean isPrefixForAll(Path prefixCandidate, Collection<Path> pat
 
   public void setUserAnalysisProperties(String configScopeId, Map<String, String> properties) {
     userAnalysisPropertiesRepository.setUserProperties(configScopeId, properties);
+    autoAnalyzeOpenFiles(configScopeId);
```
damien-urruty-sonarsource: I've had a discussion recently about clients sometimes calling methods even if no data changed. To be on the safe side, could you re-trigger the analysis only if the properties really changed?

backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/UserAnalysisPropertiesRepository.java
```diff
@@ -40,11 +39,23 @@ public Map<String, String> getUserProperties(String configurationScopeId) {
...
-  public void setUserProperties(String configurationScopeId, Map<String, String> extraProperties) {
-    propertiesByConfigScope.put(configurationScopeId, new HashMap<>(extraProperties));
+  public boolean setUserProperties(String configurationScopeId, Map<String, String> extraProperties) {
+    var oldProperties = propertiesByConfigScope.get(configurationScopeId);
+    var newProperties = new HashMap<>(extraProperties);
+    var changed = !newProperties.equals(oldProperties);
+    if (changed) {
+      propertiesByConfigScope.put(configurationScopeId, newProperties);
+    }
+    return changed;
```
damien-urruty-sonarsource: I am not sure there is a need to create a new HashMap. What about something a bit simpler: 

```
var oldProperties = propertiesByConfigScope.put(configurationScopeId, extraProperties);
return !extraProperties.equals(oldProperties);
```

This should also play nicer with concurrent access than get and put in 2 different steps.

medium-tests/src/test/java/mediumtest/analysis/AnalysisMediumTests.java
```diff
@@ -1181,6 +1181,86 @@ void it_should_analyze_xml_file_in_non_ascii_directory_and_non_decoded_uri(Sonar
     assertThat(raisedIssueDto).isNotEmpty();
   }
 
+  @SonarLintTest
+  void it_should_trigger_analysis_after_analysis_properties_change(SonarLintTestHarness harness, @TempDir Path baseDir) {
```
damien-urruty-sonarsource: There is another test class that is called `AnalysisTriggeringMediumTests`, more suited for tests around triggering, I think it would be a better fit.

backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/UserAnalysisPropertiesRepository.java
```diff
@@ -40,11 +39,23 @@ public Map<String, String> getUserProperties(String configurationScopeId) {
     return properties;
   }
 
-  public void setUserProperties(String configurationScopeId, Map<String, String> extraProperties) {
-    propertiesByConfigScope.put(configurationScopeId, new HashMap<>(extraProperties));
+  public boolean setUserProperties(String configurationScopeId, Map<String, String> extraProperties) {
```
damien-urruty-sonarsource: Nitpick: `extraProperties` -> `userProperties`? The "extra" was a bit confusing

damien-urruty-sonarsource (approved):
LGTM, I left a couple ideas for improvement. I think you will have to rebase

backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/UserAnalysisPropertiesRepository.java
kirill-knize-sonarsource: It will throw about immutable collection in `getUserProperties()` then. We put path to compile commands to this map. Am I wrong?

medium-tests/src/test/java/mediumtest/analysis/AnalysisMediumTests.java
kirill-knize-sonarsource: Changed.

backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/UserAnalysisPropertiesRepository.java
kirill-knize-sonarsource: Renamed.

backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/UserAnalysisPropertiesRepository.java
damien-urruty-sonarsource: That's correct. But if anything, I would change `getUserProperties()` to create a new `HashMap`. It's a detail, you could also keep adding a new map to propertiesByConfigScope, but still simplify the body. Something like: 

```
var oldProperties = propertiesByConfigScope.put(configurationScopeId, new HashMap(extraProperties));
return !extraProperties.equals(oldProperties);
```

PR #1398, opened 2025-07-02
SLCORE-1393 Make server settings available when loading rules

damien-urruty-sonarsource:


damien-urruty-sonarsource:
I tested this PR in integration with the changes from CFamily and I was able to detect MISRA issues:
image

jblievremont (approved):
Nice! Tested with in SLVSCODE in connected mode with Peach.

PR #1399, opened 2025-07-02
SLCORE-1509 Cancellation should include queued task that are not started

nquinquenel:
[SLCORE-1509](https://sonarsource.atlassian.net/browse/SLCORE-1509)

[SLCORE-1509]: https://sonarsource.atlassian.net/browse/SLCORE-1509?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisQueue.java
+++ b/backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisQueue.java
@@ -63,19 +63,6 @@ public class AnalysisQueue {
     return pendingTasks.stream().map(QueuedCommand::getCommand).toList();
   }
 
-  public synchronized void remove(String analysisId) {
-    var optQueuedCommand = queue.stream()
-      .filter(qc -> qc.getCommand() instanceof AnalyzeCommand)
-      .filter(qc -> ((AnalyzeCommand) qc.getCommand()).getAnalysisId().toString().equals(analysisId))
-      .findFirst();
-
-    if (optQueuedCommand.isPresent()) {
-      var queuedCommand = optQueuedCommand.get();
-      queuedCommand.getCommand().cancel();
-      queue.remove(queuedCommand);
-    }
-  }
-
   public synchronized Command takeNextCommand() throws InterruptedException {
     while (true) {
       var firstReadyCommand = pollNextReadyCommand();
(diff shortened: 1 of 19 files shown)


backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisScheduler.java
```diff
@@ -63,6 +64,14 @@ public void wakeUp() {
     analysisQueue.wakeUp();
   }
 
+  public void cancelQueuedAnalysis(String analysisId) {
```
kirill-knize-sonarsource: I think that instead directly removing it from the queue, it should the part of the tidyUp() routine. We can add list of cancelled tasks, and on tidyUp() we apply this list and clean it up.

I'm not sure this implementation is a huge crime, but by design housekeeping supposed to happen on tidyUp() and we are spreading this functionality over many places.

nquinquenel: Good to know, I wasn't aware of this, I'll have a look 👀

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/progress/TaskManager.java
```diff
@@ -28,10 +28,20 @@ public class TaskManager {
...
+    createTask(taskId, cancelMonitor);
+    runExistingTask(configurationScopeId, taskId, title, message, indeterminate, cancellable, task, cancelMonitor);
+  }
+
+  public final void runExistingTask(@Nullable String configurationScopeId, UUID taskId, String title, @Nullable String message,
+    boolean indeterminate, boolean cancellable, Task task, SonarLintCancelMonitor cancelMonitor) {
+    var progressMonitor = progressMonitorsByTaskId.get(taskId.toString());
+    if (progressMonitor == null) {
+      return;
+    }
```
damien-urruty-sonarsource: We should maybe display at least a log here, it's not normal to not find the progress monitor in the map

backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisQueue.java
```diff
@@ -85,12 +85,16 @@ private Optional<QueuedCommand> pollNextReadyCommand() {
     // cannot use iterator as priority order is not guaranteed
     while (!queue.isEmpty()) {
       var candidateCommand = queue.poll();
-      if (candidateCommand.command.isReady()) {
-        queue.addAll(commandsToKeep);
-        return Optional.of(candidateCommand);
+      if (candidateCommand.command.isCanceled()) {
+        LOG.debug("Not picking next command {}, is canceled", candidateCommand.command);
```
damien-urruty-sonarsource, line 90: I think you have to call `Command.cancel` at some point, for the analysis future to be canceled. It shows that the design is not great, because if the command is canceled, we should call cancel.... At some point we should maybe have a whiteboard session. Another option is to rely on `SonarLintCancelMonitor.onCancel`.

backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/command/AnalyzeCommand.java
```diff
@@ -96,6 +95,8 @@ public AnalyzeCommand(String moduleKey, UUID analysisId, TriggerType triggerType
     this.files = files;
     this.extraProperties = extraProperties;
     this.futureResult = futureResult;
+
+    taskManager.createTask(analysisId, cancelMonitor);
```
damien-urruty-sonarsource: It works like this but it puzzles me to have a create method that doesn't return anything. Again it's probably a design issue around this tasks/commands

damien-urruty-sonarsource (commented):
There is still some room for improvement in the design, but it should mostly work. The analysis future cancelation is something you should fix

kirill-knize-sonarsource (approved):
Looks good, we can merge it. Noisy cancellation will be fixed in separate PR.

PR #1400, opened 2025-07-02
SLCORE-1302 File exclusion should correctly handle config scopes

nquinquenel:


backend/core/src/main/java/org/sonarsource/sonarlint/core/fs/FileExclusionService.java
```diff
@@ -290,12 +289,7 @@ private List<ClientFile> filterOutExcludedFiles(String configurationScopeId, Pat
 
   @CheckForNull
   private ClientFile findFile(String configScopeId, URI fileUriToAnalyze) {
-    var clientFile = clientFileSystemService.getClientFiles(configScopeId, fileUriToAnalyze);
-    if (clientFile == null) {
-      LOG.error("File to analyze was not found in the file system: {}", fileUriToAnalyze);
```
kirill-knize-sonarsource: Why remove logging? I think it's relevant case to log.

backend/core/src/main/java/org/sonarsource/sonarlint/core/repository/config/ConfigurationRepository.java
```diff
@@ -87,6 +87,20 @@ public BindingConfiguration getBindingConfiguration(String configScopeId) {
     return bindingPerConfigScopeId.get(configScopeId);
   }
 
+  // Check the configuration scope and its parents to know if it's in connected mode
+  public boolean isConnectedMode(String configScopeId) {
+    var currentId = configScopeId;
+    while (currentId != null) {
+      var binding = bindingPerConfigScopeId.get(currentId);
```
kirill-knize-sonarsource: We usually use names like `bindingByConfigScopeId`. It looks like an old naming inconsistency.
So `bindingPerConfigScopeId` -> `bindingByConfigScopeId`
`configScopePerId` -> `configScopeById`

backend/core/src/main/java/org/sonarsource/sonarlint/core/repository/config/ConfigurationScope.java
```diff
@@ -19,40 +19,10 @@
...
-    return bindable;
-  }
-
-  public String getName() {
-    return name;
-  }
+/**
+ * @param name The name of this configuration scope. Used for auto-binding.
+ */
+public record ConfigurationScope(String id, @Nullable String parentId, boolean bindable, String name) {
```
kirill-knize-sonarsource, line 27: Great that you refactoring data classes to records, thanks!

kirill-knize-sonarsource (commented):
One question about removed logging. Rest is good.

backend/core/src/main/java/org/sonarsource/sonarlint/core/fs/FileExclusionService.java
```diff
@@ -319,12 +313,6 @@ private Set<URI> filterOutClientExcludedFiles(String configurationScopeId, Set<U
       .collect(Collectors.toSet());
   }
 
-  private boolean isConnectedMode(String configurationScopeId) {
-    return Optional.ofNullable(configRepo.getBindingConfiguration(configurationScopeId))
```
damien-urruty-sonarsource, line 323: I think the original problem is that `getBindingConfiguration` was used, which returns the binding _configured_ on a given configuration scope. It does not check for binding set on the parent. It would have worked if `getEffectiveBinding` was used

backend/core/src/main/java/org/sonarsource/sonarlint/core/repository/config/ConfigurationRepository.java
```diff
@@ -87,6 +87,20 @@ public BindingConfiguration getBindingConfiguration(String configScopeId) {
     return bindingPerConfigScopeId.get(configScopeId);
   }
 
+  // Check the configuration scope and its parents to know if it's in connected mode
+  public boolean isConnectedMode(String configScopeId) {
```
damien-urruty-sonarsource: I think a more precise name would be `isBound`. And in fact the implementation is very similar to `getEffectiveBinding`, so you could replace the whole body with `return getEffectiveBinding(configScopeId).isPresent();`

backend/core/src/main/java/org/sonarsource/sonarlint/core/repository/config/ConfigurationRepository.java
damien-urruty-sonarsource: It's probably a name I introduced, I often use `per` and `by` interchangeably, but yes we could settle on a "rule"

backend/core/src/main/java/org/sonarsource/sonarlint/core/fs/FileExclusionService.java
```diff
@@ -305,7 +299,7 @@ private void logFilteredURIs(String reason, ArrayList<URI> uris) {
   }
 
   private Set<URI> filterOutClientExcludedFiles(String configurationScopeId, Set<URI> files) {
-    if (isConnectedMode(configurationScopeId)) {
+    if (configRepo.isConnectedMode(configurationScopeId)) {
```
damien-urruty-sonarsource: From what I understand with this change, before we were wrongly applying client-defined file exclusions on bound IntelliJ modules. Now with your change we won't apply client-defined file exclusions on bound modules. I think it's great, it's more consistent, but I don't see the link with what you described in the ticket:

> I have a very standard project structure in IntelliJ and it does not respect the server exclusions
> When the analysis is made, it is done on the module, so when checking the server exclusions, it does not find any matching on the code files.

This 2nd sentence is not very clear for me, but I have the impression your problem is about server-side exclusions not being applied, meanwhile the fix is about client-side exclusions being wrongly applied for bound modules. Maybe I misunderstood something, I just want to make sure the fix (which is relevant) really fixes your specific issue

damien-urruty-sonarsource (commented):
Some improvement suggestions. I also asked for clarification because I'm not sure the ticket description really corresponds to the fix

backend/core/src/main/java/org/sonarsource/sonarlint/core/fs/FileExclusionService.java
nquinquenel: I have rewritten the content of the ticket, the main point was in the comment (which I have now removed).

I think the previous content was outdated and not relevant anymore. I've kept it because this issue of connected mode was preventing me from fully testing the behavior explained in the ticket. Now that this PR fixes the issue, I do not notice any further problems.

So the issue is: because the config scope is never detected as in connected mode, the client file exclusions are always applied even in connected mode.

backend/core/src/main/java/org/sonarsource/sonarlint/core/fs/FileExclusionService.java
nquinquenel: Good point, it was removed by mistake from another refactoring, I will add it back.

backend/core/src/main/java/org/sonarsource/sonarlint/core/repository/config/ConfigurationRepository.java
nquinquenel: Modified them to "by"

backend/core/src/main/java/org/sonarsource/sonarlint/core/fs/FileExclusionService.java
nquinquenel: It's effectively much more convenient to use the effective binding, which I forgot about. I completely replaced this connected mode method.

kirill-knize-sonarsource (approved):
LGTM

PR #1402, opened 2025-07-03
SLCORE-1510 Show feedback when 'Search in list...' fails

eray-felek-sonarsource:


backend/core/src/main/java/org/sonarsource/sonarlint/core/ConnectionService.java
```diff
@@ -183,9 +187,14 @@ public HelpGenerateUserTokenResponse helpGenerateUserToken(String serverUrl, @Nu
...
     var serverApi = sonarQubeClientManager.getForTransientConnection(transientConnection);
-    return serverApi.component().getAllProjects(cancelMonitor)
-      .stream().map(serverProject -> new SonarProjectDto(serverProject.key(), serverProject.name()))
-      .toList();
+
+    try {
+      return serverApi.component().getAllProjects(cancelMonitor)
+        .stream().map(serverProject -> new SonarProjectDto(serverProject.key(), serverProject.name()))
+        .toList();
+    } catch (UnauthorizedException e) {
```
damien-urruty-sonarsource, line 195: Could there be other types of exception raised? It's not related to this ticket, but I think this logic should be used in every case a server request is made from an RPC request.

damien-urruty-sonarsource (approved):
Good catch! I think it's a good start, but we will probably need to generalize the approach to all RPC requests that deal with HTTP calls. And it wouldn't be practical to do it for every request. We could maybe leverage the lsp4j's uncaught exception handler. I can create a ticket about that.
Last comment, the ticket and commit message is phrased from the IntelliJ perspective. Could you try to change it to be more neutral and better describe the changes you made to SLCORE?

backend/core/src/main/java/org/sonarsource/sonarlint/core/ConnectionService.java
eray-felek-sonarsource: As discussed this will be tackled on a different ticket

damien-urruty-sonarsource: For reference: https://sonarsource.atlassian.net/browse/SLCORE-1511

PR #1408, opened 2025-07-07
[DO NOT MERGE] SC-28409 Make sonar-security compatible with SLCORE

damien-urruty-sonarsource:


damien-urruty-sonarsource:
Closing this now, it was only an experiment

PR #1411, opened 2025-07-10
SLCORE-1518 SLCORE-1519 SLCORE-1520 Fetch project bindings and key and provide suggestion

eray-felek-sonarsource:


backend/core/src/main/java/org/sonarsource/sonarlint/core/BindingSuggestionProvider.java
```diff
@@ -181,6 +187,11 @@ private List<BindingSuggestionDto> suggestBindingForEligibleScope(String checked
         }
       }
     }
+
+    if(suggestions.isEmpty()) {
```
damien-urruty-sonarsource: Formatting seems off here, should be `if (...`

backend/core/src/main/java/org/sonarsource/sonarlint/core/BindingSuggestionProvider.java
```diff
@@ -191,6 +202,39 @@ private void searchGoodMatchInConnections(List<BindingSuggestionDto> suggestions
...
 
+  private void searchByRemoteUrlInConnections(List<BindingSuggestionDto> suggestions, String configScopeId, Set<String> connectionIds,
+    SonarLintCancelMonitor cancelMonitor) {
+    for (var connectionId : connectionIds) {
+      var remoteUrl = GitService.getRemoteUrl(clientFs.getBaseDir(configScopeId));
+
+      if (remoteUrl == null) {
+        LOG.debug("No remote URL found for configuration scope '{}', skipping search in connection '{}'", configScopeId, connectionId);
+        return;
+      }
```
damien-urruty-sonarsource: Couldn't this be extracted out of the loop? If there is no remote URL it's not needed to go over each connection. Also it's not necessary to read it at each iteration

backend/core/src/main/java/org/sonarsource/sonarlint/core/BindingSuggestionProvider.java
```diff
@@ -191,6 +202,39 @@ private void searchGoodMatchInConnections(List<BindingSuggestionDto> suggestions
...
+
+      Optional<String> projectIdOpt =
+        sonarQubeClientManager.withActiveClientFlatMapOptionalAndReturn(
+          connectionId,
+          api -> api.projectBindings().getProjectIdByUrl(remoteUrl, cancelMonitor)
+        );
+
+      if( projectIdOpt.isEmpty() ) {
```
damien-urruty-sonarsource: Formatting is off too here

backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/component/ComponentApi.java
```diff
@@ -82,6 +83,51 @@ private String getAllProjectsUrl() {
...
+        LOG.warn("Failed to retrieve project for ID: {} (status: {})", projectId, response.code());
+      }
+    } catch (Exception e) {
+      LOG.error("Error retrieving project for ID: {}", projectId, e);
+    }
+
+    return Optional.empty();
+  }
+
+  public static class ProjectKeyName {
```
damien-urruty-sonarsource: You could use a record here

backend/core/src/main/java/org/sonarsource/sonarlint/core/BindingSuggestionProvider.java
```diff
@@ -191,6 +202,39 @@ private void searchGoodMatchInConnections(List<BindingSuggestionDto> suggestions
...
+      if (remoteUrl == null) {
+        LOG.debug("No remote URL found for configuration scope '{}', skipping search in connection '{}'", configScopeId, connectionId);
+        return;
+      }
+
+      Optional<String> projectIdOpt =
+        sonarQubeClientManager.withActiveClientFlatMapOptionalAndReturn(
+          connectionId,
+          api -> api.projectBindings().getProjectIdByUrl(remoteUrl, cancelMonitor)
+        );
```
damien-urruty-sonarsource: I wonder if we couldn't store the binding during the sync. I'm a bit concerned that we reach out to the server every time

backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/component/ComponentApi.java
```diff
@@ -82,6 +83,51 @@ private String getAllProjectsUrl() {
...
+      if (response.isSuccessful()) {
+        var responseBody = response.bodyAsString();
+        var jsonResponse = JsonParser.parseString(responseBody).getAsJsonObject();
+        var components = jsonResponse.getAsJsonArray("components");
+
+        if (!components.isEmpty()) {
+          var firstComponent = components.get(0).getAsJsonObject();
+          var key = firstComponent.get("key").getAsString();
+          var name = firstComponent.get("name").getAsString();
+
```
damien-urruty-sonarsource: We usually read JSON in a different, more concise way. You could have a look at `SystemApi` to get inspiration

backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/component/ComponentApi.java
```diff
@@ -82,6 +83,51 @@ private String getAllProjectsUrl() {
     return searchUrl.toString();
   }
 
+  public Optional<ProjectKeyName> getProjectKeyByProjectId(String projectId, SonarLintCancelMonitor cancelMonitor) {
+    var encodedProjectId = java.net.URLEncoder.encode(projectId, java.nio.charset.StandardCharsets.UTF_8);
```
damien-urruty-sonarsource: You can static import `UrlUtils.urlEncode` instead. See other API classes

backend/core/src/main/java/org/sonarsource/sonarlint/core/BindingSuggestionProvider.java
```diff
@@ -191,6 +202,39 @@ private void searchGoodMatchInConnections(List<BindingSuggestionDto> suggestions
...
+      if( projectIdOpt.isEmpty() ) {
+        LOG.debug("No project ID found for remote URL '{}' on connection '{}'", remoteUrl, connectionId);
+        return;
+      }
+
+      var projectKeyName =
+        sonarQubeClientManager.withActiveClientFlatMapOptionalAndReturn(
+          connectionId,
+          api -> api.component().getProjectKeyByProjectId(projectIdOpt.get(), cancelMonitor)
+        );
```
damien-urruty-sonarsource: I think it's great that we implement this mapping, because I think IDs will be more and more used in the future. Same I'm a bit concerned that we fetch it everytime. Maybe we could save it in the storage too, I'm sure it will be useful in some other cases. You can check `ServerInfoSynchronizer#readOrSynchronizeServerInfo` for such an example of lazily fetch from the server if not present on disk

backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/component/ComponentApi.java
```diff
@@ -82,6 +83,51 @@ private String getAllProjectsUrl() {
     return searchUrl.toString();
   }
 
+  public Optional<ProjectKeyName> getProjectKeyByProjectId(String projectId, SonarLintCancelMonitor cancelMonitor) {
```
damien-urruty-sonarsource: Lately we are trying to just wrap the API without adding too much meaning in the *Api classes (that's also what we tried to do on MCP). What about renaming the method `searchProjects`, as the endpoint, and return a `SearchProjectsResponse` type? The fields would be the same, but I think it's more future-proof

backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/component/ComponentApi.java
```diff
@@ -82,6 +83,51 @@ private String getAllProjectsUrl() {
...
+          var name = firstComponent.get("name").getAsString();
+
+          return Optional.of(new ProjectKeyName(key, name));
+        }
+      } else {
+        LOG.warn("Failed to retrieve project for ID: {} (status: {})", projectId, response.code());
+      }
+    } catch (Exception e) {
+      LOG.error("Error retrieving project for ID: {}", projectId, e);
+    }
```
damien-urruty-sonarsource, line 111: I would avoid catching and returning an Optional from this method. I think it's better to let the problem bubble up. The `SonarQubeClientManager` should be the one taking care of that

backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/projectbindings/ProjectBindingsApi.java
```diff
@@ -0,0 +1,61 @@
...
+  private static final SonarLintLogger LOG = SonarLintLogger.get();
+
+  private final ServerApiHelper serverApiHelper;
+
+  public ProjectBindingsApi(ServerApiHelper serverApiHelper) {
+    this.serverApiHelper = serverApiHelper;
+  }
+
+  public Optional<String> getProjectIdByUrl(String url, SonarLintCancelMonitor cancelMonitor) {
+    var encodedUrl = java.net.URLEncoder.encode(url, java.nio.charset.StandardCharsets.UTF_8);
```
damien-urruty-sonarsource: Same remark here, you can use the utility method

backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/projectbindings/ProjectBindingsApi.java
```diff
@@ -0,0 +1,61 @@
...
+      if (response.isSuccessful()) {
+        var responseBody = response.bodyAsString();
+        var jsonResponse = JsonParser.parseString(responseBody).getAsJsonObject();
+        var bindings = jsonResponse.getAsJsonArray("bindings");
+
+        if (!bindings.isEmpty()) {
+          var firstBinding = bindings.get(0).getAsJsonObject();
+          var projectId = firstBinding.get("projectId").getAsString();
+          return Optional.of(projectId);
+        }
```
damien-urruty-sonarsource: Same remark here, you could simplify the deserialization

backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/projectbindings/ProjectBindingsApi.java
```diff
@@ -0,0 +1,61 @@
...
+public class ProjectBindingsApi {
+  private static final SonarLintLogger LOG = SonarLintLogger.get();
+
+  private final ServerApiHelper serverApiHelper;
+
+  public ProjectBindingsApi(ServerApiHelper serverApiHelper) {
+    this.serverApiHelper = serverApiHelper;
+  }
+
+  public Optional<String> getProjectIdByUrl(String url, SonarLintCancelMonitor cancelMonitor) {
```
damien-urruty-sonarsource: Same remark here, I would call the method `getProjectBindings` and return a `GetProjectBindingsResponse` (not an optional)

damien-urruty-sonarsource (commented):
A first round of comments. You probably also lack a few medium tests, to validate the feature in integration

backend/core/src/main/java/org/sonarsource/sonarlint/core/BindingSuggestionProvider.java
eray-felek-sonarsource: I created [this](https://sonarsource.atlassian.net/browse/SLCORE-1538) ticket for it and I am planning to take a look after the SQS implementation as discussed

backend/core/src/main/java/org/sonarsource/sonarlint/core/BindingSuggestionProvider.java
eray-felek-sonarsource: I created [this](https://sonarsource.atlassian.net/browse/SLCORE-1538) ticket for it and I am planning to take a look after the SQS implementation as discussed

backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/component/ComponentApi.java
eray-felek-sonarsource: I would prefer to catch the errors at this level since SonarQubeClientManager does not catch general exceptions and I prefer not using try catch block while calling it but I agree we should not return optional

damien-urruty-sonarsource: Then you should at least mark the method with CheckForNull

backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/projectbindings/ProjectBindingsApi.java
```diff
@@ -0,0 +1,59 @@
...
+public class ProjectBindingsApi {
+  private static final SonarLintLogger LOG = SonarLintLogger.get();
+
+  private final ServerApiHelper serverApiHelper;
+
+  public ProjectBindingsApi(ServerApiHelper serverApiHelper) {
+    this.serverApiHelper = serverApiHelper;
+  }
+
+  public ProjectBindingsResponse getProjectBindings(String url, SonarLintCancelMonitor cancelMonitor) {
```
damien-urruty-sonarsource, line 39: Same here, you should mark the method with CheckForNull

backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/projectbindings/ProjectBindingsResponseDto.java
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.sonarlint.core.serverapi.projectbindings;
+
+import java.util.List;
+
+public class ProjectBindingsResponseDto {
```
damien-urruty-sonarsource: Record?

medium-tests/src/test/java/mediumtest/BindingSuggestionsMediumTests.java
```diff
@@ -326,4 +334,71 @@ void test_uses_binding_clues_from_shared_configuration_when_updating_fs(SonarLin
...
+      .willReturn(aResponse()
+        .withStatus(200)
+        .withHeader("Content-Type", "application/x-protobuf")
+        .withResponseBody(protobufBody(Components.SearchWsResponse.newBuilder()
+          .addComponents(Components.Component.newBuilder()
+            .setKey("my-project-key")
+            .setName("My Project")
+            .build())
+          .setPaging(Common.Paging.newBuilder().setTotal(1).build())
+          .build()))));
```
damien-urruty-sonarsource, line 392: I think I mentioned that earlier, but this could be moved to the `ServerFixture`. It will make the test less verbose

damien-urruty-sonarsource (approved):
Some minor comments, else LGTM!

medium-tests/src/test/java/mediumtest/BindingSuggestionsMediumTests.java
eray-felek-sonarsource: Yes, I was also noticed this improvement but I think it can come as a nice to have after SQS and Telemetry is implemented. I will carry this comment to the feature branch

PR #1412, opened 2025-07-10
SLCORE-1522 Synchronize SCA enablement setting from the server

damien-urruty-sonarsource:
[SLCORE-1522](https://sonarsource.atlassian.net/browse/SLCORE-1522)

[SLCORE-1522]: https://sonarsource.atlassian.net/browse/SLCORE-1522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/settings/SettingsApi.java
+++ b/backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/settings/SettingsApi.java
@@ -24,7 +24,6 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.function.BiConsumer;
-import javax.annotation.Nullable;
 import org.sonarsource.sonarlint.core.commons.log.SonarLintLogger;
 import org.sonarsource.sonarlint.core.commons.progress.SonarLintCancelMonitor;
 import org.sonarsource.sonarlint.core.serverapi.ServerApiHelper;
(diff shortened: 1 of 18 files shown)


backend/server-connection/src/main/proto/sonarlint.proto
```diff
@@ -28,8 +28,11 @@ option optimize_for = SPEED;
 
 message ServerInfo {
   string version = 1;
-  optional bool isMqrMode = 2;
-  optional bool misraEarlyAccessRulesEnabled = 3;
+  // deprecated, store in the settings map below
+  optional bool isMqrMode = 2 [deprecated=true];
+  // deprecated, store in the settings map below
+  optional bool misraEarlyAccessRulesEnabled = 3 [deprecated=true];
+  map<string, string> globalSettings = 4;
```
sophio-japharidze-sonarsource, line 35: Some setting values are arrays of strings, will this still work in those cases? For example the `sonar.html.file.suffixes` setting

backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/settings/SettingsApi.java
```diff
@@ -41,32 +40,17 @@ public SettingsApi(ServerApiHelper helper) {
...
-              break;
-            }
-          }
-        } catch (IOException e) {
-          throw new IllegalStateException("Unable to parse properties from: " + response.bodyAsString(), e);
-        }
-      },
-      duration -> LOG.info("Downloaded settings in {}ms", duration));
-    return settings.get(settingKey);
+  public Map<String, String> getGlobalSettings(SonarLintCancelMonitor cancelMonitor) {
```
sophio-japharidze-sonarsource, line 43: I'm just wondering if there is any risk with synchronising ALL settings. Could we/ should we maintain an explicit list of settings that we are interested in?

backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/StoredServerInfo.java
```diff
@@ -19,41 +19,20 @@
...
-  public StoredServerInfo(Version version, @Nullable Boolean mode, boolean misraEarlyAccessRulesEnabled) {
-    this.version = version;
-    this.misraEarlyAccessRulesEnabled = misraEarlyAccessRulesEnabled;
-    if (mode == null) {
-      this.severityMode = SeverityModeDetails.DEFAULT;
-    } else if (mode) {
-      this.severityMode = SeverityModeDetails.MQR;
-    } else {
-      this.severityMode = SeverityModeDetails.STANDARD;
+  public boolean shouldConsiderMultiQualityModeEnabled() {
```
sophio-japharidze-sonarsource, line 30: Sorry if I missed it, but did we lose the check for SQC & MQR mode? The setting only exists on SQS

backend/server-connection/src/main/proto/sonarlint.proto
damien-urruty-sonarsource: You can check `org.sonarsource.sonarlint.core.serverapi.settings.SettingsApi#processSetting`, multi-values settings are grouped into a single string

backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/settings/SettingsApi.java
damien-urruty-sonarsource: I was wondering the same, but I did not find any problem. There is a separate ticket to avoid synchronizing sensitive settings ([SLCORE-1449](https://sonarsource.atlassian.net/browse/SLCORE-1449)) that should improve this case. Other than that I think it's fine, we should better align the behavior with the server by honoring the same settings (at least for loading rules today)

backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/StoredServerInfo.java
damien-urruty-sonarsource: There was a case for SQC in `org.sonarsource.sonarlint.core.mode.SeverityModeService#isMQRModeForConnection` already. Do you think I should make this clearer?

sophio-japharidze-sonarsource: OK perfect! No I just did not find the place, all good 👍

sophio-japharidze-sonarsource (approved):

PR #1418, opened 2025-07-17
SLCORE-1528 Allow clients to fetch Dependency Risk description fields

damien-urruty-sonarsource:
[SLCORE-1528](https://sonarsource.atlassian.net/browse/SLCORE-1528)

[SLCORE-1528]: https://sonarsource.atlassian.net/browse/SLCORE-1528?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/API_CHANGES.md
+++ b/API_CHANGES.md
@@ -5,6 +5,7 @@
 * Allow changing status of SCA issues via `org.sonarsource.sonarlint.core.rpc.protocol.backend.sca.ScaRpcService.changeStatus`.
   * Required parameters are `configScopeId`, `issueId` and `transition`.
   * If transition is `ACCEPT`, `FIXED`, or `SAFE`, a `comment` field is mandatory
+* Add a new `org.sonarsource.sonarlint.core.rpc.protocol.backend.sca.ScaRpcService.getDependencyRiskDetails`.
 
 # 10.26
 
--- a/backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaService.java
+++ b/backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaService.java
@@ -21,12 +21,19 @@ package org.sonarsource.sonarlint.core.sca;
 
 import java.util.UUID;
 import javax.annotation.CheckForNull;
+import org.eclipse.lsp4j.jsonrpc.ResponseErrorException;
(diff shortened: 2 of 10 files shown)


backend/core/src/main/java/org/sonarsource/sonarlint/core/tracking/ScaIssueTrackingService.java
```diff
@@ -105,4 +127,27 @@ private static ScaIssueDto toDto(ServerScaIssue serverScaIssue) {
         .map(transition -> ScaIssueDto.Transition.valueOf(transition.name()))
         .toList());
   }
+
+  private static GetDependencyRiskDetailsResponse convertToRpcResponse(org.sonarsource.sonarlint.core.serverapi.sca.GetDependencyRiskDetailsServerResponse serverResponse) {
+    var affectedPackages = serverResponse.affectedPackages().stream()
+      .map(pkg -> AffectedPackageDto.builder()
```
sophio-japharidze-sonarsource: is `packageName`, `packageVersion`, and `CVE id` missing intentionally?

sophio-japharidze-sonarsource (changes requested):
I believe `packageVersion`, `packageName`, and `vulnerabilityId` (CVE key) are missing

backend/core/src/main/java/org/sonarsource/sonarlint/core/tracking/ScaIssueTrackingService.java
damien-urruty-sonarsource: packageName and packageVersion, yes it is intentional. This data is already made available when publishing the SCA issues, so I thought I could skip them here. Do you think they should appear here too?
For CVE ID, I missed it since it was not in the example payload. Where am I supposed to find it?

sophio-japharidze-sonarsource: I think it would be nice to have all the necessary data in the response. Otherwise it would mean that client will need to make additional operations after receiving the response to get the matching risk, and get its package name and version.

sophio-japharidze-sonarsource (approved):

sophio-japharidze-sonarsource:
We need to update API_CHANGES doc as well

PR #1427, opened 2025-07-18
SLCORE-1529 Implement SQS Binding

eray-felek-sonarsource:
[SLCORE-1529](https://sonarsource.atlassian.net/browse/SLCORE-1529)

[SLCORE-1529]: https://sonarsource.atlassian.net/browse/SLCORE-1529?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/core/src/main/java/org/sonarsource/sonarlint/core/BindingSuggestionProvider.java
+++ b/backend/core/src/main/java/org/sonarsource/sonarlint/core/BindingSuggestionProvider.java
@@ -32,6 +32,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import org.jetbrains.annotations.NotNull;
 import org.sonarsource.sonarlint.core.commons.log.SonarLintLogger;
 import org.sonarsource.sonarlint.core.commons.progress.ExecutorServiceShutdownWatchable;
 import org.sonarsource.sonarlint.core.commons.progress.SonarLintCancelMonitor;
@@ -47,7 +48,7 @@ import org.sonarsource.sonarlint.core.repository.connection.ConnectionConfigurat
 import org.sonarsource.sonarlint.core.rpc.protocol.SonarLintRpcClient;
 import org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding.BindingSuggestionDto;
 import org.sonarsource.sonarlint.core.rpc.protocol.client.binding.SuggestBindingParams;
-import org.sonarsource.sonarlint.core.serverapi.projectbindings.ProjectBindingsResponse;
+import org.sonarsource.sonarlint.core.serverapi.ServerApi;
 import org.springframework.context.event.EventListener;
 
 import static java.lang.String.join;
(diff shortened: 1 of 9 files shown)


backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/projectbindings/ProjectBindingsApi.java
```diff
@@ -36,18 +36,52 @@ public ProjectBindingsApi(ServerApiHelper serverApiHelper) {
...
     var encodedUrl = UrlUtils.urlEncode(url);
+
+    return getSQCBinding(url, cancelMonitor, encodedUrl);
+  }
+
+  @CheckForNull
+  public SQSProjectBindingsResponse getSQSProjectBindings(String url, SonarLintCancelMonitor cancelMonitor) {
+    var encodedUrl = UrlUtils.urlEncode(url);
+
+    return getSQSBinding(url, cancelMonitor, encodedUrl);
```
damien-urruty-sonarsource: Why having a separate private method here? It's called only from here, so I would suggest to inline it. Same for the other one

backend/core/src/main/java/org/sonarsource/sonarlint/core/BindingSuggestionProvider.java
```diff
@@ -212,19 +213,33 @@ private void searchByRemoteUrlInConnections(List<BindingSuggestionDto> suggestio
     }
 
     for (var connectionId : connectionIds) {
-      Optional<ProjectBindingsResponse> projectIdOpt = sonarQubeClientManager.withActiveClientFlatMapOptionalAndReturn(connectionId,
-        api -> Optional.ofNullable(api.projectBindings().getProjectBindings(remoteUrl, cancelMonitor)));
+      var suggestion = sonarQubeClientManager.withActiveClientFlatMapOptionalAndReturn(connectionId, api ->
```
eray-felek-sonarsource: sonarQubeClientManager.withActiveClientFlatMapOptionalAndReturn can throw an HttpHostConnectException. We should probably wrap it around try catch too

backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/projectbindings/ProjectBindingsApi.java
```diff
@@ -36,18 +36,52 @@ public ProjectBindingsApi(ServerApiHelper serverApiHelper) {
...
   @CheckForNull
-  public ProjectBindingsResponse getProjectBindings(String url, SonarLintCancelMonitor cancelMonitor) {
+  public SQCProjectBindingsResponse getSQCProjectBindings(String url, SonarLintCancelMonitor cancelMonitor) {
     var encodedUrl = UrlUtils.urlEncode(url);
+
+    return getSQCBinding(url, cancelMonitor, encodedUrl);
+  }
+
+  @CheckForNull
+  public SQSProjectBindingsResponse getSQSProjectBindings(String url, SonarLintCancelMonitor cancelMonitor) {
```
damien-urruty-sonarsource: Instead of having 2 different methods for SQS and SQC, and let the caller decide which one to call, I think it would make more sense to encapsulate this in the ProjectBindingsApi, or have 2 different variants. But this probably requires a more in-depth refactoring

backend/server-api/src/test/java/org/sonarsource/sonarlint/core/serverapi/projectbindings/ProjectBindingsApiTests.java
```diff
@@ -46,49 +46,97 @@ void setUp() {
...
 
-    var result = underTest.getProjectBindings(url, new SonarLintCancelMonitor());
+    var result = underTest.getSQCProjectBindings(url, new SonarLintCancelMonitor());
 
     assertThat(result).isNull();
   }
 
   @Test
-  void should_return_empty_when_request_fails() {
+  void should_return_empty_when_request_fails_for_sqc() {
```
damien-urruty-sonarsource: Random idea: Instead of suffixing every method, you could also use `@Nested` and have one for SQS and one for SQC

backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/projectbindings/ProjectBindingsApi.java
eray-felek-sonarsource: I considered something like Either structure but I would prefer to have different responses to keep things separate.

PR #1431, opened 2025-07-21
SLCORE-1521 Add telemetry for remote URL binding suggestion

eray-felek-sonarsource:
[SLCORE-1521](https://sonarsource.atlassian.net/browse/SLCORE-1521)

[SLCORE-1521]: https://sonarsource.atlassian.net/browse/SLCORE-1521?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/core/src/main/java/org/sonarsource/sonarlint/core/BindingSuggestionProvider.java
+++ b/backend/core/src/main/java/org/sonarsource/sonarlint/core/BindingSuggestionProvider.java
@@ -49,6 +49,7 @@ import org.sonarsource.sonarlint.core.rpc.protocol.SonarLintRpcClient;
 import org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding.BindingSuggestionDto;
 import org.sonarsource.sonarlint.core.rpc.protocol.client.binding.SuggestBindingParams;
 import org.sonarsource.sonarlint.core.serverapi.ServerApi;
+import org.sonarsource.sonarlint.core.telemetry.TelemetryService;
 import org.springframework.context.event.EventListener;
 
 import static java.lang.String.join;
@@ -69,10 +70,11 @@ public class BindingSuggestionProvider {
   private final AtomicBoolean enabled = new AtomicBoolean(true);
   private final SonarQubeClientManager sonarQubeClientManager;
   private final ClientFileSystemService clientFs;
+  private final TelemetryService telemetryService;
 
   @Inject
   public BindingSuggestionProvider(ConfigurationRepository configRepository, ConnectionConfigurationRepository connectionRepository, SonarLintRpcClient client,
-    BindingClueProvider bindingClueProvider, SonarProjectsCache sonarProjectsCache, SonarQubeClientManager sonarQubeClientManager, ClientFileSystemService clientFs) {
+    BindingClueProvider bindingClueProvider, SonarProjectsCache sonarProjectsCache, SonarQubeClientManager sonarQubeClientManager, ClientFileSystemService clientFs, TelemetryService telemetryService) {
     this.configRepository = configRepository;
     this.connectionRepository = connectionRepository;
     this.client = client;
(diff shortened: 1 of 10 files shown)


backend/core/src/main/java/org/sonarsource/sonarlint/core/BindingSuggestionProvider.java
```diff
@@ -192,6 +195,9 @@ private List<BindingSuggestionDto> suggestBindingForEligibleScope(String checked
 
     if (suggestions.isEmpty()) {
       searchByRemoteUrlInConnections(suggestions, checkedConfigScopeId, candidateConnectionIds, cancelMonitor);
+      if (!suggestions.isEmpty()) {
+        telemetryService.suggestedRemoteBinding();
+      }
```
damien-urruty-sonarsource, line 200: Could we aim for something a bit more ambitious, and measure all types of suggestions instead of only this specific case? We could collect where the suggestion was found (shared connected mode, properties file, or project binding), I don't think it would be much harder to implement

backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/payload/ShareConnectedModePayload.java
```diff
@@ -25,5 +25,6 @@
 public record ShareConnectedModePayload(@SerializedName("manual_bindings_count") @Nullable Integer manualAddedBindingsCount,
                                         @SerializedName("imported_bindings_count") @Nullable Integer importedAddedBindingsCount,
                                         @SerializedName("auto_bindings_count") @Nullable Integer autoAddedBindingsCount,
-                                        @SerializedName("exported_connected_mode_count") @Nullable Integer exportedConnectedModeCount) {
+                                        @SerializedName("exported_connected_mode_count") @Nullable Integer exportedConnectedModeCount,
+                                        @SerializedName("suggested_remote_bindings_count") @Nullable Integer suggestedRemoteBindingsCount) {
```
damien-urruty-sonarsource: I think you don't need this one, this is the old way to measure things. I see you modified  `TelemetryMeasuresBuilder` already so you should be good

backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/measures/payload/TelemetryMeasuresBuilder.java
```diff
@@ -85,6 +85,7 @@ private void addConnectedModeMeasures(ArrayList<TelemetryMeasuresValue> values)
       values.add(new TelemetryMeasuresValue("shared_connected_mode.imported", String.valueOf(storage.getImportedAddedBindingsCount()), INTEGER, DAILY));
       values.add(new TelemetryMeasuresValue("shared_connected_mode.auto", String.valueOf(storage.getAutoAddedBindingsCount()), INTEGER, DAILY));
       values.add(new TelemetryMeasuresValue("shared_connected_mode.exported", String.valueOf(storage.getExportedConnectedModeCount()), INTEGER, DAILY));
+      values.add(new TelemetryMeasuresValue("shared_connected_mode.remote_url", String.valueOf(storage.getSuggestedRemoteBindingsCount()), INTEGER, DAILY));
```
damien-urruty-sonarsource, line 88: I don't see a test to verify that the measure is sent, you could check in the codebase for other measures but I'm pretty sure there will be tests around them

backend/core/src/main/java/org/sonarsource/sonarlint/core/BindingSuggestionProvider.java
eray-felek-sonarsource: I will discuss this with Farah but as we discussed this would not be part of this MMF

PR #1434, opened 2025-07-22
SLCORE-1545 Update dependencies to fix CVEs

nquinquenel:
[SLCORE-1545](https://sonarsource.atlassian.net/browse/SLCORE-1545)

Fixing following dependencies:
- okio-jvm
- commons-fileupload
- commons-lang3
- commons-lang
- tomcat-embed-core

Changes implied:
- `MockResponse()` constructor is not supported anymore -> Builder
- `isBlank`, `trimToNull`, `StringEscapeUtils`, etc. removed from `commons-lang3` -> Add dependency on `commons-text`
- `removeEnd` deprecated -> `Strings.CS.removeEnd` as a replacement
- Old ITs projects depending on old dependencies for `analyzer-commons`, `sonar-plugin-api` and `org.sonarsource.sonarqube` -> Modified to fit the versions used by LTA SQ:S (9.9)
- Kotlin version updated to fit the new `okhttp` version

[SLCORE-1545]: https://sonarsource.atlassian.net/browse/SLCORE-1545?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/commons/src/test/java/org/sonarsource/sonarlint/core/commons/testutils/MockWebServerExtension.java
+++ b/backend/commons/src/test/java/org/sonarsource/sonarlint/core/commons/testutils/MockWebServerExtension.java
@@ -27,7 +27,6 @@ import mockwebserver3.MockResponse;
 import mockwebserver3.MockWebServer;
 import mockwebserver3.RecordedRequest;
 import okio.Buffer;
-import org.jetbrains.annotations.NotNull;
 import org.junit.jupiter.api.extension.AfterEachCallback;
 import org.junit.jupiter.api.extension.BeforeEachCallback;
 import org.junit.jupiter.api.extension.ExtensionContext;
@@ -51,7 +50,6 @@ public class MockWebServerExtension implements BeforeEachCallback, AfterEachCall
     server = new MockWebServer();
     responsesByPath.clear();
     final Dispatcher dispatcher = new Dispatcher() {
-      @NotNull
       @Override
       public MockResponse dispatch(RecordedRequest request) {
         if (responsesByPath.containsKey(request.getPath())) {
--- a/backend/commons/src/test/java/org/sonarsource/sonarlint/core/commons/testutils/MockWebServerExtension.java
+++ b/backend/commons/src/test/java/org/sonarsource/sonarlint/core/commons/testutils/MockWebServerExtension.java
@@ -27,6 +27,7 @@ import mockwebserver3.MockResponse;
 import mockwebserver3.MockWebServer;
 import mockwebserver3.RecordedRequest;
 import okio.Buffer;
+import org.jetbrains.annotations.NotNull;
 import org.junit.jupiter.api.extension.AfterEachCallback;
 import org.junit.jupiter.api.extension.BeforeEachCallback;
 import org.junit.jupiter.api.extension.ExtensionContext;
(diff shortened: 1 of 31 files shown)


backend/commons/src/test/java/org/sonarsource/sonarlint/core/commons/testutils/MockWebServerExtension.java
```diff
@@ -27,6 +27,7 @@
 import mockwebserver3.MockWebServer;
 import mockwebserver3.RecordedRequest;
 import okio.Buffer;
+import org.jetbrains.annotations.NotNull;
```
damien-urruty-sonarsource: I was about to comment that I find usage of `jetbrains` annotations a bit strange, but in fact we have them in a lot of places !
I think we could get rid of most of them since normally the `package-info.java` and its usual content should be enough for parameters. And for return types we shouldn't need them either. But it's not directly related to your changes, so we could do it separately

backend/commons/src/test/java/org/sonarsource/sonarlint/core/commons/testutils/MockWebServerExtension.java
```diff
@@ -50,12 +51,13 @@ public void start() {
     server = new MockWebServer();
     responsesByPath.clear();
     final Dispatcher dispatcher = new Dispatcher() {
+      @NotNull
```
damien-urruty-sonarsource: This one for example shouldn't be needed I think

backend/server-api/src/test/java/org/sonarsource/sonarlint/core/serverapi/push/PushApiTests.java
```diff
@@ -62,13 +62,14 @@ void setUp() {
...
-      "\"key\": \"key1\"," +
-      "\"value\": \"value1\"" +
-      "}\n\n");
+    var mockResponse = new MockResponse.Builder()
+      .body("event: SettingChanged\n" +
+        "data: {" +
+        "\"projects\": [\"projectKey1\", \"projectKey2\"]," +
+        "\"key\": \"key1\"," +
+        "\"value\": \"value1\"" +
+        "}\n\n")
```
damien-urruty-sonarsource: Unrelated, but since you touched this code, we could use text blocks in this class, it would make it a bit more readable

damien-urruty-sonarsource (approved):
Nice job, thanks for simplifying our lives with Mend!
I just left some minor comments, feel free to ignore

damien-urruty-sonarsource:
@nquinquenel Forgot to comment, I know that at least SLLS declares the Kotlin version [here](https://github.com/SonarSource/sonarlint-language-server/blob/40a92e77a8112a6de58a32f623af0b4f66c0dfe8/pom.xml#L117). I don't know if we should update the version there too, or if we could just remove this dependency (that should be pulled transitively?)

backend/commons/src/test/java/org/sonarsource/sonarlint/core/commons/testutils/MockWebServerExtension.java
nquinquenel: IntelliJ complains about this but let's remove it anyway

backend/server-api/src/test/java/org/sonarsource/sonarlint/core/serverapi/push/PushApiTests.java
nquinquenel: I was too lazy at first, but thankfully my AI friend did it!

PR #1442, opened 2025-07-23
SLCORE-1554 Update local storage and notify clients after changing dependency risk status

damien-urruty-sonarsource:


backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/DependencyRiskService.java
```diff
@@ -91,14 +150,26 @@ public void changeStatus(String configurationScopeId, UUID dependencyRiskKey, De
       throw new IllegalArgumentException("Transition " + transition + " is not allowed for this dependency risk");
     }
 
-    if ((transition == DependencyRiskTransition.ACCEPT || transition == DependencyRiskTransition.SAFE || transition == DependencyRiskTransition.FIXED)
+    if ((transition == DependencyRiskTransition.ACCEPT || transition == DependencyRiskTransition.SAFE)
       && (comment == null || comment.isBlank())) {
       throw new IllegalArgumentException("Comment is required for ACCEPT, FIXED, and SAFE transitions");
```
sophio-japharidze-sonarsource: Need to adjust the log, and several other places (I think there's a reference in tests & API_CHANGES doc as well)

sophio-japharidze-sonarsource (approved):

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/DependencyRiskService.java
damien-urruty-sonarsource: Good catch! Fixed

PR #1443, opened 2025-07-24
SLCORE-1550 Use ServerFixture for BindingSuggestionsMediumTests

eray-felek-sonarsource:
[SLCORE-1550](https://sonarsource.atlassian.net/browse/SLCORE-1550)

[SLCORE-1550]: https://sonarsource.atlassian.net/browse/SLCORE-1550?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/medium-tests/src/test/java/mediumtest/BindingSuggestionsMediumTests.java
+++ b/medium-tests/src/test/java/mediumtest/BindingSuggestionsMediumTests.java
@@ -27,6 +27,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.List;
+import java.util.UUID;
 import java.util.Map;
 import java.util.concurrent.ExecutionException;
 import org.eclipse.jgit.api.errors.GitAPIException;
@@ -45,7 +46,6 @@ import org.sonarsource.sonarlint.core.rpc.protocol.backend.connection.config.Did
 import org.sonarsource.sonarlint.core.rpc.protocol.backend.connection.config.SonarQubeConnectionConfigurationDto;
 import org.sonarsource.sonarlint.core.rpc.protocol.backend.file.DidUpdateFileSystemParams;
 import org.sonarsource.sonarlint.core.rpc.protocol.common.ClientFileDto;
-import org.sonarsource.sonarlint.core.serverapi.UrlUtils;
 import org.sonarsource.sonarlint.core.serverapi.proto.sonarqube.ws.Common;
 import org.sonarsource.sonarlint.core.serverapi.proto.sonarqube.ws.Components;
 import org.sonarsource.sonarlint.core.test.utils.junit5.SonarLintTest;
@@ -53,8 +53,6 @@ import org.sonarsource.sonarlint.core.test.utils.junit5.SonarLintTestHarness;
 
 import static com.github.tomakehurst.wiremock.client.WireMock.aResponse;
 import static com.github.tomakehurst.wiremock.client.WireMock.get;
-import static com.github.tomakehurst.wiremock.client.WireMock.urlEqualTo;
-import static com.github.tomakehurst.wiremock.client.WireMock.urlPathMatching;
 import static com.github.tomakehurst.wiremock.core.WireMockConfiguration.wireMockConfig;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.tuple;
@@ -71,7 +69,7 @@ class BindingSuggestionsMediumTests {
   public static final String CONFIG_SCOPE_ID = "myProject1";
   public static final String SLCORE_PROJECT_KEY = "org.sonarsource.sonarlint:sonarlint-core-parent";
   public static final String SLCORE_PROJECT_NAME = "SonarLint Core";
-  public static final String PROJECT_ID = "my-project-id";
+  public static final String PROJECT_ID = "123e4567-e89b-12d3-a456-426614174000";
   public static final String REMOTE_URL = "git@github.com:myorg/myproject.git";
 
   @RegisterExtension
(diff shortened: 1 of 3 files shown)


medium-tests/src/test/java/mediumtest/BindingSuggestionsMediumTests.java
```diff
@@ -359,6 +354,9 @@ void should_suggest_binding_by_remote_url_when_no_other_suggestions_found(SonarL
     var scServer = harness.newFakeSonarCloudServer()
       .withOrganization("orgKey", organization ->
         organization.withProject(SLCORE_PROJECT_KEY, project -> project.withBranch("main")))
+      .withDopTranslation(dop -> dop
+        .withProjectBinding(REMOTE_URL, PROJECT_ID, SLCORE_PROJECT_KEY)
+        .withSearchProjectsResponse("orgKey", PROJECT_ID, 200, "{\"components\":[{\"key\":\"" + SLCORE_PROJECT_KEY + "\",\"name\":\"" + SLCORE_PROJECT_NAME + "\"}]}\n"))
```
damien-urruty-sonarsource: I would do this slightly differently. I think `withDopTranslation` and `withProjectBinding` are good, but I would remove `withSearchProjectsResponse`. Just above, there is already a call to `organization.withProject`, you could make sure this call registers a response to `api/components/search_projects`. You can also modify the project builder to add a `withName`. This would make the test even less verbose

medium-tests/src/test/java/mediumtest/BindingSuggestionsMediumTests.java
```diff
@@ -486,6 +441,7 @@ void should_return_empty_when_sqc_project_bindings_is_null(SonarLintTestHarness
     var scServer = harness.newFakeSonarCloudServer()
       .withOrganization("orgKey", organization ->
         organization.withProject(SLCORE_PROJECT_KEY, project -> project.withBranch("main")))
+      .withCustomEndpointResponse(expectedPath, 500, "text/plain", "Internal Server Error")
```
damien-urruty-sonarsource: Instead of introducing a new method, you could use `withResponseCodes`. You can see how it's used in other tests, and you will have to adjust it. It's not perfect but it should work

medium-tests/src/test/java/mediumtest/BindingSuggestionsMediumTests.java
```diff
@@ -619,24 +552,16 @@ void should_return_empty_when_sqs_server_project_is_not_present(SonarLintTestHar
 
     var sqServer = harness.newFakeSonarQubeServer()
       .withProject(SLCORE_PROJECT_KEY, project -> project.withBranch("main"))
+      .withDopTranslation(dop -> dop
+        .withProjectBinding(REMOTE_URL, PROJECT_ID, SLCORE_PROJECT_KEY)
+        .withComponentShowStatus(SLCORE_PROJECT_KEY, 404, "Project not found"))
```
damien-urruty-sonarsource: Same here

damien-urruty-sonarsource (commented):
Some comments to improve it even further

test-utils/src/main/java/org/sonarsource/sonarlint/core/test/utils/server/ServerFixture.java
```diff
@@ -1579,5 +1630,58 @@ public String url(String path) {
     public WireMockServer getMockServer() {
       return mockServer;
     }
+
+    private void registerDopTranslationApiResponses() {
+      dopTranslation.projectBindings.forEach((repositoryUrl, projectBinding) -> {
+        String encodedUrl = UrlUtils.urlEncode(repositoryUrl);
```
nquinquenel: We have the habit of using `var` in the project

medium-tests/src/test/java/mediumtest/BindingSuggestionsMediumTests.java
```diff
@@ -603,25 +531,15 @@ void should_return_empty_when_sqs_server_project_is_not_present(SonarLintTestHar
       .build();
 
     var sqServer = harness.newFakeSonarQubeServer()
-      .withProject(SLCORE_PROJECT_KEY, project -> project.withBranch("main"))
+      .withDopTranslation(dop -> dop
+        .withProjectBinding(REMOTE_URL, PROJECT_ID, SLCORE_PROJECT_KEY))
```
damien-urruty-sonarsource, line 541: What I just suggested might be a problem for this specific test. In fact, you could keep both systems: being able to declare a binding on its own, and being able to declare a binding via the project (and this second mechanism could under the hood use the first one).

test-utils/src/main/java/org/sonarsource/sonarlint/core/test/utils/server/ServerFixture.java
```diff
@@ -1364,6 +1404,28 @@ private void registerComponentsApiResponses() {
       registerComponentsSearchApiResponses();
       registerComponentsShowApiResponses();
       registerComponentsTreeApiResponses();
+
+      dopTranslation.projectBindings.forEach((repositoryUrl, projectBinding) -> {
```
damien-urruty-sonarsource: I think the starting point here should be the list of projects, not the list of bindings

test-utils/src/main/java/org/sonarsource/sonarlint/core/test/utils/server/ServerFixture.java
```diff
@@ -1364,6 +1404,28 @@ private void registerComponentsApiResponses() {
...
+          return;
+        }
+        var organizationKey = project.organizationKey;
+        if (organizationKey == null) {
+          return;
+        }
+        var projectName = project.name;
+        if (projectName != null) {
+          String endpoint = "/api/components/search_projects?projectIds=" + projectBinding.projectId() + "&organization=" + organizationKey;
+          String body = "{\"components\":[{\"key\":\"" + projectKey + "\",\"name\":\"" + projectName + "\"}]}";
```
damien-urruty-sonarsource: You could use `var`s everywhere

damien-urruty-sonarsource: You could use a text block here, it's usually more readable

test-utils/src/main/java/org/sonarsource/sonarlint/core/test/utils/server/ServerFixture.java
```diff
@@ -1581,5 +1643,28 @@ public String url(String path) {
     public WireMockServer getMockServer() {
       return mockServer;
     }
+
+    private void registerDopTranslationApiResponses() {
+      dopTranslation.projectBindings.forEach((repositoryUrl, projectBinding) -> {
+        var encodedUrl = UrlUtils.urlEncode(repositoryUrl);
+        if (serverKind == ServerKind.SONARCLOUD) {
+          String endpoint = "/dop-translation/project-bindings?url=" + encodedUrl;
+          String responseBody = "{\"bindings\":[{\"projectId\":\"" + projectBinding.projectId() + "\"}]}";
```
damien-urruty-sonarsource: Same here

PR #1445, opened 2025-07-25
SLCORE-1576 Notify client for invalid token on connection attempt

vnaskos-sonar:
[SLVSCODE-1279](https://sonarsource.atlassian.net/browse/SLVSCODE-1279)

# For SonarSourcers:

- [x] Prefix the commit message with the ticket number, i.e. `SLCORE-XXXX` if you already have a ticket in Jira
- [ ] For standalone PRs without issue in Jira:
    - [ ] Mention Epic ID in this descrition to create a new Task in Jira
    - [ ] Mention Issue ID in this descrition to create a new Sub-Task in Jira
    - [ ] Do not mention any Jira issue to create a new Task in Jira without a parent
- [ ] When changing an API:
-  [x] Make sure the tests adhere to the convention:

[SLVSCODE-1279]: https://sonarsource.atlassian.net/browse/SLVSCODE-1279?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/core/src/main/java/org/sonarsource/sonarlint/core/ConnectionService.java
```diff
@@ -232,4 +237,13 @@ public Map<String, String> getProjectNamesByKey(Either<TransientSonarQubeConnect
...
 
+  private boolean isAnonymousConnection(Either<TransientSonarQubeConnectionDto, TransientSonarCloudConnectionDto> transientConnection) {
+    Either<TokenDto, UsernamePasswordDto> credentials = transientConnection
+      .map(TransientSonarQubeConnectionDto::getCredentials, TransientSonarCloudConnectionDto::getCredentials);
+
+    boolean isTokenEmpty = credentials.getLeft() == null || credentials.getLeft().getToken() == null || credentials.getLeft().getToken().isEmpty();
+    boolean isUserPassEmpty = credentials.getRight() == null || credentials.getRight().getUsername() == null || credentials.getRight().getPassword() == null;
+
+    return isTokenEmpty && isUserPassEmpty;
+  }
```
nquinquenel: These fields are not nullable, the client should likely throw an error (or handle this in a correct way) instead of sending null values 🤔

vnaskos-sonar: Hmm, on the `ConnectionAwareHttpClientProvider.java` there were already some null checks for those.

When a user imports the VSCode settings to Cursor, the secure storage is not imported so the tokens are null. However, the null token is reported as valid from the `serverApi.authentication().validate`, like an anonymous user and because we didn't have the `null` checks here (to prevent anonymous connections), the UI on Cursor connections list displays them as `ok`. All the other connected mode actions that use the `ConnectionAwareHttpClientProvider` are failing correctly because of invalid credentials. 🤔

nquinquenel: I don't think we have any null check on the SLLS when calling `validateConnection`, in `createValidateConnectionParams` we should not create the DTO if token is null

damien-urruty-sonarsource: In general, the backend should validate what the client sends, like in a traditional web app, the backend shouldn't always trust the frontend and should check every input. The changes in `ConnectionAwareHttpClientProvider` were a first step in this direction.
The problem is that it would be very costly to check all parameters for each method. I had created https://sonarsource.atlassian.net/browse/SLCORE-1282 to show the need to introduce some validation, but as usual it was not prioritized... My idea would be to have some kind of framework and not leave the validation to every single method.
I agree with you Nico this could also be checked on the client side. In VSCode there is a situation where the connection is not complete/coherent, so you could also avoid declaring the connection and consider it "broken". You could offer a special UX to make it complete again

vnaskos-sonar: There are 3 issues here
1. SLLS doesn't check for null
    > which I agree that it has to be checked, the frontend should not pass invalid data
1. SLCore does not prevent you from sending an invalid DTO
    > which can be fixed with https://sonarsource.atlassian.net/browse/SLCORE-1282
1. SLCore on `validateConnection` does a request to the `api/authentication/validate?format=json` where null token means anonymous user
    > an anonymous user is a valid case for SQC, SQS users, but it's not valid for the SQIDE scope. In this specific case I believe we have to add the additional check (at least until the SLCORE-1282 is implemented).

backend/core/src/main/java/org/sonarsource/sonarlint/core/connection/SonarQubeClient.java
```diff
@@ -33,24 +33,21 @@ public class SonarQubeClient {
 
   private static final Period WRONG_TOKEN_NOTIFICATION_INTERVAL = Period.ofDays(1);
   private final String connectionId;
+  @Nullable
   private final ServerApi serverApi;
```
vnaskos-sonar, line 37: @nquinquenel @damien-urruty-sonarsource 

Could this be a problem?

Some context:

When invalid credentials were provided the `SonarQubeClient` was unable to be constructed. The `ConnectionAwareHttpClientProvider#getHttpClient` was throwing an `IllegalStateException` so the `SonarQubeClientManager` was retrying on every request to create a new `SonarQubeClient` with the same broken credentials.

The result was, that the frontend was receiving multiple times the Exception for the same connection and the `invalidToken` notification was never triggered.

Changing the `serverApi` to nullable allows the `SonarQubeClient` to be properly constructed and reach the `invalidToken` notification. The `state` becomes `INVALID_CREDENTIALS` and the frontend receives only once that the connection is broken.

damien-urruty-sonarsource: I would find it confusing to have a `SonarQubeClient` without `ServerApi`. What about not having the SonarQubeClient in the first place? And trigger the `invalidToken` from another place if needed

vnaskos-sonar: > What about not having the SonarQubeClient in the first place?

This is the case today, on an invalid token the `SonarQubeClient` can't be created. The `SonarQubeClientManager` throws Exceptions each time it tries to create a new instance of `SonarQubeClient`. (The exceptions are thrown because the httpClient can't be created when the token is invalid, and httpClient is a dependency of `ServerApi` which is a dependency of `SonarQubeClient`)

This also means that the `SonarQubeClient#state` field is not working properly. The only case that the state is `INVALID_CREDENTIALS` is if the credentials become invalid (eg. expire) while the connection is already in use, which is kinda rare, otherwise the `INVALID_CREDENTIALS` state is unreachable.

> And trigger the invalidToken from another place if needed

Calling the `invalidToken` elsewhere is kinda tricky, because we'll get multiple times the notification for the same connection. Since the `SonarQubeClient` already has the state, it seems appropriate to use it for reporting the invalidToken.

The `serverApi` is not exposed outside the `SonarQubeClient`, so it should be safe to have it as nullable. Another option would be to have a state on the ServerApi, but it seems more complicated.

damien-urruty-sonarsource: I would still find it confusing. There are 2 cases:
* no token is provided by the client for a transient connection. In this case I think we should throw right away, transient connections are used from RPC requests and the best we can do is throw so that clients can react to a failed request.
* no token is provided by the client for regular connections. Here it's a bit more annoying because `ConnectionRpcService#didUpdateConnections` is a RPC notification, so we cannot fail it. I think that would be a good time to call `invalidToken` and maybe put the connection in an invalid state right away. 

For those 2 cases, when proper validation will be later implemented, it should ideally throw on the client side when instantiating the DTO, so before even reaching the RPC layer

damien-urruty-sonarsource: After a call we decided to move forward with the proposed approach. It's not ideal in terms of design but it should do the job, and the ticket about the validation should avoid such situations and simplify this case in the future

backend/core/src/test/java/org/sonarsource/sonarlint/core/http/ConnectionAwareHttpClientProviderTests.java
```diff
@@ -0,0 +1,71 @@
...
+      .thenReturn(CompletableFuture.completedFuture(response));
+
+    assertThatThrownBy(() -> underTest.getHttpClient(connectionId, shouldUseBearer))
+      .isInstanceOf(IllegalStateException.class)
+      .hasMessage("No token for connection " + connectionId);
+    ArgumentCaptor<InvalidTokenParams> paramsCaptor = ArgumentCaptor.forClass(InvalidTokenParams.class);
+    verify(client).invalidToken(paramsCaptor.capture());
+    InvalidTokenParams capturedParams = paramsCaptor.getValue();
+    assertThat(capturedParams.getConnectionId()).isEqualTo(connectionId);
+  }
```
nquinquenel: We usually use `var`

backend/core/src/main/java/org/sonarsource/sonarlint/core/SonarQubeClientManager.java
```diff
@@ -26,6 +26,7 @@
 import java.util.function.Function;
 import org.eclipse.lsp4j.jsonrpc.ResponseErrorException;
 import org.eclipse.lsp4j.jsonrpc.messages.ResponseError;
+import org.jetbrains.annotations.Nullable;
```
nquinquenel: We try to use `javax.annotation.Nullable`

backend/core/src/test/java/org/sonarsource/sonarlint/core/SonarQubeClientManagerTests.java
```diff
@@ -61,84 +65,177 @@ void setUp() {
...
-      false));
-    var httpClient = mock(HttpClient.class);
-    when(awareHttpClientProvider.getHttpClient("sc1", true)).thenReturn(httpClient);
-    when(awareHttpClientProvider.getHttpClient()).thenReturn(httpClient);
-    var httpResponse = mock(HttpClient.Response.class);
-    when(httpResponse.isSuccessful()).thenReturn(true);
-    when(httpResponse.bodyAsString()).thenReturn("{\"id\": \"20160308094653\",\"version\": \"9.9\",\"status\": \"UP\"}");
-    when(httpClient.getAsyncAnonymous("/api/system/status")).thenReturn(CompletableFuture.completedFuture(httpResponse));
+  void getClientOrThrow_for_sonarcloud_with_trailing_slash_notConnected() {
+    URI uriWithSlash = URI.create(SonarCloudRegion.EU.getProductionUri() + "/");
```
nquinquenel:
```suggestion
    var uriWithSlash = URI.create(SonarCloudRegion.EU.getProductionUri() + "/");
```

nquinquenel (approved):

PR #1450, opened 2025-07-29
SLCORE-1314 Investigate storage exceptions on synchronization

nquinquenel:


backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/IssueDownloader.java
```diff
@@ -106,7 +106,7 @@ public PullResult downloadFromPull(ServerApi serverApi, String projectKey, Strin
       .filter(i -> i.getMainLocation().hasFilePath())
       .filter(not(IssueLite::getClosed))
       .map(IssueDownloader::convertLiteIssue)
-      .toList();
+      .collect(Collectors.toList());
```
nquinquenel, line 109: `toList()` is raising an error in IntelliJ which prevents compiling, due to stricter rules regarding generic types

backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisService.java
```diff
@@ -293,23 +293,31 @@ public GetAnalysisConfigResponse getAnalysisConfig(String configScopeId, boolean
...
+        // This can happen when a standalone analysis was scheduled and a synchronization happened in between.
+        // The config scope is bound, but the config file is not yet created.
+        // In this case, we still trigger the analysis as if it was in standalone instead of failing.
+        // This log should not appear when a synchronization is not happening.
+        LOG.warn("Could not retrieve connected analysis configuration, falling back to standalone configuration");
+      }
+    }
+    var standaloneActiveRules = startChild(trace, "buildStandaloneActiveRules", GET_ANALYSIS_CFG, this::buildStandaloneActiveRules);
+    var embeddedPluginPaths = startChild(trace, "getEmbeddedPluginPaths", GET_ANALYSIS_CFG, pluginsService::getEmbeddedPluginPaths);
+    return new GetAnalysisConfigResponse(standaloneActiveRules, userAnalysisProperties, nodeJsDetailsDto, Set.copyOf(embeddedPluginPaths));
```
vnaskos-sonar, line 330: nitpick: This is becoming a bit difficult to read. I would suggest to extract some methods and reduce the nesting.

eg.
```java
    var bindingOpt = configurationRepository.getEffectiveBinding(configScopeId);
    var userAnalysisProperties = createAnalysisProperties(configScopeId);
    var nodeJsDetailsDto = getNodeJsDetailsDto(bindingOpt.isPresent(), trace);

    if (bindingOpt.isEmpty()) {
      return getStandaloneAnalysisConfig(userAnalysisProperties, nodeJsDetailsDto, trace);
    }

    var binding = bindingOpt.get();
    var analyzerConfig = storageService.binding(binding).analyzerConfiguration();
    if (!analyzerConfig.isValid()) {
      // This can happen when a standalone analysis was scheduled and a synchronization happened in between.
      // The config scope is bound, but the config file is not yet created.
      // In this case, we still trigger the analysis as if it was in standalone instead of failing.
      // This log should not appear when a synchronization is not happening.
      LOG.warn("Could not retrieve connected analysis configuration, falling back to standalone configuration");
      return getStandaloneAnalysisConfig(userAnalysisProperties, nodeJsDetailsDto, trace);
    }

    return getConnectedModeAnalysisConfig(userAnalysisProperties, binding, hotspotsOnly, nodeJsDetailsDto, trace);
```

vnaskos-sonar (approved):
👍

PR #1467, opened 2025-08-14
SLCORE-1551 Improve Current Binding Telemetry

eray-felek-sonarsource:


backend/core/src/main/java/org/sonarsource/sonarlint/core/BindingSuggestionProvider.java
```diff
@@ -179,7 +180,8 @@ private List<BindingSuggestionDto> suggestBindingForEligibleScope(String checked
         sonarProjectsCache
           .getSonarProject(connectionId, sonarProjectKey, cancelMonitor)
           .ifPresent(serverProject -> suggestions.add(new BindingSuggestionDto(connectionId, sonarProjectKey, serverProject.name(),
-            bindingClueWithConnections.getBindingClue().isFromSharedConfiguration())));
+            bindingClueWithConnections.getBindingClue().getBindingSuggestionOrigin() == BindingSuggestionOrigin.SHARED_CONFIGURATION,
+            bindingClueWithConnections.getBindingClue().getBindingSuggestionOrigin())));
```
damien-urruty-sonarsource: For the code to be a little bit less verbose, and also to prepare for the future and the removal of the boolean parameter, you could introduce a second constructor in `BindingSuggestionDto`, that would only accept the origin parameter and not the boolean, and would populate the boolean field appropriately. When we decide to drop the deprecated field, it will "just" be a deprecated constructor to remove.

backend/core/src/main/java/org/sonarsource/sonarlint/core/BindingSuggestionProvider.java
```diff
@@ -265,7 +267,8 @@ private void searchGoodMatchInConnection(List<BindingSuggestionDto> suggestions,
           break;
         }
         bestScore = serverProjectScoreEntry.getValue();
-        suggestions.add(new BindingSuggestionDto(connectionId, serverProjectScoreEntry.getKey().key(), serverProjectScoreEntry.getKey().name(), false));
+        suggestions.add(new BindingSuggestionDto(connectionId, serverProjectScoreEntry.getKey().key(),
+          serverProjectScoreEntry.getKey().name(), false, BindingSuggestionOrigin.PROJECT_NAME));
```
damien-urruty-sonarsource: It would also simplify such cases

backend/core/src/main/java/org/sonarsource/sonarlint/core/event/BindingConfigChangedEvent.java
```diff
@@ -19,7 +19,17 @@
  */
 package org.sonarsource.sonarlint.core.event;
 
+import javax.annotation.Nullable;
 import org.sonarsource.sonarlint.core.repository.config.BindingConfiguration;
+import org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding.BindingMode;
+import org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding.BindingSuggestionOrigin;
+
+public record BindingConfigChangedEvent(String configScopeId, BindingConfiguration previousConfig,
+                                        BindingConfiguration newConfig, @Nullable BindingMode bindingMode, @Nullable BindingSuggestionOrigin origin) {
```
damien-urruty-sonarsource: The formatting looks odd here, could you check if you follow the SonarSource convention?

medium-tests/src/test/java/mediumtest/BindingSuggestionsMediumTests.java
```diff
@@ -127,6 +128,7 @@ void test_connection_added_should_suggest_binding_with_matches(SonarLintTestHarn
     assertThat(bindingSuggestions.get(CONFIG_SCOPE_ID))
       .extracting(BindingSuggestionDto::getConnectionId, BindingSuggestionDto::getSonarProjectKey, BindingSuggestionDto::getSonarProjectName, BindingSuggestionDto::isFromSharedConfiguration)
       .containsExactly(tuple(MYSONAR, SLCORE_PROJECT_KEY, SLCORE_PROJECT_NAME, false));
+    assertThat(bindingSuggestions.get(CONFIG_SCOPE_ID).get(0).getOrigin()).isEqualTo(BindingSuggestionOrigin.PROJECT_NAME);
```
damien-urruty-sonarsource: Couldn't you extract `BindingSuggestionDto::getOrigin` as it's done above? It would be a little less verbose. You can probably change that for all assertions you added to this suite

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/config/binding/BindingMode.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding;
+
+public enum BindingMode {
+  MANUAL,
+  ASSISTED
```
damien-urruty-sonarsource: I think this name could be misleading. It makes me think about `assistBinding` which is one specific method in the RPC protocol, and we could think that the mode corresponds to this specific case. Maybe `FROM_SUGGESTION`, or `ACCEPTED_SUGGESTION`?

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/config/binding/BindingSuggestionDto.java
```diff
@@ -24,13 +24,17 @@ public class BindingSuggestionDto {
   private final String connectionId;
   private final String sonarProjectKey;
   private final String sonarProjectName;
+  @Deprecated(forRemoval = true)
```
damien-urruty-sonarsource, line 27: You also need to deprecate the getter, with both annotation and Javadoc. See other examples in the codebase

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/config/binding/DidUpdateBindingParams.java
```diff
@@ -19,15 +19,39 @@
...
+
 
+  public DidUpdateBindingParams(String configScopeId, BindingConfigurationDto updatedBinding,BindingMode bindingMode, BindingSuggestionOrigin origin) {
+    this.configScopeId = configScopeId;
+    this.updatedBinding = updatedBinding;
+    this.bindingMode = bindingMode;
+    this.origin = origin;
+  }
 
   public DidUpdateBindingParams(String configScopeId, BindingConfigurationDto updatedBinding) {
```
damien-urruty-sonarsource, line 53: You should also deprecate this constructor

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/config/binding/DidUpdateBindingParams.java
```diff
@@ -19,15 +19,39 @@
...
+import javax.annotation.Nullable;
+
 public class DidUpdateBindingParams {
 
   private final String configScopeId;
   private final BindingConfigurationDto updatedBinding;
+  @Nullable
+  private final BindingMode bindingMode;
+  @Nullable
+  private final BindingSuggestionOrigin origin;
```
damien-urruty-sonarsource: I would add a bit of documentation on those fields. The first one is temporarily Nullable, it's more for backward compatibility, but at some point clients will be expected to pass it. The second one is nullable and should be set if the value of the first field is "ASSISTED" (with the current naming)

backend/core/src/main/java/org/sonarsource/sonarlint/core/BindingCandidatesFinder.java
```diff
@@ -76,7 +77,8 @@ private Optional<ConfigurationScopeSharedContext> checkIfScopeIsGoodCandidateFor
 
 
     if (!cluesWithMatchingProjectKey.isEmpty()) {
-      var isFromSharedConfiguration = cluesWithMatchingProjectKey.stream().anyMatch(c -> c.getBindingClue().isFromSharedConfiguration());
+      var isFromSharedConfiguration = cluesWithMatchingProjectKey.stream().anyMatch(
+        c -> c.getBindingClue().getBindingSuggestionOrigin() == BindingSuggestionOrigin.SHARED_CONFIGURATION);
       return Optional.of(new ConfigurationScopeSharedContext(scope, isFromSharedConfiguration));
```
damien-urruty-sonarsource: This `ConfigurationScopeSharedContext` is used in the context of `AssistBindingParams`. Should it also be modified to introduce a new 'origin' concept and replace `isFromSharedConfiguration`? If not, how would the clients determine what value to pass to `DidUpdateBindingParams`?

backend/core/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryService.java
```diff
@@ -240,6 +243,30 @@ public void addedAutomaticBindings() {
     updateTelemetry(TelemetryLocalStorage::incrementAutoAddedBindingsCount);
   }
 
+  public void addedNewBinding(BindingMode bindingMode, @Nullable BindingSuggestionOrigin bindingSuggestionOrigin) {
```
damien-urruty-sonarsource, line 246: You will need to deprecate the 3 methods in `TelemetryRpcService` that should no longer be used

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/config/binding/DidUpdateBindingParams.java
damien-urruty-sonarsource, line 38: All those changes need to be documented in the `API_CHANGES.md` file so that clients know how to update

API_CHANGES.md
```diff
@@ -1,3 +1,38 @@
+# 10.29
+## New features
+* Clients can now access more granular origin information via the BindingSuggestionDto, AssistBindingParams and ConfigurationScopeSharedContext classes.
```
damien-urruty-sonarsource: The last class is not part of the API, it's not in the `rpc-protocol` module, so you can omit it from here

API_CHANGES.md
```diff
@@ -1,3 +1,38 @@
...
+* Added BindingMode and BindingSuggestionOrigin to DidUpdateBindingParams and allowed clients to provide this information when updating a binding.
+* This will trigger a telemetry event during the didUpdateBinding call. Clients no longer need to call separate telemetry methods when adding bindings.
+
+## Deprecation
+
+* Deprecate the `org.sonarsource.sonarlint.core.ConfigurationScopeSharedContext#isFromSharedConfiguration` method
+* Since it is not used anymore. Use getOrigin instead
+* Deprecate `isFromSharedConfiguration` parameter from `org.sonarsource.sonarlint.core.ConfigurationScopeSharedContext`
+* This value on no longer needed on the backend side.
+* Deprecate the `org.sonarsource.sonarlint.core.ConfigurationScopeSharedContext` constructor. Use the other one.
```
damien-urruty-sonarsource: This can be removed

API_CHANGES.md
```diff
@@ -1,3 +1,38 @@
+# 10.29
+## New features
+* Clients can now access more granular origin information via the BindingSuggestionDto, AssistBindingParams and ConfigurationScopeSharedContext classes.
+* Added BindingMode and BindingSuggestionOrigin to DidUpdateBindingParams and allowed clients to provide this information when updating a binding.
+* This will trigger a telemetry event during the didUpdateBinding call. Clients no longer need to call separate telemetry methods when adding bindings.
+
+## Deprecation
+
+* Deprecate the `org.sonarsource.sonarlint.core.ConfigurationScopeSharedContext#isFromSharedConfiguration` method
+* Since it is not used anymore. Use getOrigin instead
```
damien-urruty-sonarsource: This 2nd sentence is a development of the first one. I wouldn't put it at the same level in a bullet list. You could keep a single point and move the second point as the second sentence of the first point. You can do that for all items below

API_CHANGES.md
```diff
@@ -1,3 +1,38 @@
...
+* Deprecate the `org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding.AssistBindingParams#isFromSharedConfiguration` method
+* since it is not used anymore. Use getOrigin instead
+* Deprecate `isFromSharedConfiguration` parameter from `org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding.AssistBindingParams`
+* This value on no longer needed on the backend side.
+* Deprecate the `org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding.AssistBindingParams` constructor. Use the other one.
+
+* Deprecate the `org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding.DidUpdateBindingParams` constructor. Use the other one.
+
+* Deprecate the `org.sonarsource.sonarlint.core.telemetry.TelemetryService#addedManualBindings` method
+* Rely on `org.sonarsource.sonarlint.core.telemetry.TelemetryService#addedNewBinding` instead.
```
damien-urruty-sonarsource: Not really, this shouldn't be an API method. You can say that this will be automatically handled if using `didUpdateBinding` and passing the mode and origin.

API_CHANGES.md
```diff
@@ -1,3 +1,38 @@
...
+
+* Deprecate the `org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding.AssistBindingParams#isFromSharedConfiguration` method
+* since it is not used anymore. Use getOrigin instead
+* Deprecate `isFromSharedConfiguration` parameter from `org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding.AssistBindingParams`
+* This value on no longer needed on the backend side.
+* Deprecate the `org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding.AssistBindingParams` constructor. Use the other one.
+
+* Deprecate the `org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding.DidUpdateBindingParams` constructor. Use the other one.
+
+* Deprecate the `org.sonarsource.sonarlint.core.telemetry.TelemetryService#addedManualBindings` method
```
damien-urruty-sonarsource: You are mixing the API and the implementation. It's `org.sonarsource.sonarlint.core.rpc.protocol.backend.telemetry.TelemetryRpcService`

backend/core/src/main/java/org/sonarsource/sonarlint/core/BindingCandidatesFinder.java
```diff
@@ -76,12 +77,20 @@ private Optional<ConfigurationScopeSharedContext> checkIfScopeIsGoodCandidateFor
...
+      var isFromSharedConfiguration = cluesWithMatchingProjectKey.stream().anyMatch(
+        c -> c.getBindingClue().getBindingSuggestionOrigin() == BindingSuggestionOrigin.SHARED_CONFIGURATION);
+      if (isFromSharedConfiguration) {
+        return Optional.of(new ConfigurationScopeSharedContext(scope, BindingSuggestionOrigin.SHARED_CONFIGURATION));
+      }
+      var isFromPropertiesFile = cluesWithMatchingProjectKey.stream().anyMatch(
+        c -> c.getBindingClue().getBindingSuggestionOrigin() == BindingSuggestionOrigin.PROPERTIES_FILE);
+      if (isFromPropertiesFile) {
+        return Optional.of(new ConfigurationScopeSharedContext(scope, BindingSuggestionOrigin.PROPERTIES_FILE));
+      }
```
damien-urruty-sonarsource: I think this changes the functionality. Prior to the change, a `ConfigurationScopeSharedContext` was always returned from this if block, not it's not always the case. Maybe you should just pick the first clue and use the origin from this clue to pass it to `ConfigurationScopeSharedContext`?

API_CHANGES.md
damien-urruty-sonarsource: Please use code delimiters and fully qualified names

backend/core/src/main/java/org/sonarsource/sonarlint/core/BindingClueProvider.java
```diff
@@ -279,17 +281,17 @@ public interface BindingClue {
     @CheckForNull
     String getSonarProjectKey();
 
-    boolean isFromSharedConfiguration();
+    BindingSuggestionOrigin getBindingSuggestionOrigin();
```
damien-urruty-sonarsource: Nitpick: I think the method name could be just `getOrigin`. It's the origin of the clue

backend/core/src/main/java/org/sonarsource/sonarlint/core/ConfigurationScopeSharedContext.java
```diff
@@ -20,23 +20,43 @@
 package org.sonarsource.sonarlint.core;
 
 import org.sonarsource.sonarlint.core.repository.config.ConfigurationScope;
+import org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding.BindingSuggestionOrigin;
 
 public class ConfigurationScopeSharedContext {
 
   private final ConfigurationScope configurationScope;
+  @Deprecated(forRemoval = true)
   private final boolean isFromSharedConfiguration;
```
damien-urruty-sonarsource: As I mentioned earlier, this class is not API, so no need to mark things as deprecated. In this case I think you could drop the field

backend/core/src/main/java/org/sonarsource/sonarlint/core/ConfigurationScopeSharedContext.java
```diff
@@ -20,23 +20,43 @@
...
+  private final BindingSuggestionOrigin origin;
 
+  @Deprecated(forRemoval = true)
   ConfigurationScopeSharedContext(ConfigurationScope configurationScope, boolean isFromSharedConfiguration) {
     this.configurationScope = configurationScope;
     this.isFromSharedConfiguration = isFromSharedConfiguration;
+    this.origin = isFromSharedConfiguration ? BindingSuggestionOrigin.SHARED_CONFIGURATION : null;
+  }
+
+  ConfigurationScopeSharedContext(ConfigurationScope configurationScope, BindingSuggestionOrigin origin) {
```
damien-urruty-sonarsource, line 30: Same, no need to have 2 different constructors

backend/core/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryService.java
```diff
@@ -228,18 +231,55 @@ private void issueStatusChanged(String ruleKey) {
     updateTelemetry(telemetryLocalStorage -> telemetryLocalStorage.addIssueStatusChanged(ruleKey));
   }
 
+  /**
+   * @deprecated avoid calling this method if possible, since it will be removed once all the clients are migrated.
+   * Rely on {@link #addedNewBinding(BindingMode, BindingSuggestionOrigin)}}  instead.
+   */
+  @Deprecated(forRemoval = true)
```
damien-urruty-sonarsource: You deprecated the wrong methods. This class is not an RPC API. See e.g. `org.sonarsource.sonarlint.core.rpc.protocol.backend.telemetry.TelemetryRpcService#addedManualBindings`

backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/measures/payload/TelemetryMeasuresBuilder.java
```diff
@@ -115,6 +117,17 @@ private void addConnectedModeMeasures(ArrayList<TelemetryMeasuresValue> values)
     }
   }
 
+  private void addNewBindingsMeasures(ArrayList<TelemetryMeasuresValue> values) {
+    if (liveAttributes.usesConnectedMode()) {
```
damien-urruty-sonarsource, line 121: I'm not sure this check is needed. Imagine the user bound a project but decided to unbind, we wouldn't collect the count. I think you should unconditionally add the measures

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/config/binding/BindingSuggestionDto.java
```diff
@@ -24,15 +24,31 @@ public class BindingSuggestionDto {
...
   private final boolean isFromSharedConfiguration;
+  private BindingSuggestionOrigin origin;
 
-  public BindingSuggestionDto(String connectionId, String sonarProjectKey, String sonarProjectName, boolean isFromSharedConfiguration) {
+  /**
+   * @deprecated Will be removed in a future version, use {@link #BindingSuggestionDto(String, String, String, BindingSuggestionOrigin)}
+   */
+  @Deprecated(forRemoval = true)
+  public BindingSuggestionDto(String connectionId, String sonarProjectKey, String sonarProjectName,
+    boolean isFromSharedConfiguration) {
```
damien-urruty-sonarsource: Is this constructor still used? You need to set a value to the `origin` field, else it will be null

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/config/binding/DidUpdateBindingParams.java
```diff
@@ -19,15 +19,51 @@
...
+  private final BindingMode bindingMode;
+  /**
+   * This value should only be set if the bindingMode is:
+   * {@link org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding.BindingMode#FROM_SUGGESTION}
+   */
+  @Nullable
+  private final BindingSuggestionOrigin origin;
+
 
+  public DidUpdateBindingParams(String configScopeId, BindingConfigurationDto updatedBinding,BindingMode bindingMode, BindingSuggestionOrigin origin) {
```
damien-urruty-sonarsource: Formatting seems wrong (missing space around comma)

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/client/binding/AssistBindingParams.java
```diff
@@ -19,19 +19,35 @@
...
   private final String projectKey;
   private final String configScopeId;
+  @Deprecated(forRemoval = true)
   private final boolean isFromSharedConfiguration;
+  BindingSuggestionOrigin origin;
 
+  /**
+   * @deprecated Will be removed in a future version, use {@link #AssistBindingParams(String, String, String, BindingSuggestionOrigin)}
+   */
+  @Deprecated(forRemoval = true)
```
damien-urruty-sonarsource: Same, `origin` is not set here

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/client/binding/AssistBindingParams.java
```diff
@@ -19,19 +19,35 @@
...
 
+import org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding.BindingSuggestionOrigin;
+
 public class AssistBindingParams {
   private final String connectionId;
   private final String projectKey;
   private final String configScopeId;
+  @Deprecated(forRemoval = true)
   private final boolean isFromSharedConfiguration;
+  BindingSuggestionOrigin origin;
```
damien-urruty-sonarsource: The field should be private final

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/config/binding/BindingSuggestionDto.java
```diff
@@ -24,15 +24,31 @@ public class BindingSuggestionDto {
   private final String connectionId;
   private final String sonarProjectKey;
   private final String sonarProjectName;
+  @Deprecated(forRemoval = true)
   private final boolean isFromSharedConfiguration;
+  private BindingSuggestionOrigin origin;
```
damien-urruty-sonarsource: The field should be final

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/config/binding/BindingSuggestionDto.java
eray-felek-sonarsource: removed as discussed

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/client/binding/AssistBindingParams.java
eray-felek-sonarsource: I removed the constructor since it is not called from client

backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/measures/payload/TelemetryMeasuresBuilder.java
eray-felek-sonarsource: This would require a discussion with the data team since it was decided like this by them.

API_CHANGES.md
```diff
@@ -1,3 +1,27 @@
...
+* Added `org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding.BindingMode` and `org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding.BindingSuggestionOrigin` to `org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding.DidUpdateBindingParams` and allowed clients to provide this information when calling `org.sonarsource.sonarlint.core.rpc.protocol.backend.config.ConfigurationRpcService#didUpdateBinding` which will trigger telemetry events for binding updates. Clients no longer need to call separate telemetry methods while adding bindings.
+
+## Deprecation
+
+* Deprecate the `org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding.BindingSuggestionDto#isFromSharedConfiguration` method since it is not used anymore. Use `org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding.BindingSuggestionDto#getOrigin` instead.
+* Deprecate the `org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding.BindingSuggestionDto` constructor. Use the other one.
+
```
damien-urruty-sonarsource: Nitpick: I would remove empty lines between bullets

API_CHANGES.md
```diff
@@ -1,3 +1,27 @@
...
+* Deprecate `isFromSharedConfiguration` parameter from `org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding.AssistBindingParams`.
+
+* Deprecate the `org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding.AssistBindingParams` constructor. Use the other one.
+
+* Deprecate the `org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding.DidUpdateBindingParams` constructor. Use the other one.
+
+* Deprecate the `org.sonarsource.sonarlint.core.rpc.protocol.backend.telemetry.TelemetryRpcService#addedManualBindings` method. 
+* This will be automatically handled by passing the `org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding.BindingSuggestionOrigin` and `org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding.BindingMode` to the `org.sonarsource.sonarlint.core.rpc.protocol.backend.config.binding.DidUpdateBindingParams` constructor during the `org.sonarsource.sonarlint.core.rpc.protocol.backend.config.ConfigurationRpcService#didUpdateBinding` call.
```
damien-urruty-sonarsource: This line should also go on the previous line since it's a continuation of the previous point. Same is true below for other points

rpc-protocol/src/test/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/config/binding/DidUpdateBindingParamsTests.java
```diff
@@ -0,0 +1,51 @@
...
+  void legacy_constructor_keeps_mode_and_origin_null() {
+    var dto = new BindingConfigurationDto("conn", "proj", false);
+
+    var params = new DidUpdateBindingParams("scope", dto);
+
+    assertThat(params.getConfigScopeId()).isEqualTo("scope");
+    assertThat(params.getUpdatedBinding()).isEqualTo(dto);
+    assertThat(params.getBindingMode()).isNull();
+    assertThat(params.getOrigin()).isNull();
+  }
```
damien-urruty-sonarsource, line 50: This kind of tests have low value, same for `AssistBindingParamsTests`. They are not testing any functionality, they are testing that a constructor saves fields and return them via getters. It's not very useful. You already modified `BindingSuggestionsMediumTests` to check the origin, so these tests are not useful IMO

backend/core/src/test/java/org/sonarsource/sonarlint/core/telemetry/TelemetryServiceTests.java
```diff
@@ -0,0 +1,134 @@
...
+import static org.mockito.Mockito.clearInvocations;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoMoreInteractions;
+import static org.mockito.Mockito.when;
+
+class TelemetryServiceTests {
```
damien-urruty-sonarsource, line 43: This is another kind of tests that I think are not very useful. You are testing the `TelemetryService` in isolation from the rest, how can we know that this service is invoked at the right time in "real life"? I would suggest to replace these tests with medium tests, trying to bind a configuration scope with different mode and origins, and check that the telemetry is updated appropriately. These would be much more powerful tests

backend/core/src/test/java/org/sonarsource/sonarlint/core/ConfigurationServiceTests.java
```diff
@@ -193,4 +195,60 @@ void should_clear_binding_if_connection_removed() {
       .containsExactly(BINDING_DTO_3.getConnectionId(), BINDING_DTO_3.getSonarProjectKey(), BINDING_DTO_3.isBindingSuggestionDisabled());
   }
 
+  @Test
+  void should_propagate_origin_and_binding_mode_in_didUpdateBinding_for_shared_configuration() {
```
damien-urruty-sonarsource, line 199: These 4 tests are the ones that could be replaced with medium tests

damien-urruty-sonarsource (approved):
Good job overall, I think it's mergeable like this! To be even better, you could replace some of the unit tests with medium tests to test the full functionality in a more integrated way

rpc-protocol/src/test/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/config/binding/DidUpdateBindingParamsTests.java
eray-felek-sonarsource: I'd still prefer to keep these instead of deleting

PR #1469, opened 2025-08-14
SLCORE-1597 Let clients know if SCA is supported

nquinquenel:
[SLCORE-1597](https://sonarsource.atlassian.net/browse/SLCORE-1597)

[SLCORE-1597]: https://sonarsource.atlassian.net/browse/SLCORE-1597?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/DependencyRiskService.java
```diff
@@ -99,6 +104,48 @@ public void onDependencyRisksSynchronized(DependencyRisksSynchronizedEvent event
...
+    if (effectiveBinding.isEmpty()) {
+      return new CheckDependencyRiskSupportedResponse(false, "The project is not bound, please bind it to SonarQube Server Enterprise 2025.4 or higher");
+    }
+    var connectionId = effectiveBinding.get().connectionId();
+    if (connectionRepository.getConnectionById(connectionId) == null) {
+      var error = new ResponseError(SonarLintRpcErrorCode.CONNECTION_NOT_FOUND, "The provided configuration scope is bound to an unknown connection: " + connectionId,
+        connectionId);
+      throw new ResponseErrorException(error);
+    }
+    return sonarQubeClientManager.withActiveClientAndReturn(effectiveBinding.get().connectionId(), serverApi -> {
```
damien-urruty-sonarsource: Why create a client here? If you store the result of `connectionRepository.getConnectionById(connectionId)` you can query the connection "kind"

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/DependencyRiskService.java
```diff
@@ -99,6 +104,48 @@ public void onDependencyRisksSynchronized(DependencyRisksSynchronizedEvent event
...
+      if (serverApi.isSonarCloud()) {
+        return new CheckDependencyRiskSupportedResponse(false, "SonarQube Cloud does not yet support dependency risks");
+      }
+      var optServerInfo = storageService.connection(connectionId).serverInfo().read();
```
damien-urruty-sonarsource: I'm not sure how this plays if the connection was never synchronized, it might throw

backend/core/src/test/java/org/sonarsource/sonarlint/core/sca/DependencyRiskServiceTests.java
```diff
@@ -19,18 +19,212 @@
...
 class DependencyRiskServiceTests {
 
+  private ConfigurationRepository configurationRepository;
+  private ConnectionConfigurationRepository connectionRepository;
+  private StorageService storageService;
+  private SonarQubeClientManager sonarQubeClientManager;
+  private DependencyRiskService underTest;
+
+  @BeforeEach
+  void setUp() {
```
damien-urruty-sonarsource: Why using unit tests with mocks? You could write medium tests, I think they would be less verbose, and you could verify what the client really receives as a response, not what one specific method throws

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/DependencyRiskService.java
nquinquenel: Not exactly sure how to handle this, I tested and could not reproduce a potential problem

PR #1474, opened 2025-08-22
SLCORE-1563 SLCORE-1566 Persist flight recorder data with a scheduler

Krosovok:
[SLCORE-1563](https://sonarsource.atlassian.net/browse/SLCORE-1563)
[SLCORE-1566](https://sonarsource.atlassian.net/browse/SLCORE-1566)

[SLCORE-1563]: https://sonarsource.atlassian.net/browse/SLCORE-1563?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SLCORE-1566]: https://sonarsource.atlassian.net/browse/SLCORE-1566?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/core/src/main/java/org/sonarsource/sonarlint/core/flight/recorder/FlightRecorderService.java
+++ b/backend/core/src/main/java/org/sonarsource/sonarlint/core/flight/recorder/FlightRecorderService.java
@@ -1,0 +1,63 @@
+/*
+ * SonarLint Core - Implementation
+ * Copyright (C) 2016-2025 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.sonarlint.core.flight.recorder;
+
+import com.google.common.util.concurrent.MoreExecutors;
+import jakarta.annotation.PostConstruct;
+import jakarta.annotation.PreDestroy;
+import java.time.Clock;
+import java.util.Map;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import org.sonarsource.sonarlint.core.commons.log.SonarLintLogger;
+import org.sonarsource.sonarlint.core.commons.util.FailSafeExecutors;
+
+public class FlightRecorderService {
+
+  public static final String SONARLINT_FLIGHT_RECORDER_PERIOD_PROPERTY = "sonarlint.internal.flight.recorder.interval.seconds";
+  private static final SonarLintLogger LOG = SonarLintLogger.get();
+  private static final int DEFAULT_15_MINUTES = 900;
+
+  private final FlightRecorderStorageService persister;
+  private final ScheduledExecutorService flightRecorder = FailSafeExecutors.newSingleThreadScheduledExecutor("Flight Recorder");
+
+  public FlightRecorderService(FlightRecorderStorageService persister) {
+    this.persister = persister;
+  }
+
+  @PostConstruct
+  public void launch() {
+    persister.populateSessionInitData(Map.of("param", "1"));
+    flightRecorder.scheduleAtFixedRate(this::update, 0,
+      Integer.getInteger(SONARLINT_FLIGHT_RECORDER_PERIOD_PROPERTY, DEFAULT_15_MINUTES), TimeUnit.SECONDS);
+  }
+
+  @PreDestroy
+  public void shutdown() {
+    if (!MoreExecutors.shutdownAndAwaitTermination(flightRecorder, 1, TimeUnit.SECONDS)) {
+      LOG.warn("Unable to stop flight recorder executor service in a timely manner");
+    }
+  }
+
+  private void update() {
+    // todo actual data https://sonarsource.atlassian.net/browse/SLCORE-1565
+    persister.appendData(Clock.systemDefaultZone(), Map.of("param", "1"));
+  }
+}
--- a/backend/core/src/main/java/org/sonarsource/sonarlint/core/flight/recorder/FlightRecorderStorageService.java
+++ b/backend/core/src/main/java/org/sonarsource/sonarlint/core/flight/recorder/FlightRecorderStorageService.java
@@ -1,0 +1,99 @@
+/*
+ * SonarLint Core - Implementation
+ * Copyright (C) 2016-2025 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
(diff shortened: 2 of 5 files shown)


Krosovok:
@nquinquenel Feel free to take a look too.

backend/core/src/main/java/org/sonarsource/sonarlint/core/flight/recorder/FlightRecorderStorageService.java
```diff
@@ -0,0 +1,101 @@
...
+        records = populateWithInitialData(data);
+      }
+      var recordsWithHeader = getRecordsWithHeader(records, dateTime);
+      Files.write(filePath, recordsWithHeader, StandardOpenOption.APPEND);
+    } catch (IOException e) {
+      LOG.error("Failed to write to a flight recorder file.", e);
+    }
+  }
+
+  @NotNull
```
damien-urruty-sonarsource: We usually don't rely on `@NotNull` annotations. We assume objects are `NotNull` by default, and explicitly use `@Nullable` when something can be null

backend/core/src/test/java/org/sonarsource/sonarlint/core/flight/recorder/FlightRecorderStorageServiceTests.java
```diff
@@ -0,0 +1,147 @@
...
+  @AfterEach
+  void tearDown() throws IOException {
+    FileUtils.delete(LOG_FOLDER.toFile(), FileUtils.RECURSIVE);
+  }
+
+  @Test
+  void should_create_file_with_date_time_in_name_and_date_row() {
+    tested.appendData(BASE_TIME, Map.of());
+
+    assertThat(TIMESTAMPED_FOLDER).isDirectoryContaining("glob:**" + FlightRecorderStorageService.FILE_NAME);
```
damien-urruty-sonarsource: I think you don't need a glob pattern here, couldn't you use `isDirectoryContaining( FlightRecorderStorageService.FILE_NAME)`?

medium-tests/src/test/java/mediumtest/flight/recorder/FlightRecorderMediumTests.java
```diff
@@ -0,0 +1,97 @@
...
+import org.sonarsource.sonarlint.core.test.utils.junit5.SonarLintTest;
+import org.sonarsource.sonarlint.core.test.utils.junit5.SonarLintTestHarness;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.await;
+import static org.sonarsource.sonarlint.core.flight.recorder.FlightRecorderService.SONARLINT_FLIGHT_RECORDER_PERIOD_PROPERTY;
+
+class FlightRecorderMediumTests {
+
+  private final Path logFolder = SonarLintUserHome.get().resolve("log");
```
damien-urruty-sonarsource: You will end up writing in the real location. If you stop a test in the middle (while e.g. debugging) the data will remain. It could mess with real data, even if it's unlikely. You could instead on `backend.getUserHome`. The fixture already takes care of creating a temporary user home. In fact I don't even know how the tests pass, since the user home is customized in the fixture

backend/core/src/main/java/org/sonarsource/sonarlint/core/flight/recorder/FlightRecorderService.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.concurrent.TimeUnit;
+import org.sonarsource.sonarlint.core.commons.util.FailSafeExecutors;
+
+public class FlightRecorderService {
+
+  public static final String SONARLINT_FLIGHT_RECORDER_PERIOD_PROPERTY = "sonarlint.internal.flight.recorder.interval.seconds";
+  private static final int DEFAULT_15_MINUTES = 900;
+
+  private final FlightRecorderStorageService persister;
+  private final ScheduledExecutorService flightRecorder = FailSafeExecutors.newSingleThreadScheduledExecutor("Flight Recorder");
```
damien-urruty-sonarsource, line 39: This executor service needs to be shutdown in a `@PreDestroy` method, you can check other services that use this `FailSafeExecutors`

damien-urruty-sonarsource (commented):
It's mostly good, I left suggestions, and the executor service needs to be shutdown

backend/core/src/test/java/org/sonarsource/sonarlint/core/flight/recorder/FlightRecorderStorageServiceTests.java
Krosovok: No, the method requires either glob or regex.

medium-tests/src/test/java/mediumtest/flight/recorder/FlightRecorderMediumTests.java
Krosovok: FYI It was working because of relying on same `SonarLintUserHome.get()` that uses env variable, not the values in `UserPaths` that are actually replaced.

backend/core/src/main/java/org/sonarsource/sonarlint/core/flight/recorder/FlightRecorderStorageService.java
```diff
@@ -0,0 +1,100 @@
...
+public class FlightRecorderStorageService {
+
+  public static final String FILE_NAME = "flight-records.txt";
+  private static final SonarLintLogger LOG = SonarLintLogger.get();
+  private static final DateTimeFormatter FILE_NAME_FORMATTER = DateTimeFormatter.ofPattern("'flight-recording-session'-dd-MM-yyyy-HH-mm");
+  private static final DateTimeFormatter RECORD_HEADER_FORMATTER = DateTimeFormatter.ofPattern("_____dd/MM/yyyy-HH:mm_____");
+
+  private String sessionFolderName;
+  private final Map<String, String> sessionInitData = new HashMap<>();
+  private final Path logFolder;
```
nquinquenel: I usually prefer to put `private final` variables before `private` variables, most of the code convention recommend this

backend/core/src/main/java/org/sonarsource/sonarlint/core/flight/recorder/FlightRecorderStorageService.java
```diff
@@ -0,0 +1,100 @@
...
+  public void populateSessionInitData(Map<String, String> initData) {
+    sessionInitData.putAll(initData);
+  }
+
+  public void appendData(Clock clock, Map<String, String> data) {
+    var dateTime = LocalDateTime.now(clock);
+    try {
+      var filePath = getRecordFile(dateTime);
+      var isEmptyFile = Files.size(filePath) == 0;
+      Map<String, String> records = data;
```
nquinquenel:
```suggestion
      var records = data;
```

backend/core/src/main/java/org/sonarsource/sonarlint/core/flight/recorder/FlightRecorderStorageService.java
```diff
@@ -0,0 +1,100 @@
...
+    return populated;
+  }
+
+  private static List<String> getRecordsWithHeader(Map<String, String> data, LocalDateTime dateTime) {
+    var records = data.entrySet()
+      .stream()
+      .map(entry -> entry.getKey() + "=" + entry.getValue())
+      .toArray(String[]::new);
+    return Lists.asList(RECORD_HEADER_FORMATTER.format(dateTime), records);
+  }
```
nquinquenel: I feel like we don't really make use of guava in the code, what about something like this:

```
  private static List<String> getRecordsWithHeader(Map<String, String> data, LocalDateTime dateTime) {
    return Stream.concat(Stream.of(RECORD_HEADER_FORMATTER.format(dateTime)),
      data.entrySet().stream().map(e -> e.getKey() + "=" + e.getValue())).toList();
  }
```

backend/core/src/main/java/org/sonarsource/sonarlint/core/flight/recorder/FlightRecorderService.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import org.sonarsource.sonarlint.core.commons.log.SonarLintLogger;
+import org.sonarsource.sonarlint.core.commons.util.FailSafeExecutors;
+
+public class FlightRecorderService {
+
+  public static final String SONARLINT_FLIGHT_RECORDER_PERIOD_PROPERTY = "sonarlint.internal.flight.recorder.interval.seconds";
+  private static final SonarLintLogger LOG = SonarLintLogger.get();
+  private static final int DEFAULT_15_MINUTES = 900;
```
nquinquenel, line 36:
```suggestion
  private static final int DEFAULT_INTERVAL_SECONDS = 900;
```
What about this more generic name?

nquinquenel (commented):
My main remark is that this FlightRecorderService class will never be used in 99.9% of the time. As far I know, with this implementation we are always creating the executor service (unless I'm mistaken). It might not be in the scope yet to handle this, but I would avoid starting the executor service eagerly unless the right property/params are provided. So I'd probably start the executor service in the launch() method after a check whether the feature is enabled or not (TODO in another ticket I suppose).

Otherwise good 😄

PR #1485, opened 2025-09-01
SLCORE-1615 add telemetry for didChangeAutomaticAnalysisSetting

sophio-japharidze-sonarsource:


backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisService.java
```diff
@@ -716,6 +719,7 @@ public void didChangeAutomaticAnalysisSetting(boolean enabled) {
     if (!previouslyEnabled) {
       triggerAnalysisForOpenFiles();
     }
+    telemetryService.automaticAnalysisSettingToggled();
```
damien-urruty-sonarsource: Instead of introducing a dependency on the telemetry service, I would suggest to use the event publisher, to decouple things a bit. Telemetry is a secondary concern compared to business logic, so in most places we manage it with events, and the telemetry service reacts to those events

damien-urruty-sonarsource (commented):
Looks mostly good but I think the design should be changed

backend/core/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryService.java
```diff
@@ -369,6 +377,11 @@ public void onIssuesRaised(IssuesRaisedEvent event) {
     updateTelemetry(localStorage -> localStorage.addIssuesWithPossibleAiFixFromIde(issuesToReport));
   }
 
+  @EventListener
+  public void onAutomaticAnalysisSettingChanged(AutomaticAnalysisSettingChangedEvent event) {
+    automaticAnalysisSettingToggled();
```
damien-urruty-sonarsource, line 382: I think it's fine to manage the update with a toggle, but then we need to make sure the event is emitted only when the value actually changed, which I think is not the case

PR #1505, opened 2025-09-22
SLCORE-1631 Migrate to GitHub Actions

eray-felek-sonarsource:
[SLCORE-1685](https://sonarsource.atlassian.net/browse/SLCORE-1685)

[SLCORE-1685]: https://sonarsource.atlassian.net/browse/SLCORE-1685?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/build.yml
```diff
@@ -0,0 +1,283 @@
...
+      - uses: SonarSource/ci-github-actions/build-maven@v1
+        id: build
+        with:
+          deploy-pull-request: true
+          artifactory-reader-role: private-reader
+          artifactory-deployer-role: qa-deployer
+          maven-args: -P dist-no-arch,dist-windows_x64,dist-linux_x64,dist-linux_aarch64,dist-macosx_x64,dist-macosx_aarch64 -Dmaven.test.skip=true -Dsonar.skip=true
+      - name: Config Maven (cache setup)
+        run: |
+          mvn -B -e -V -Pits dependency:go-offline # populate cache including ITs deps too
```
damien-urruty-sonarsource, line 39: I don't understand what this does, are you preparing some cache for the next QA tasks? I think we cannot assume QAs will run on the same runner

.github/workflows/build.yml
```diff
@@ -0,0 +1,283 @@
...
+      - name: debug
+        if: failure()
+        shell: bash
+        run: |
+          echo "=== Listing surefire-reports contents ==="
+          find ./its/tests/target/surefire-reports -type f || true
+          echo "=== Checking if directory is empty ==="
+          [ -d ./its/tests/target/surefire-reports ] && ls -la ./its/tests/target/surefire-reports/ || echo "Directory doesn't exist"
+
+  test-windows:
```
damien-urruty-sonarsource: I would recommend to move this job upper in the file, next to test-linux, because they belong together

.github/workflows/build.yml
```diff
@@ -0,0 +1,283 @@
...
+        id: secrets-gh
+        uses: SonarSource/vault-action-wrapper@320bd31b03e5dacaac6be51bbbb15adf7caccc32 # v3.1.0
+        with:
+          secrets: |
+            development/github/token/licenses-ro token | GITHUB_TOKEN;
+
+      - uses: SonarSource/ci-github-actions/config-maven@v1
+        with:
+          artifactory-reader-role: private-reader
+      - name: Run QA (direct)
```
damien-urruty-sonarsource: What do you mean by 'Direct' here?

.github/workflows/build.yml
```diff
@@ -0,0 +1,283 @@
...
+        if: ${{ github.event_name == 'pull_request' || github.ref_name == github.event.repository.default_branch || startsWith(github.ref_name, 'branch-') || startsWith(github.ref_name, 'dogfood-on-') }}
+        env:
+          MAVEN_OPTS: -Xmx4g
+          SONARCLOUD_IT_TOKEN: ${{ steps.secrets-sc.outputs.vault && fromJSON(steps.secrets-sc.outputs.vault).SONARCLOUD_IT_TOKEN || '' }}
+          SONARCLOUD_REGION: ${{ matrix.sc && matrix.region || '' }}
+          BUILD_NUMBER: ${{ needs.build.outputs.build_number }}
+          GITHUB_TOKEN: ${{ fromJSON(steps.secrets-gh.outputs.vault).GITHUB_TOKEN }}
+        run: |
+          mvn -f its/pom.xml -Dsonar.runtimeVersion=${{ matrix.sq_version }} ${{ matrix.category }} verify surefire-report:report
+      - name: Run QA (for else)
```
damien-urruty-sonarsource: And what do you mean by 'for else' here?

.github/workflows/build.yml
eray-felek-sonarsource: In the qa_script before, whe had if else condition, this is how I replicated that behaviour

.github/workflows/build.yml
eray-felek-sonarsource: referenced in the comment above

.github/workflows/build.yml
eray-felek-sonarsource: image
Hi, I discussed this with eng-exp and it is indeed re-used by other runners even if it is not the same runner

PR #1511, opened 2025-09-24
SLCORE-1625 Prevent setting up a SQ:S connection with a SQ:C URL

Krosovok:
[SLCORE-1625](https://sonarsource.atlassian.net/browse/SLCORE-1625)

[SLCORE-1625]: https://sonarsource.atlassian.net/browse/SLCORE-1625?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/core/src/main/java/org/sonarsource/sonarlint/core/embedded/server/filter/ParseParamsFilter.java
```diff
@@ -0,0 +1,56 @@
...
+import org.apache.hc.core5.http.ClassicHttpRequest;
+import org.apache.hc.core5.http.HttpException;
+import org.apache.hc.core5.http.NameValuePair;
+import org.apache.hc.core5.http.io.HttpFilterChain;
+import org.apache.hc.core5.http.io.HttpFilterHandler;
+import org.apache.hc.core5.http.protocol.HttpContext;
+import org.apache.hc.core5.net.URIBuilder;
+import org.sonarsource.sonarlint.core.embedded.server.AttributeUtils;
+
+public class ParseParamsFilter implements HttpFilterHandler {
```
damien-urruty-sonarsource, line 37: I think introducing this `ParseParamsFilter` is a good first step in the direction of simplifying handlers and factorizing code. I think we could go even further:
* There is one 'smell' that I encounter often and that I don't like, it's the usage of 'Utils' classes. I always try to avoid them. It usually means we didn't really find a better place, or rather didn't find a meaningful concept(s) to hold this data and behavior together.
*  I don't like the adherence to Apache HTTP in every handler. The fact that we use an attribute to transport the parsed parameters goes further into the integration with Apache HTTP. 
* I think it's also somehow misusing filters, that are usually more about stopping or continuing handling of a request

In our case, I think we could build our own Request class, that would encapsulate the Apache request object and ease access to the origin or request parameters. But it's a bit more work, so it doesn't have to be implemented now, it's just my 2cents

backend/core/src/main/java/org/sonarsource/sonarlint/core/embedded/server/filter/ValidationFilter.java
```diff
@@ -0,0 +1,63 @@
...
+  private final SonarLintRpcClient client;
+
+  public ValidationFilter(SonarLintRpcClient client) {
+    this.client = client;
+  }
+
+  @Override
+  public void handle(ClassicHttpRequest request, ResponseTrigger responseTrigger, HttpContext context, HttpFilterChain chain) throws HttpException, IOException {
+    var params = AttributeUtils.getParams(context);
+    if (params.containsKey("server")) {
```
damien-urruty-sonarsource, line 50: I think the usage of a filter is more adapted here, but the implementation is a bit risky. Imagine later we introduce a new API that accepts a server parameter, but that represents something else. The fact that this `ValidationFilter` is unconditionally executed would fail such a case. I see 2 ways to improve that:
* also check the request path in the filter, to check the server parameter only for certain paths. But this can also be risky if we add a new 'open in IDE' handler
* have this check done in a separate place, maybe in a common abstract handler on top of other 'open in ide' handlers. Not a big fan of abstract things either, maybe there are other solutions

damien-urruty-sonarsource (approved):
LGTM, what you did should work. I am a bit concerned with the usage of a filter for checking the server Origin with query param. I left you some suggestions to improve this even further

backend/core/src/main/java/org/sonarsource/sonarlint/core/embedded/server/filter/ParseParamsFilter.java
Krosovok: Thank you! For now I am merging as is, but let's talk today about further improvements.

PR #1512, opened 2025-09-24
SLCORE-1690 Cache native Git command availability

damien-urruty-sonarsource:
[SLCORE-1690](https://sonarsource.atlassian.net/browse/SLCORE-1690)

[SLCORE-1690]: https://sonarsource.atlassian.net/browse/SLCORE-1690?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/NativeGitLocator.java
```diff
@@ -23,18 +23,44 @@
...
+    if (checkedForNativeGitExecutable) {
+      return Optional.ofNullable(nativeGitExecutable);
+    }
+    var nativeGit = getGitExecutableName().map(gitExecutable -> {
+      var git = new NativeGit(gitExecutable);
+      if (git.version().filter(version -> version.satisfiesMinRequirement(MINIMUM_REQUIRED_GIT_VERSION)).isPresent()) {
+        return git;
+      }
+      return null;
+    });
```
Krosovok: Personally, I would prefer moving whole if condition to a private static function and omit this way `if` and both `returns` (as they will be basically handled by Optional API)

```suggestion
    var nativeGit = getGitExecutableName()
      .map(NativeGit::new)
      .filter(NativeGitLocator::hasRequiredVersion);
```

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/NativeGit.java
```diff
@@ -0,0 +1,92 @@
...
+import org.sonarsource.sonarlint.core.commons.Version;
+import org.sonarsource.sonarlint.core.commons.log.SonarLintLogger;
+import org.sonarsource.sonarlint.core.commons.util.FileUtils;
+
+import static java.lang.String.format;
+
+public class NativeGit {
+  private static final SonarLintLogger LOG = SonarLintLogger.get();
+  private static final String GIT_VERSION_OUTPUT_PREFIX = "git version";
+  private static final Period blameHistoryWindow = Period.ofDays(365);
```
Krosovok: `BLAME_HISTORY_WINDOW` would fit better here as it's basically a constant.

backend/commons/src/test/java/org/sonarsource/sonarlint/core/commons/util/git/NativeGitLocatorTests.java
```diff
@@ -19,40 +19,68 @@
...
+  private Path projectDirPath;
+
+  @BeforeEach
+  void prepare() {
+    underTest = spy(new NativeGitLocator());
+  }
+
+  @AfterEach
+  void cleanup() throws IOException {
+    org.eclipse.jgit.util.FileUtils.delete(projectDirPath.toFile(), RECURSIVE);
```
Krosovok: Full name is not needed here.

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/NativeGit.java
```diff
@@ -0,0 +1,106 @@
...
+        if (version.length > 1) {
+          try {
+            // keep only MAJOR and MINOR numbers, it's sufficient for checking support
+            return Optional.of(Version.create(version[0] + "." + version[1]));
+          } catch (Exception e) {
+            // error will be logged below
+          }
+        }
+      }
+    }
```
Krosovok: I would suggest doing it with stream/optional API to reduce nesting and checking, creating a sequential path to the result. `findFirst` provides exactly what we need and we keep all of `Optional.of()` and stuff like that under the hood, creating less noise.

```suggestion
    return lines.stream().findFirst()
      .map(String::trim)
      .filter(line -> line.startsWith(GIT_VERSION_OUTPUT_PREFIX))
      .map(line -> Strings.CI.removeStart(line, GIT_VERSION_OUTPUT_PREFIX))
      .map(String::trim)
      .map(actualVersion -> actualVersion.split("\\.", 3))
      .filter(versionParts -> versionParts.length > 1)
      .map(NativeGit::createVersion); // try/catch part to a different method to avoid noise here
```

Krosovok (approved):
LGTM! I added a few suggestions (mostly on stream/optional API), but they are more of a style choice.

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/NativeGitLocator.java
damien-urruty-sonarsource: Very good idea, I like it!

backend/commons/src/test/java/org/sonarsource/sonarlint/core/commons/util/git/NativeGitLocatorTests.java
damien-urruty-sonarsource: Yes

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/NativeGit.java
damien-urruty-sonarsource: Yes

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/NativeGit.java
damien-urruty-sonarsource: I tried your approach, the end result is maybe even more verbose than before but I like the style, so I will keep it. Another good idea!

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/git/NativeGit.java
```diff
@@ -39,38 +39,49 @@
...
-      }
+    var version = lines.stream().findFirst()
+      .map(String::trim)
+      .filter(line -> line.startsWith(GIT_VERSION_OUTPUT_PREFIX))
+      .map(line -> line.substring(GIT_VERSION_OUTPUT_PREFIX.length()))
+      .map(String::trim)
+      .map(actualVersion -> actualVersion.split("\\.", 3))
+      .filter(versionParts -> versionParts.length > 1)
+      .flatMap(NativeGit::tryCreateVersion);
+    if (version.isPresent()) {
```
Krosovok: `isAbsent`?

damien-urruty-sonarsource: ahh good catch!

PR #1513, opened 2025-09-24
SLCORE-1657 Fix fuzzy search to properly work with underscores

Krosovok:
[SLCORE-1657](https://sonarsource.atlassian.net/browse/SLCORE-1657)

[SLCORE-1657]: https://sonarsource.atlassian.net/browse/SLCORE-1657?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/API_CHANGES.md
+++ b/API_CHANGES.md
@@ -9,6 +9,10 @@
 * Add a new `getMCPServerConfiguration` method to `org.sonarsource.sonarlint.core.rpc.protocol.backend.connection.ConnectionRpcService`
   * It accepts `connectionId` and `token` as parameters
   * It returns JSON string containing MCP server settings (without the `sonarqube` parent item)
+* Introduce an RPC notification `embeddedServerStarted` in `org.sonarsource.sonarlint.core.rpc.protocol.SonarLintRpcClient`
+  * It is sent by the backend to notify the client that the embedded server has started
+  * It contains the embedded server port
+  * Example usage is by the MCP Server to establish the bridge connection
 
 # 10.31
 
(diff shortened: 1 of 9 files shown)


backend/core/src/main/java/org/sonarsource/sonarlint/core/TextSearchIndex.java
```diff
@@ -45,18 +45,15 @@
...
 public class TextSearchIndex<T> {
-  private static final String DEFAULT_SPLIT_PATTERN = "\\W";
-  private final String splitPattern;
-  private TreeMap<String, List<DictEntry>> termToObj;
-  private Map<T, Integer> objToWordFrequency;
 
-  public TextSearchIndex() {
-    this(DEFAULT_SPLIT_PATTERN);
-  }
+  private static final String DEFAULT_SPLIT_PATTERN = "[\\W_]+";
```
nquinquenel: The format does not have to contain an underscore, this is automatically generated sometimes like this, but users are free to put the name they want. I don't think it's necessarily the right split pattern.

Krosovok: I am not sure I understood you. The idea is of the ticket is to add split by the underscore so that search can work correctly with names like `SonarSource_peachee-dotnet`.
If you are talking about leaving constructor with a parameter pattern, I don't see much value in leaving it if it's not gonna be used anywhere. We can always add it back if we need.

nquinquenel: Ok nevermind my previous comment, this was maybe not relevant.

I think this works, but for clarity I personally prefer `[^a-zA-Z0-9]`, which essentially resolves to the same pattern. I'm also not sure the `+` is needed at the end

Krosovok: `[^a-zA-Z0-9]` makes sense as well. `+` is needed so that we don't get a bunch of unnecessary (empty) groups when trying to split something with several ssplit symbols in a row, like `test__item` would be just 2 groups instead of 3 (one of them being nothingness between two underscores).

nquinquenel (approved):

PR #1591, opened 2025-11-21
SLCORE-1849 Support hooks for Cursor

nquinquenel:


voz:
@nquinquenel can we close it or define path forward if we need to? Thx for helping to keep things tidy.

PR #1620, opened 2025-12-09
SLCORE-1865 Print logs to disk

damien-urruty-sonarsource:


voz:
@damien-urruty-sonarsource can we close it or define path forward if we need to? Thx for helping to keep things tidy.

PR #1630, opened 2025-12-16
SLCORE-1432 Introduce a cache for file contents

Krosovok:
I didn't notice any performance boosts, but it should help us with making sure that files are not staying open as `inputStream()` implementation now relies on in-memory copy instead of going straight to the file.

Feel free to discuss default cache size as it did lead to almost a 50/50 between hits and misses in my test run. But even after increasing the size I didn't notice performance boost.

### DIFF
--- a/API_CHANGES.md
+++ b/API_CHANGES.md
@@ -4,6 +4,12 @@
 
 * Remove the `org.sonarsource.sonarlint.core.rpc.protocol.backend.sca.DependencyRiskRpcService.getDependencyRiskDetails` method and associated DTOs.
 
+## New features
+
+* Added caching for file contents during analysis. Following environment variables are used to control it: 
+  * `CACHE_STATS_ENABLED_PROPERTY` - when `true`, it enables cache stats gathering (like amount of cache hits/misses and evictions caused by overflow).
+  * `SONARLINT_CACHE_SIZE` - to set another cap to total length of strings in the cache, 1.000.000 by default.
+
 # 10.38
 
 ## New features
--- a/backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/BackendInputFile.java
+++ b/backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/BackendInputFile.java
@@ -19,10 +19,10 @@
  */
 package org.sonarsource.sonarlint.core.analysis;
 
-import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
 import java.nio.charset.Charset;
+import org.apache.commons.io.IOUtils;
 import org.sonarsource.sonarlint.core.analysis.api.ClientInputFile;
 import org.sonarsource.sonarlint.core.commons.api.SonarLanguage;
 import org.sonarsource.sonarlint.core.commons.util.FileUtils;
@@ -56,8 +56,8 @@ public class BackendInputFile implements ClientInputFile {
   }
 
   @Override
-  public InputStream inputStream() throws IOException {
-    return clientFile.inputStream();
+  public InputStream inputStream() {
+    return IOUtils.toInputStream(contents(), getCharset());
   }
 
   @Override
--- a/backend/core/src/main/java/org/sonarsource/sonarlint/core/fs/ClientFileSystemService.java
+++ b/backend/core/src/main/java/org/sonarsource/sonarlint/core/fs/ClientFileSystemService.java
@@ -58,20 +58,22 @@ public class ClientFileSystemService {
   private final TelemetryService telemetryService;
   private final SmartCancelableLoadingCache<String, Map<URI, ClientFile>> filesByConfigScopeIdCache =
     new SmartCancelableLoadingCache<>("sonarlint-filesystem", this::initializeFileSystem);
+  private final ReadThroughFileCache readThroughFileCache;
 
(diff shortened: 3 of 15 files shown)


backend/core/src/main/java/org/sonarsource/sonarlint/core/fs/ClientFileSystemService.java
```diff
@@ -140,6 +144,7 @@ public void didUpdateFileSystem(DidUpdateFileSystemParams params) {
       if (clientFile != null) {
         filesByConfigScopeIdCache.get(clientFile.getConfigScopeId()).remove(uri);
         removed.add(clientFile);
+        readThroughFileCache.remove(Path.of(uri), clientFile.getCharset());
```
damien-urruty-sonarsource, line 147: A URI doesn't necessarily represent a local path, so this conversion might throw. Usually we try to reference files via their URI in the file system service, couldn't you do the same here?

backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/BackendInputFile.java
```diff
@@ -56,8 +56,8 @@ public ClientFile getClientObject() {
   }
 
   @Override
-  public InputStream inputStream() throws IOException {
-    return clientFile.inputStream();
+  public InputStream inputStream() {
+    return IOUtils.toInputStream(contents(), getCharset());
```
damien-urruty-sonarsource, line 60: What is the reason for this change?

pom.xml
```diff
@@ -276,6 +276,16 @@
...
+      <dependency>
+        <groupId>com.github.ben-manes.caffeine</groupId>
+        <artifactId>caffeine</artifactId>
+        <version>3.2.3</version>
+      </dependency>
+      <dependency>
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot-starter-cache</artifactId>
+        <version>4.0.0</version>
+      </dependency>
```
damien-urruty-sonarsource, line 288: I am a bit worried to pull spring-boot stuff, mostly for dependency risks that it might bring in the future. Also we probably use very little from it. Would it be hard to not rely on it?

API_CHANGES.md
```diff
@@ -4,6 +4,12 @@
 
 * Remove the `org.sonarsource.sonarlint.core.rpc.protocol.backend.sca.DependencyRiskRpcService.getDependencyRiskDetails` method and associated DTOs.
 
+## New features
+
+* Added caching for file contents during analysis. Following environment variables are used to control it: 
+  * `CACHE_STATS_ENABLED_PROPERTY` - when `true`, it enables cache stats gathering (like amount of cache hits/misses and evictions caused by overflow).
```
damien-urruty-sonarsource, line 10: Could you also prefix this env var with `SONARLINT_`? To reduce the chance to collide with other variables

damien-urruty-sonarsource (commented):
I have some comments about implementation details. But my main concern is about performance. If it doesn't improve performance, I would be reluctant to merge this relatively complex change. My original idea was that caching would help both performance and unreleased file handles, killing 2 birds with one stone. If it's not the case, then maybe there is a simpler solution to treat the problems of unreleased file handles alone? And kill one bird with one smaller stone ? :)

Krosovok:
> I have some comments about implementation details. But my main concern is about performance. If it doesn't improve performance, I would be reluctant to merge this relatively complex change. My original idea was that caching would help both performance and unreleased file handles, killing 2 birds with one stone. If it's not the case, then maybe there is a simpler solution to treat the problems of unreleased file handles alone? And kill one bird with one smaller stone ? :)

Yes, that was what I thought too. The thing is that in my tests performance was vastly inconsistent even for the same build (with or without caching). And the caching approach seems the same, but it's really hard to say if there would be a change in average. 

Unreleased file handles alone were partially fixed in scope of [the previous ticket](https://sonarsource.atlassian.net/browse/SLCORE-1870). The only thing that it doesn't handle is making sure that `inputStream` that we give away does not lead to an actual file content but to a in-memory string instead. But if we do only that (without caching) then I suspect it would lead to decreased performance.

So from my perspective we should either:
1) Discard these changes (then hypothetically some not closed input streams from analyzers can hold files)
2) Try with cache and see a broader effect on performance (we have telemetry for analysis time, right?)

backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/BackendInputFile.java
Krosovok: I removed `inputStream` from `clientFile` and opened input streams on in-memory strings (cached or new) to make sure that input stream never actually leads to an not closed file.

pom.xml
Krosovok: I do think that it generally provides value (it basically allows not no mix caching with logic) but in this case as it was all focused in one place I think it can be removed with minimal impact.

Krosovok:
As agreed with @damien-urruty-sonarsource, closing this one as caching didn't yield performance boost.
To be followed up with investigation tickets to address underlying problems and possibly return to caching solution later: 
https://sonarsource.atlassian.net/browse/SLCORE-1928
https://sonarsource.atlassian.net/browse/SLCORE-1929

PR #1888, opened 2026-03-10
SLCORE-2061 SLCORE-2062 [WIP] On-demand analyzers.

Krosovok:
Adding on-demand fallback for CFamily and C# analyzers. If they are not provided by the IDE, they will be downloaded from binaries.

I moved good chunk of plugin resolution logic to a new `PluginArtifactProvider` that contains a chain of resolution for plugins. It handles:
Connected mode - synchronized with SQS/SQC,
Embedded - provided by IDE,
On-demand - download from binaries when supported

This way we clearly know where the plugin came from (or why is it not active). Download happens asynchronously and then status is updated with events.

What is WIP: ANSIBLE/GITHUBACTIONS status is tricky to figure out as they are either iac or iacenterprise depending on server version.

What I want to do separately:
- retry mechanism
- `overriddenVersion` field - for now it's always empty


backend/core/src/main/java/org/sonarsource/sonarlint/core/plugin/PluginArtifactProvider.java
```diff
@@ -0,0 +1,355 @@
...
+    var serverPluginsExpectedInStorage = computeExpectedInStorage(serverPluginList, storedPluginsByKey);
+
+    var enabledLanguages = languageSupportRepository.getEnabledLanguagesInConnectedMode();
+    var disabledPluginKeys = computeDisabledPluginKeys(enabledLanguages);
+
+    enabledLanguages.forEach(language -> resolveSync(language, connectionId));
+
+    serverPluginList.stream()
+      .filter(plugin -> shouldSyncPlugin(plugin, storedPluginsByKey))
+      .forEach(plugin -> connectedModeArtifactResolver.downloadPluginSync(connectionId, plugin));
```
nquinquenel: You need to refetch `storedPluginsByKey` after you do `resolveSync`, otherwise I think you will download everything again on the `shouldSyncPlugin`

PR #1926, opened 2026-03-18
SLCORE-2244 Implement IDESupportedLanguageViewed Gessie telemetry event

nquinquenel:


backend/core/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryServerAttributesProvider.java
```diff
@@ -157,4 +160,34 @@ private boolean hasDisableNotifications(BoundScope binding) {
...
+   * Returns connection-related user information for Gessie events, resolved from the effective binding
+   * of the given config scope ID. Returns empty if the scope is not bound.
+   */
+  public Optional<GessieConnectionInfo> getGessieConnectionInfo(String configScopeId) {
+    return configurationRepository.getEffectiveBinding(configScopeId).flatMap(binding -> {
+      var connectionId = binding.connectionId();
+      var connection = connectionConfigurationRepository.getConnectionById(connectionId);
+      if (connection == null) {
+        return Optional.empty();
+      }
```
Krosovok: Since we are already in optional chain, I'd move out this to a separate `map`. Then null check would be redundant.

Can be two maps for `connectionId` and `connection` or lambda/method with both.

backend/core/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryServerAttributesProvider.java
```diff
@@ -157,4 +160,34 @@ private boolean hasDisableNotifications(BoundScope binding) {
...
+      var storage = storageService.connection(connectionId);
+      var userUuid = nullIfEmpty(storage.user().read().orElse(null));
+      if (connection instanceof SonarCloudConnectionConfiguration) {
+        var organizationUuidV4 = storage.organization().read().map(Organization::uuidV4).map(Object::toString).orElse(null);
+        return Optional.of(new GessieConnectionInfo(ConnectionType.SQC, userUuid, organizationUuidV4, null));
+      } else if (connection instanceof SonarQubeConnectionConfiguration) {
+        var sqsInstallationId = nullIfEmpty(storage.serverInfo().read().map(StoredServerInfo::serverId).orElse(null));
+        return Optional.of(new GessieConnectionInfo(ConnectionType.SQS, userUuid, null, sqsInstallationId));
+      }
+      return Optional.empty();
```
Krosovok: I'd move it to a separate method returning null in the end.

It would clean up the chain and remove Optional.of/empty from actual logic.

backend/core/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryService.java
```diff
@@ -414,6 +426,29 @@ public void close() {
     }
   }
 
+  public void supportedLanguageViewed(String configScopeId) {
+    if (!gessieService.isEnabled()) {
+      return;
+    }
+    var localUserId = monitoringUserIdStore.getOrCreate().map(Object::toString).orElse(null);
+    if (localUserId == null) {
```
Krosovok: Optional.isAbsent instead of null-check

Claude Code loved doing that so I added this to CLAUDE.md and honestly it helped a lot:
```
## Java Optional
- Prefer Optional chains over multiple null checks when Optional is already in use
- When a lambda body spans more than one line or contains more than one method/constructor invocation, extract it to a named private method; pass it as a method reference or call it from the lambda
```

backend/core/src/test/java/org/sonarsource/sonarlint/core/telemetry/TelemetryServiceTests.java
```diff
@@ -124,4 +147,80 @@ void should_increment_properties_file_counter_for_assisted_properties_file() {
...
+    underTest.supportedLanguageViewed("scopeId");
+
+    var captor = ArgumentCaptor.forClass(IDESupportedLanguageViewedPayload.class);
+    verify(gessieService).supportedLanguageViewed(captor.capture());
+    var payload = captor.getValue();
+    assertThat(payload.localUserId()).isEqualTo("aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee");
+    assertThat(payload.connectionType()).isNull();
+    assertThat(payload.userUuid()).isNull();
+    assertThat(payload.organizationUuidV4()).isNull();
+    assertThat(payload.sqsInstallationId()).isNull();
```
Krosovok: Similar thing here. Claude loves to have asserts per-field for each field of the object, that doesn't make much sense to me (unless it's just a couple of fields out of a dozen).

It is much more of a preference but it does raise Sonar issues with > 10~15 asserts.

So I asked it that:
```
## Java Tests - Assertions
- In the Assert phase, construct an expected object (record, DTO, POJO) in Arrange and compare against it — don't assert individual fields separately
```

medium-tests/src/test/java/mediumtest/gessie/GessieSupportedLanguageViewedMediumTests.java
```diff
@@ -0,0 +1,149 @@
...
+    System.setProperty("sonarlint.http.retry.interval.seconds", "0");
+    System.setProperty(GessieSpringConfig.PROPERTY_GESSIE_API_KEY, "value");
+    gessieEndpointMock.stubFor(post(IDE_ENDPOINT).willReturn(aResponse().withStatus(202)));
+  }
+
+  @AfterEach
+  void tearDown() {
+    System.clearProperty("sonarlint.http.retry.interval.seconds");
+    System.clearProperty(GessieSpringConfig.PROPERTY_GESSIE_API_KEY);
+  }
```
Krosovok: Can use `@SystemStubs` for that

backend/core/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryService.java
```diff
@@ -414,6 +426,29 @@ public void close() {
     }
   }
 
+  public void supportedLanguageViewed(String configScopeId) {
+    if (!gessieService.isEnabled()) {
```
Krosovok: Not sure about this one, but a food for thought.

I am not sure if this method belongs here. Yes, it is telemetry, but Gessie is also a separate thing. And it also makes you expose `gessieService.isEnabled()` to the outside.

It would make more sense for the method to be in GessieService (or some new class that would use it and have `isEnabled` logic there) and use it directly in `TelemetryRpcServiceDelegate`. So it's all telemetry but implementations are different and not reference each other.

medium-tests/src/test/java/mediumtest/gessie/GessieSupportedLanguageViewedMediumTests.java
nquinquenel: Right, that's better!

backend/core/src/main/java/org/sonarsource/sonarlint/core/telemetry/TelemetryService.java
nquinquenel: I kinda agree, I still want to have this enablement check in this method to avoid computing all this if telemetry is disabled.

I created a dedicated `GessieEventPublisher` class, to support the logic of such events related to Gessie, to avoid having them in `TelemetryService` class.

`GessieService` is focused on the HOW to send event, while `GessieEventPublisher` focuses on WHAT data to gather.

backend/core/src/test/java/org/sonarsource/sonarlint/core/telemetry/TelemetryServiceTests.java
nquinquenel: I agree, makes sense with relatively small object!

Krosovok (approved):

PR #1975, opened 2026-04-09
SLCORE-2062 Introduce standalone artifacts loading strategy

damien-urruty-sonarsource:
Add the ArtifactSource interface, the standalone source implementations (EmbeddedPluginSource, BinariesArtifactSource) and their supporting infrastructure (ArtifactState, ArtifactOrigin, ResolvedArtifact, …), together with StandaloneArtifactsLoadingStrategy and ArtifactsLoadingStrategy interface.
Renames commons Plugin → EnterpriseReplacement to reflect its meaning.


backend/core/src/main/java/org/sonarsource/sonarlint/core/plugin/loading/strategy/ArtifactsLoadingResult.java
```diff
@@ -0,0 +1,56 @@
...
+import org.sonarsource.sonarlint.core.plugin.source.ResolvedArtifact;
+
+public record ArtifactsLoadingResult(Set<SonarLanguage> enabledLanguages, Map<String, ResolvedArtifact> resolvedArtifactsByKey) {
+  public void whenAllArtifactsDownloaded(Runnable runnable) {
+    var pendingDownloads = resolvedArtifactsByKey.values().stream().map(ResolvedArtifact::downloadFuture)
+      .filter(Objects::nonNull)
+      .toList();
+    if (pendingDownloads.isEmpty()) {
+      return;
+    }
```
nquinquenel, line 38: I think there's a logic issue in this method. We set a Runnable to run when all artifacts are downloaded, but if they are already downloaded (maybe cached for example?), it just returns and never runs the Runnable. Is it expected?

backend/core/src/main/java/org/sonarsource/sonarlint/core/plugin/source/UniqueTaskExecutor.java
```diff
@@ -17,38 +17,35 @@
...
+  public Future<?> scheduleIfAbsent(String key, Runnable task) {
+    return inProgress.computeIfAbsent(key, k -> {
       var logOutput = SonarLintLogger.get().getTargetForCopy();
-      executor.submit(() -> {
-        var previousLogOutput = SonarLintLogger.get().getTargetForCopy();
-        if (logOutput != null) {
-          SonarLintLogger.get().setTarget(logOutput);
-        }
+      return executor.submit(() -> {
+        SonarLintLogger.get().setTarget(logOutput);
```
nquinquenel, line 41: Is it expected to never clear the log target? In `AbstractRpcServiceDelegate` we set it back to null afterwards, could it leak `LogOutput`?

backend/core/src/main/java/org/sonarsource/sonarlint/core/plugin/source/ResolvedArtifact.java
```diff
@@ -0,0 +1,32 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.sonarlint.core.plugin.source;
+
+import java.nio.file.Path;
+import java.util.concurrent.Future;
+import javax.annotation.Nullable;
+import org.sonarsource.sonarlint.core.commons.Version;
+
+public record ResolvedArtifact(ArtifactState state, @Nullable Path path, @Nullable ArtifactOrigin source, @Nullable Version version,
```
nquinquenel, line 27: There's another `ResolvedArtifact` record in the code base, is it temporary? Otherwise it might be confusing (what's the difference between the two?)

backend/core/src/main/java/org/sonarsource/sonarlint/core/plugin/loading/strategy/StandaloneArtifactsLoadingStrategy.java
```diff
@@ -0,0 +1,91 @@
...
+ * {@link ArtifactState#PREMIUM} when no other source can provide them.</p>
+ */
+public class StandaloneArtifactsLoadingStrategy implements ArtifactsLoadingStrategy {
+
+  private final LanguageSupportRepository languageSupportRepository;
+  private final List<ArtifactSource> artifactSourcesSortedByAscendingPriority;
+
+  public StandaloneArtifactsLoadingStrategy(InitializeParams params, BinariesArtifactSource binariesSource, LanguageSupportRepository languageSupportRepository) {
+    // Ascending priority: binaries < embedded. Later source overwrites for the same key
+    this.artifactSourcesSortedByAscendingPriority = List.of(binariesSource, EmbeddedPluginSource.forStandalone(params));
```
nquinquenel: `forStandalone` is called during the constructor, which then call `buildPluginKeyToPathMap` which reads JAR with `fromJar`. Is it safe to do from the constructor? Also if the JAR is corrupted it will crash the init I think

backend/core/src/test/java/org/sonarsource/sonarlint/core/plugin/source/binaries/BinariesArtifactSourceTest.java
```diff
@@ -0,0 +1,262 @@
...
+    mockBlockingHttpClient(proceedLatch);
+    var source = buildSource();
+    try {
+      var result = source.load("cpp");
+
+      assertThat(result)
+        .isPresent()
+        .get() // Extracts the value from the Optional
+        .usingRecursiveComparison()
+        .ignoringFields("downloadFuture") // Replace with the field name you want to skip
```
nquinquenel: The comments look bit overkill 😅 
Same below

backend/core/src/main/java/org/sonarsource/sonarlint/core/plugin/loading/strategy/ArtifactsLoadingResult.java
damien-urruty-sonarsource: Yes. This callback exists so we can reload plugins when all downloads are over. But in all cases, plugins that are already present on disk  are loaded once just after calling this method

backend/core/src/main/java/org/sonarsource/sonarlint/core/plugin/source/UniqueTaskExecutor.java
damien-urruty-sonarsource: I'm not sure, I added it back just in case

backend/core/src/main/java/org/sonarsource/sonarlint/core/plugin/source/ResolvedArtifact.java
damien-urruty-sonarsource: It is there for the previous "resolvers" abstraction, it is removed together with this concept in the last PR of this series

backend/core/src/test/java/org/sonarsource/sonarlint/core/plugin/source/binaries/BinariesArtifactSourceTest.java
damien-urruty-sonarsource: I'm so dumb

backend/core/src/main/java/org/sonarsource/sonarlint/core/plugin/loading/strategy/StandaloneArtifactsLoadingStrategy.java
damien-urruty-sonarsource: Correct. I will load it once lazily from the resolve method

nquinquenel (approved):

PR #1977, opened 2026-04-09
SLCORE-2062 Wire artifacts loading strategies into PluginsService

damien-urruty-sonarsource:
Replace the old resolver chain (EmbeddedArtifactResolver, ConnectedModeArtifactResolver, …) with the new
ArtifactsLoadingStrategy / ArtifactSource abstraction in PluginsService and SonarLintSpringAppConfig. Delete all old resolver classes and update AnalysisService, LanguageSupportRepository, PluginStatus and the medium tests accordingly.


backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisService.java
```diff
@@ -618,11 +618,7 @@ private AnalyzeCommand getAnalyzeCommand(String configurationScopeId, Set<URI> f
       () -> getAnalysisConfigForEngine(configurationScopeId, files, Map.of(), hotspotsOnly, triggerType, trace),
       issue -> streamIssue(configurationScopeId, analysisId, rawIssues, issue), trace,
       new SonarLintCancelMonitor(), taskManager, inputFiles -> analysisStarted(configurationScopeId, analysisId, inputFiles),
-      () -> {
-        var isScopeReady = analysisReadinessByConfigScopeId.getOrDefault(configurationScopeId, false);
-        var connectionId = configurationRepository.getEffectiveBinding(configurationScopeId).map(Binding::connectionId).orElse(null);
-        return isScopeReady && !pluginsService.areAnyPluginsDownloading(connectionId);
-      }, files, Map.of());
+      () -> analysisReadinessByConfigScopeId.getOrDefault(configurationScopeId, false), files, Map.of());
```
nquinquenel, line 622: Maybe I don't have the full picture correctly, but the analysis will start regardess of analyzers being currently downloaded? Is it ok?

There's a callback with `whenAllArtifactsDownloaded` but it happens after downloads are finished

backend/core/src/main/java/org/sonarsource/sonarlint/core/plugin/PluginsService.java
```diff
@@ -155,70 +159,30 @@ public LoadedPlugins getPlugins(String connectionId) {
   }
 
   private PluginsLoadResult loadPlugins(@Nullable String connectionId) {
-    var pluginsToLoadByKey = new HashMap<String, Path>();
-    var enabledLanguages = EnumSet.noneOf(SonarLanguage.class);
+    var strategy = getPluginLoadingStrategy(connectionId);
+    var result = strategy.resolveArtifacts();
+    result.whenAllArtifactsDownloaded(() -> eventPublisher.publishEvent(new PluginsSynchronizedEvent(connectionId)));
```
nquinquenel, line 164: I think same question I asked before but just to double check - this `PluginsSynchronizedEvent` event might be skipped if all analyzers are already downloaded. It's ok?

backend/core/src/main/java/org/sonarsource/sonarlint/core/plugin/PluginsService.java
```diff
@@ -155,70 +159,30 @@ public LoadedPlugins getPlugins(String connectionId) {
   }
 
   private PluginsLoadResult loadPlugins(@Nullable String connectionId) {
-    var pluginsToLoadByKey = new HashMap<String, Path>();
-    var enabledLanguages = EnumSet.noneOf(SonarLanguage.class);
+    var strategy = getPluginLoadingStrategy(connectionId);
+    var result = strategy.resolveArtifacts();
```
nquinquenel, line 163: I was checking `resolveArtifacts()` and it can also be called from `getPluginStatuses` that is triggered on client side. Could it be an issue, maybe regarding duplications? I think the `UniqueTaskExecutor` prevents this but just to be sure

medium-tests/src/test/java/mediumtest/analysis/AnalysisMediumTests.java
```diff
@@ -751,31 +752,6 @@ void should_pass_eslint_bridge_path_to_sensors_if_provided(SonarLintTestHarness
       .untilAsserted(() -> assertThat(baseDir.resolve("property.dump")).hasContent(baseDir.resolve("eslint-bridge").toString()));
   }
 
-  @SonarLintTest
-  void should_not_pass_eslint_bridge_path_to_sensors_if_not_provided(SonarLintTestHarness harness, @TempDir Path baseDir) {
```
nquinquenel, line 755: What's the reason for removing this one?

doc/analyzer_management.md
```diff
@@ -1,78 +1,225 @@
...
+- Concurrent duplicate downloads for the same artifact are deduplicated (`UniqueTaskExecutor`).
+
+#### `ServerPluginSource`
+
+Backed by a specific SonarQube Server or SonarQube Cloud connection. One instance per connection,
+cached by `ConnectedArtifactsLoadingStrategyFactory`.
+
+- `listAvailableArtifacts(Set<SonarLanguage> enabledLanguages)` returns what the server currently exposes (converted from the server
+  plugin list). Falls back to an empty list if the server is unreachable.
+- `listServerPlugins()` returns the raw `ServerPlugin` list (richer metadata, used by the loading
```
nquinquenel, line 92: I think the method doesn't exist?

nquinquenel (approved):

backend/core/src/main/java/org/sonarsource/sonarlint/core/analysis/AnalysisService.java
damien-urruty-sonarsource: Yes, and yes. The `analysisReadinessByConfigScopeId` map is changed when the config scope becomes ready, and one of the condition for connected mode is that the plugins storage file exists. So the idea is that analysis might start with an incomplete set of plugins, and indeed if there was any download and `whenAllArtifactsDownloaded` is called, the plugins will be reloaded for the next analysis.

backend/core/src/main/java/org/sonarsource/sonarlint/core/plugin/PluginsService.java
damien-urruty-sonarsource: Correct. We wouldn't send it. The plugins will be loaded lazily from disk when needed

backend/core/src/main/java/org/sonarsource/sonarlint/core/plugin/PluginsService.java
damien-urruty-sonarsource: This is one thing I kept from the original implementation, but I'm not a big fan. It means that we might re-trigger downloads when the method is called, e.g. when the users visit the UI. I think we should instead keep a state somewhere after the resolution is done, so we don't have to re-resolve plugins every time. This is a future improvement to make, I might have some time to clean such things up after I'm done testing everything. In the meantime, apart from re-computing and potentially re-downloading plugins, I don't see any big problems

medium-tests/src/test/java/mediumtest/analysis/AnalysisMediumTests.java
damien-urruty-sonarsource: Several reasons: it was flaky, the assertion is misleading, and the test does not have much value (proving the absence of something). 

The assertion looks like we wait during 3s to validate that nothing is written to disk. But in fact most of the time the assertion exits successfully right away, before the sensor has a chance to write to disk. Except when the analysis goes fast enough, and something was written to disk before awaitility does its first check.

I could put it back and fix the assertion, WDYT?

doc/analyzer_management.md
damien-urruty-sonarsource: Ouch there was quite a few outdated data there. Fixed it

damien-urruty-sonarsource: Thanks for catching that!

medium-tests/src/test/java/mediumtest/analysis/AnalysisMediumTests.java
nquinquenel: I think it's fine given how the test was made

nquinquenel (approved):

backend/core/src/main/java/org/sonarsource/sonarlint/core/plugin/PluginsService.java
```diff
@@ -155,70 +159,30 @@ public LoadedPlugins getPlugins(String connectionId) {
...
-    for (var companionResolver : companionPluginResolvers) {
-      if (companionResolver == null) continue;
-      for (var status : companionResolver.resolveCompanionPlugins(connectionId).values()) {
-        if (status.state() == ArtifactState.DOWNLOADING) {
-          return true;
-        }
+    var pluginsToLoadByKey = new HashMap<String, Path>();
+    for (var entry : result.resolvedArtifactsByKey().entrySet()) {
+      var key = entry.getKey();
+      var artifact = entry.getValue();
```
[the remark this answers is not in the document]
damien-urruty-sonarsource: I will address this in a follow-up PR

PR #1999, opened 2026-04-30
SLCORE-2304 MISRA for cloud

Krosovok:


backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/LocalStorageSynchronizer.java
```diff
@@ -95,11 +96,14 @@ private AnalyzerConfiguration downloadAnalyzerConfig(ServerApi serverApi, String
     }
     var shouldForceRuleSetUpdate = outdatedSchema(currentSchemaVersion);
     var currentRuleSetsFinal = currentRuleSets;
-    var settings = new Settings(serverApi.settings().getProjectSettings(projectKey, cancelMonitor));
+    var projectSettings = new HashMap<>(serverApi.settings().getProjectSettings(projectKey, cancelMonitor));
+    if (serverApi.isSonarCloud()) {
+      projectSettings.put(ServerSettings.MISRA_COMPLIANCE_ENABLED, "true");
+    }
```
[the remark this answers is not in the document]
Krosovok: Too few lines + creates a hiding mutator method. Would make sense if at least a couple flags or other logic is common.

damien-urruty-sonarsource: Do we need this at all? I think propagating the global setting should be enough

damien-urruty-sonarsource (approved):
LGTM, one idea to simplify things

backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/LocalStorageSynchronizer.java
Krosovok: Hm, you seem to be right. I will remove it.

PR #2007, opened 2026-05-04
SLCORE-2311 Fix infinite plugin reload loop on failed on-demand artifact download

kirill-knize-sonarsource:


medium-tests/src/test/java/mediumtest/plugin/OnDemandAnalyzersMediumTests.java
```diff
@@ -87,6 +87,26 @@ void should_download_and_cache_cfamily_in_standalone_mode(SonarLintTestHarness h
     });
   }
 
+  @SonarLintTest
+  void failed_cfamily_download_should_not_trigger_infinite_reload_loop(SonarLintTestHarness harness) {
+    for (int i = 0; i < 20; i++) {
+      mockWebServer.enqueue(new mockwebserver3.MockResponse.Builder().code(500).build());
```
nquinquenel: static import?

backend/core/src/main/java/org/sonarsource/sonarlint/core/plugin/source/binaries/BinariesArtifactSource.java
```diff
@@ -179,12 +180,20 @@ private void downloadAndFireEvent(BinariesArtifact artifact) {
...
-    } catch (Exception e) {
-      LOG.error("Failed to download artifact with key {}", artifact.artifactKey(), e);
-      eventPublisher.publishEvent(new PluginStatusUpdateEvent(null, createdFailedStatuses(artifact)));
+    } catch (RuntimeException e) {
+      logAndPublishFailure(artifact, e);
+      throw e;
+    } catch (IOException e) {
+      logAndPublishFailure(artifact, e);
+      throw new UncheckedIOException(e);
     }
```
nquinquenel, line 194: It's not clear what's the difference here when you read this code.

nquinquenel (approved):
I guess it's ok, but it's not super clear to me when reading the code what the fix is about.
cc @damien-urruty-sonarsource just in case, as he worked on this

backend/core/src/main/java/org/sonarsource/sonarlint/core/plugin/loading/strategy/ArtifactsLoadingResult.java
```diff
@@ -84,9 +84,11 @@ public void whenAllArtifactsDownloaded(Runnable runnable) {
...
       .ifPresent(future -> {
         var logOutput = SonarLintLogger.get().getTargetForCopy();
-        future.thenRun(() -> {
-          SonarLintLogger.get().setTarget(logOutput);
-          runnable.run();
+        future.whenComplete((ignored, error) -> {
+          if (error == null) {
+            SonarLintLogger.get().setTarget(logOutput);
+            runnable.run();
+          }
```
damien-urruty-sonarsource: I think the new code does the same thing as the previous code, but in a more verbose way. Is your assumption that `thenRun` executes the callback even in case the future completes exceptionally? Because I don't think it's the case, it should execute only when the future completes normally

backend/core/src/main/java/org/sonarsource/sonarlint/core/plugin/source/binaries/BinariesArtifactSource.java
```diff
@@ -179,12 +180,20 @@ private void downloadAndFireEvent(BinariesArtifact artifact) {
     try {
       var path = downloadAndCache(artifact);
       eventPublisher.publishEvent(new PluginStatusUpdateEvent(null, createSuccessStatuses(artifact, path)));
-    } catch (Exception e) {
-      LOG.error("Failed to download artifact with key {}", artifact.artifactKey(), e);
-      eventPublisher.publishEvent(new PluginStatusUpdateEvent(null, createdFailedStatuses(artifact)));
+    } catch (RuntimeException e) {
+      logAndPublishFailure(artifact, e);
+      throw e;
```
damien-urruty-sonarsource: If we throw from these catch blocks, it means the plugin loading will be stopped. So a network problem when trying to download one plugin will cause no plugins to be loaded, I don't think this is what we want

kirill-knize-sonarsource: Addressed

backend/core/src/main/java/org/sonarsource/sonarlint/core/plugin/loading/strategy/ArtifactsLoadingResult.java
kirill-knize-sonarsource: Fixed

backend/core/src/main/java/org/sonarsource/sonarlint/core/plugin/source/binaries/BinariesArtifactSource.java
kirill-knize-sonarsource: Refactored and added comment

medium-tests/src/test/java/mediumtest/plugin/OnDemandAnalyzersMediumTests.java
kirill-knize-sonarsource: Yep

damien-urruty-sonarsource (approved):
LGTM, but I think the fix is a bit complex, maybe because of a bad design. The sources should maybe not send events directly, and let callers check the future and act in consequence. I suggest we work on a follow-up ticket to clean this further

PR #2009, opened 2026-05-06
SLCORE-2316 Fix failing IT consecutive to rule change

damien-urruty-sonarsource:


Krosovok (commented):
Why not just use another Hotspot? S5332 for example. And change the sample instead:

```
apiVersion: batch/v1
kind: Job
metadata:
  name: curl
spec:
  template:
    spec:
      containers:
      - name: curl
        image: curlimages/curl
        command: ["curl"]
        args: ["http://example.com/"] # Sensitive
```

damien-urruty-sonarsource:
@Krosovok 2 reasons:
* the goal with those ITs is to check that we can get "something" out of analyzers, a hotspot or issue does not make a big difference.
* I think this hotspot -> issue move is not an isolated example but part of a bigger move to eventually remove hotspots. I would not swap a hotspot for another one in our tests, as we will probably have to change again in some time

Krosovok:
> @Krosovok 2 reasons:
> 
> * the goal with those ITs is to check that we can get "something" out of analyzers, a hotspot or issue does not make a big difference.
> * I think this hotspot -> issue move is not an isolated example but part of a bigger move to eventually remove hotspots. I would not swap a hotspot for another one in our tests, as we will probably have to change again in some time

Then just another issue? It works as a quick fix, but I am sure it's just better to have a consistent check across versions.

Krosovok (approved):

damien-urruty-sonarsource:
@Krosovok makes sense yes, thanks for pushing :) Changed to issue, if you want to have another look

Krosovok:
Thanks! Some checks are red. I think the rule itself is not raised on SQLts99 for some reason. Others seem to be 503s.

damien-urruty-sonarsource:
@Krosovok yes in fact the rule I picked was not yet there in 9.9. And in fact I have the impression only hotspots were present in 9.9. I will move on with issues only but execute the tests only on more recent versions than 9.9. Which I hope to remove in a near future

PR #2010, opened 2026-05-06
SLCORE-2317 Replace deprecated public runners (sonar-xs-public, sonar-m-public)

kirill-knize-sonarsource:
----
## Summary by Gitar

- **CI/CD Configuration:**
  - Replaced deprecated `sonar-xs-public` and `sonar-m-public` runners with `sonar-xs` and `sonar-m` in multiple GitHub workflow files.
  - Updated runners across build, test, release, and PR management workflows to ensure continued compatibility with infrastructure updates.

### DIFF
--- a/.github/workflows/PullRequestClosed.yml
+++ b/.github/workflows/PullRequestClosed.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   PullRequestClosed_job:
     name: Pull Request Closed
-    runs-on: sonar-xs-public
+    runs-on: sonar-xs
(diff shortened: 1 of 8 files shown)


damien-urruty-sonarsource (commented):
It's not only about `sonar-xs-public` -> `sonar-xs`, from my understanding all public runners are deprecated and should be replaced, e.g. we also have `sonar-m-public` and probably others

damien-urruty-sonarsource (approved):

PR #2071, opened 2026-05-13
SLCORE-2378 Make implicit default time zone explicit in now() calls

tomshafir-sonarsource:
Part of DEX-13


nquinquenel (approved):
Looks good!

PR #2088, opened 2026-05-22
SLCORE-2387 SLCORE-2394 SLCORE-2395 Trigger local SCA analysis

eray-felek-sonarsource:
----
## Summary by Gitar

- **SCA Analysis Infrastructure:**
  - Added `ScaProjectAnalysisService` to orchestrate on-demand manual SCA analysis.
  - Added `ScaAnalysisContextResolver` to resolve binding, scanner endpoints, and configuration for local analysis.
  - Implemented `LocalDependencyRiskService` (replacing `LocalDependencyRiskAnalysisCache`) to manage and merge local SCA scanner findings.
  - Added `DependencyRiskDtoMapper` to reconcile and enrich risk data with local analysis details.
- **RPC Protocol Updates:**
  - Added `analyzeProject` RPC method in `DependencyRiskRpcService`.
  - Updated `DependencyRiskDto` to use a builder pattern and include fields for local enrichment (`presence`, `localAnalysisDetails`).
- **Testing Infrastructure:**
  - Added `ScaProjectAnalysisServiceTests` and `DependencyRiskServiceTests`.
  - Added `AnalyzeDependencyRiskProjectMediumTests` to verify analysis triggers and error handling.
- **Build Configuration:**
  - Added `com.sonarsource.sca:sca-scanner` dependency to `backend/core/pom.xml`.


eray-felek-sonarsource:
> Code Review ⚠️ Changes requested 3 resolved / 5 findings
> Implements local SCA analysis trigger for project dependency risks, but requires handling potential NPEs from null JSON deserialization fields and collection copies.
> 
> ⚠️ Edge Case: NPE risk when params fields are null from JSON deserialization
> 📄 [backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java:125](https://github.com/SonarSource/sonarlint-core/pull/2088/files#diff-707be5eff43ea7dd19560bc1b87504452f0188166d1002aaf9444e3c7aa82800R125) 📄 [backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java:127](https://github.com/SonarSource/sonarlint-core/pull/2088/files#diff-707be5eff43ea7dd19560bc1b87504452f0188166d1002aaf9444e3c7aa82800R127) 📄 [rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/sca/AnalyzeDependencyRiskProjectParams.java:28-29](https://github.com/SonarSource/sonarlint-core/pull/2088/files#diff-8fb01d3c03ff5bf9fa03f7051a9bb379c801ae727378ec4731e90ad13d3c22cdR28-R29)
> 
> When `AnalyzeDependencyRiskProjectParams` is deserialized via JSON-RPC (e.g., Gson reflective deserialization), fields `excludedPaths` and `scannerProperties` may be `null` if omitted from the request. At lines 125 and 127, `List.copyOf(params.getExcludedPaths())` and `Map.copyOf(params.getScannerProperties())` will throw `NullPointerException` on null input.
> 
> Since this is a public RPC API, clients may not always provide these optional-looking fields.
> 
> Add null guards before calling List.copyOf / Map.copyOf
> ```
> .setExcludedPaths(params.getExcludedPaths() == null ? List.of() : List.copyOf(params.getExcludedPaths()))
> .setScmExclusionEnabled(params.getScmExclusionEnabled() == null || params.getScmExclusionEnabled())
> .setScannerProperties(params.getScannerProperties() == null ? Map.of() : Map.copyOf(params.getScannerProperties()))
> ```
> 
> ⚠️ Edge Case: NPE risk: List.copyOf/Map.copyOf on potentially null params fields
> 📄 [backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java:125](https://github.com/SonarSource/sonarlint-core/pull/2088/files#diff-707be5eff43ea7dd19560bc1b87504452f0188166d1002aaf9444e3c7aa82800R125) 📄 [backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java:127](https://github.com/SonarSource/sonarlint-core/pull/2088/files#diff-707be5eff43ea7dd19560bc1b87504452f0188166d1002aaf9444e3c7aa82800R127)
> 
> At lines 125 and 127, `List.copyOf(params.getExcludedPaths())` and `Map.copyOf(params.getScannerProperties())` will throw `NullPointerException` if the deserialized params object has null values for these fields. The LSP4J JSON-RPC layer uses Gson which can bypass constructors during deserialization, meaning `excludedPaths` and `scannerProperties` may be `null` even though the full constructor initializes them to `List.of()`/`Map.of()`. This is consistent with the previous finding on this file that remains active.
> 
> Add null guards before calling List.copyOf/Map.copyOf to handle cases where Gson deserialization leaves fields null.
> ```
> .setExcludedPaths(params.getExcludedPaths() == null ? List.of() : List.copyOf(params.getExcludedPaths()))
> .setScmExclusionEnabled(params.getScmExclusionEnabled() == null || params.getScmExclusionEnabled())
> .setScannerProperties(params.getScannerProperties() == null ? Map.of() : Map.copyOf(params.getScannerProperties()))
> ```
> 
> ✅ 3 resolved
> ✅ Bug: Temporary work directory is never cleaned up after analysis
> > 📄 backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java:108-122
> > The `createWorkDir()` method creates a new temporary directory under `sca-scanner/work/analyze-project-*` on every invocation of `analyzeProject`, but no cleanup is performed after the analysis completes (neither in a `finally` block nor via try-with-resources). Over time, repeated analyses will accumulate stale directories consuming disk space.
> > Consider wrapping the analysis in a try-finally block that deletes the work directory after use (or on failure).
> 
> ✅ Bug: Race condition: rerun request can be lost between threads
> > 📄 backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaAnalysisManager.java:96-101 📄 backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaAnalysisManager.java:109-118 📄 backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaAnalysisManager.java:258-265
> > In `ScaAnalysisManager`, there is a race window between `onAnalysisCompleted` and `handleAlreadyRunningAnalysis` that can cause an automatic rerun request to be silently lost.
> > Scenario:
> > 
> > 1. Thread A calls `analyzeProject` with AUTOMATIC trigger, gets `currentAnalysis` from `runningAnalysis.get()`
> > 2. Thread B (executor) enters `onAnalysisCompleted`, successfully does `compareAndSet(analysis, null)`, then calls `consumeRerunParams()` → returns null
> > 3. Thread A proceeds into `handleAlreadyRunningAnalysis`, calls `currentAnalysis.requestRerun(params)` — sets `rerunParams` on the now-stale analysis
> > 4. Thread B evaluates `rerunParams != null` → false (already consumed), skips the rerun
> > 5. The automatic rerun request is permanently lost
> > 
> > The root cause is that `runningAnalysis` is cleared (CAS to null) before `consumeRerunParams`, creating a window where new callers can still call `requestRerun` on the old analysis object via their stale reference.
> 
> ✅ Bug: cancelMonitor from RPC framework is unused in analyzeProject
> > 📄 backend/rpc-impl/src/main/java/org/sonarsource/sonarlint/core/rpc/impl/DependencyRiskRpcServiceDelegate.java:55-56 📄 backend/rpc-impl/src/main/java/org/sonarsource/sonarlint/core/rpc/impl/DependencyRiskRpcServiceDelegate.java:51-53
> > In `DependencyRiskRpcServiceDelegate.analyzeProject`, the `cancelMonitor` provided by the RPC framework is ignored — it's not passed to `ScaAnalysisManager.analyzeProject(params)`. This means if the RPC client cancels the request (e.g., IDE closes the request), the `ScaAnalysisManager` won't know about it. The only way to cancel is via the explicit `cancelAnalysis` endpoint. While the analysis itself can be cancelled via `cancelAnalysis`, the blocking `result.join()` in `awaitResult()` will keep the RPC thread occupied until the analysis finishes or is explicitly cancelled through the separate endpoint.
> 
> 🤖 Prompt for agents
> ```
> Code Review: Implements local SCA analysis trigger for project dependency risks, but requires handling potential NPEs from null JSON deserialization fields and collection copies.
> 
> 1. ⚠️ Edge Case: NPE risk when params fields are null from JSON deserialization
>    Files: backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java:125, backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java:127, rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/sca/AnalyzeDependencyRiskProjectParams.java:28-29
> 
>    When `AnalyzeDependencyRiskProjectParams` is deserialized via JSON-RPC (e.g., Gson reflective deserialization), fields `excludedPaths` and `scannerProperties` may be `null` if omitted from the request. At lines 125 and 127, `List.copyOf(params.getExcludedPaths())` and `Map.copyOf(params.getScannerProperties())` will throw `NullPointerException` on null input.
>    
>    Since this is a public RPC API, clients may not always provide these optional-looking fields.
> 
>    Fix (Add null guards before calling List.copyOf / Map.copyOf):
>    .setExcludedPaths(params.getExcludedPaths() == null ? List.of() : List.copyOf(params.getExcludedPaths()))
>    .setScmExclusionEnabled(params.getScmExclusionEnabled() == null || params.getScmExclusionEnabled())
>    .setScannerProperties(params.getScannerProperties() == null ? Map.of() : Map.copyOf(params.getScannerProperties()))
> 
> 2. ⚠️ Edge Case: NPE risk: List.copyOf/Map.copyOf on potentially null params fields
>    Files: backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java:125, backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java:127
> 
>    At lines 125 and 127, `List.copyOf(params.getExcludedPaths())` and `Map.copyOf(params.getScannerProperties())` will throw `NullPointerException` if the deserialized params object has null values for these fields. The LSP4J JSON-RPC layer uses Gson which can bypass constructors during deserialization, meaning `excludedPaths` and `scannerProperties` may be `null` even though the full constructor initializes them to `List.of()`/`Map.of()`. This is consistent with the previous finding on this file that remains active.
> 
>    Fix (Add null guards before calling List.copyOf/Map.copyOf to handle cases where Gson deserialization leaves fields null.):
>    .setExcludedPaths(params.getExcludedPaths() == null ? List.of() : List.copyOf(params.getExcludedPaths()))
>    .setScmExclusionEnabled(params.getScmExclusionEnabled() == null || params.getScmExclusionEnabled())
>    .setScannerProperties(params.getScannerProperties() == null ? Map.of() : Map.copyOf(params.getScannerProperties()))
> ```
> 
> Tip
> 
> Comment `Gitar fix CI` or enable auto-apply: `gitar auto-apply:on`
> 
> Options
> Auto-apply is off → Gitar will not commit updates to this branch.Display: compact → Showing less information.Unblock → Override a blocking verdict and allow merging.
> 
> Comment with these commands to change:
> 
> Auto-apply	Compact	Unblock
> ```
> gitar auto-apply:on         
> ```
> 
> ```
> gitar display:verbose         
> ```
> 
> ```
> gitar unblock         
> ```
> 
> Was this helpful? React with 👍 / 👎 | [Gitar](https://gitar.ai)

The 2 remaining issues are addressed, I think this is a false positive

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java
```diff
@@ -0,0 +1,255 @@
...
+    this.storageService = storageService;
+    this.userPaths = userPaths;
+    this.scaScannerFactory = scaScannerFactory;
+  }
+
+  public AnalyzeDependencyRiskProjectResponse analyzeProject(AnalyzeDependencyRiskProjectParams params) {
+    return analyzeProject(params, new SonarLintCancelMonitor());
+  }
+
+  public AnalyzeDependencyRiskProjectResponse analyzeProject(AnalyzeDependencyRiskProjectParams params, SonarLintCancelMonitor cancelMonitor) {
```
georgii-borovinskikh-sonarsource, line 68: Long method - break down into logical steps and extract one method per step

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java
```diff
@@ -0,0 +1,255 @@
...
+      cancelMonitor.onCancel(() -> {
+        if (active.get()) {
+          analysisThread.interrupt();
+        }
+      });
+      var scanner = scaScannerFactory.create(ScaScannerOptionsBuilder.builder()
+        .setApiBaseUrl(apiBaseUrl)
+        .setDownloadBaseUrl(downloadBaseUrl)
+        .setSonarToken(sonarToken)
+        .setCacheDir(userPaths.getWorkDir().resolve("sca-scanner/cache"))
```
georgii-borovinskikh-sonarsource: this is where the scanner is going to store it's dependencies - should it be in the work dir? maybe a directory under the storage root instead?

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java
```diff
@@ -0,0 +1,255 @@
...
+    } finally {
+      active.set(false);
+      if (cancelMonitor.isCanceled()) {
+        Thread.interrupted();
+      }
+      deleteWorkDir(workDir);
+    }
+  }
+
+  private static AnalyzeDependencyRiskProjectResponse toDto(AnalyzeProjectResponse response) {
```
georgii-borovinskikh-sonarsource: I would move DTO conversion into a separate class and add it as a dependency

georgii-borovinskikh-sonarsource: Also, should we be converting to RPC dtos here? I'm not super familiar how SLCore is layered

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java
```diff
@@ -0,0 +1,255 @@
...
+    var credentials = sonarQubeClient.getCredentials();
+    if (!credentials.isLeft()) {
+      throw invalidArgument("SCA project analysis requires token credentials");
+    }
+    var sonarToken = credentials.getLeft().getToken();
+    var sqServerVersion = endpointParams.isSonarCloud() ? null : storageService.connection(binding.connectionId()).serverInfo().read()
+      .map(info -> info.version().getName())
+      .orElse(null);
+    Path workDir = null;
+    var active = new AtomicBoolean(true);
```
georgii-borovinskikh-sonarsource: Is this supposed to be used for synchronization? This is a local variable, so there may be an issue

backend/rpc-impl/src/main/java/org/sonarsource/sonarlint/core/rpc/impl/DependencyRiskRpcServiceDelegate.java
```diff
@@ -44,6 +47,11 @@ public CompletableFuture<ListAllDependencyRisksResponse> listAll(ListAllParams p
       .listAll(params.getConfigurationScopeId(), params.shouldRefresh(), cancelMonitor)));
   }
 
+  @Override
+  public CompletableFuture<AnalyzeDependencyRiskProjectResponse> analyzeProject(AnalyzeDependencyRiskProjectParams params) {
```
georgii-borovinskikh-sonarsource, line 51: What do you see as pros and cons of putting this here versus the analysis rpc service?

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/sca/AnalyzeDependencyRiskProjectParams.java
```diff
@@ -0,0 +1,69 @@
...
+  @Nullable
+  private final Map<String, String> scannerProperties;
+  private final Boolean scmExclusionEnabled;
+  private final Boolean debug;
+
+  public AnalyzeDependencyRiskProjectParams(String configurationScopeId) {
+    this(configurationScopeId, List.of(), Map.of(), null, null);
+  }
+
+  public AnalyzeDependencyRiskProjectParams(String configurationScopeId, @Nullable List<String> excludedPaths, @Nullable Map<String, String> scannerProperties,
```
georgii-borovinskikh-sonarsource: shouldn't SLCore manage these properties (exclusions, scanner props, scm) as opposed to expect it to be provided by the client?

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/sca/DependencyRiskRpcService.java
```diff
@@ -33,6 +33,9 @@ public interface DependencyRiskRpcService {
   @JsonRequest
   CompletableFuture<ListAllDependencyRisksResponse> listAll(ListAllParams params);
 
+  @JsonRequest
+  CompletableFuture<AnalyzeDependencyRiskProjectResponse> analyzeProject(AnalyzeDependencyRiskProjectParams params);
```
georgii-borovinskikh-sonarsource, line 44: the response here is problematic - the clients will have to handle 2 different models for dependency risks (the one that is returned by listAll and this one). SLCore should merge them and ideally reuse the existing change notification

georgii-borovinskikh-sonarsource (changes requested):

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/sca/AnalyzeDependencyRiskProjectResponse.java
```diff
@@ -0,0 +1,364 @@
...
+  public static class AnalyzeDependencyRiskProjectIssueDto {
+    @Nullable
+    private final String key;
+    private final String severity;
+    @Nullable
+    private final Boolean showIncreasedSeverityWarning;
+    private final String type;
+    private final String quality;
+    @Nullable
+    private final String status;
```
nquinquenel: We have existing enum for these, for dependency risks

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/sca/AnalyzeDependencyRiskProjectResponse.java
```diff
@@ -0,0 +1,364 @@
...
+        this.cvssScore = cvssScore;
+        this.spdxLicenseId = spdxLicenseId;
+        this.versionOptions = versionOptions;
+      }
+    }
+
+    @Nullable
+    public String getKey() {
+      return key;
+    }
```
nquinquenel: We always use `@CheckForNull` for getter in the codebase, `@Nullable` is for fields and parameters

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/sca/AnalyzeDependencyRiskProjectParams.java
```diff
@@ -0,0 +1,69 @@
...
+import javax.annotation.Nullable;
+
+public class AnalyzeDependencyRiskProjectParams {
+  private final String configurationScopeId;
+  @Nullable
+  private final List<String> excludedPaths;
+  @Nullable
+  private final Map<String, String> scannerProperties;
+  private final Boolean scmExclusionEnabled;
+  private final Boolean debug;
```
nquinquenel: These are nullable

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java
```diff
@@ -0,0 +1,255 @@
...
+      var response = scanner.analyzeProject(AnalyzeProjectOptionsBuilder.builder()
+        .setProjectKey(binding.sonarProjectKey())
+        .setBaseDir(baseDir)
+        .setWorkDir(workDir)
+        .setExcludedPaths(List.copyOf(params.getExcludedPaths()))
+        .setScmExclusionEnabled(params.getScmExclusionEnabled() == null || params.getScmExclusionEnabled())
+        .setScannerProperties(Map.copyOf(params.getScannerProperties()))
+        .setInsideSqc(false)
+        .setSqServerVersion(sqServerVersion)
+        .setDebug(Boolean.TRUE.equals(params.getDebug()))
```
nquinquenel: This kind of operation should be in the params class

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/sca/DependencyRiskRpcService.java
nquinquenel, line 44: Please add JavaDoc here to explain what it does

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/sca/DependencyRiskRpcService.java
```diff
@@ -33,6 +33,9 @@ public interface DependencyRiskRpcService {
...
+
     /**
```
nquinquenel: Wrong indentation

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java
```diff
@@ -0,0 +1,255 @@
...
+    this.sonarQubeClientManager = sonarQubeClientManager;
+    this.clientFileSystemService = clientFileSystemService;
+    this.storageService = storageService;
+    this.userPaths = userPaths;
+    this.scaScannerFactory = scaScannerFactory;
+  }
+
+  public AnalyzeDependencyRiskProjectResponse analyzeProject(AnalyzeDependencyRiskProjectParams params) {
+    return analyzeProject(params, new SonarLintCancelMonitor());
+  }
```
nquinquenel: This has no purpose in production code, let's not expose test-only API

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java
```diff
@@ -0,0 +1,255 @@
...
+      }
+      return toDto(response);
+    } catch (IOException e) {
+      if (cancelMonitor.isCanceled()) {
+        Thread.interrupted();
+        throw new CancellationException();
+      }
+      throw invalidArgument("SCA project analysis failed: " + e.getMessage());
+    } catch (IllegalArgumentException | IllegalStateException e) {
+      throw invalidArgument(e.getMessage());
```
nquinquenel: Should not be invalid argument argument for IOException and IllegalStateException

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java
```diff
@@ -0,0 +1,255 @@
...
+    } finally {
+      active.set(false);
+      if (cancelMonitor.isCanceled()) {
+        Thread.interrupted();
+      }
```
nquinquenel: That's an usual way of dealing with interruption, we usually rely on cancelMonitor check method

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java
```diff
@@ -0,0 +1,255 @@
...
+  @CheckForNull
+  private static String firstNonBlank(@Nullable String first, @Nullable String second) {
+    if (first != null && !first.isBlank()) {
+      return first;
+    }
+    return second != null && !second.isBlank() ? second : null;
+  }
+
+  private static ResponseErrorException invalidArgument(String message) {
+    return new ResponseErrorException(new ResponseError(SonarLintRpcErrorCode.INVALID_ARGUMENT, Strings.CS.removeEnd(message, "."), null));
```
nquinquenel: Why do you strip the "."?

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java
```diff
@@ -0,0 +1,255 @@
...
+
+  private static String toScaDownloadBaseUrl(EndpointParams endpointParams) {
+    if (endpointParams.isSonarCloud()) {
+      return SONARCLOUD_SCA_SCANNER_CDN_URL;
+    }
+    return ServerApiHelper.concat(endpointParams.getBaseUrl(), "/api/v2/sca/clis");
+  }
+
+  @CheckForNull
+  private static String firstNonBlank(@Nullable String first, @Nullable String second) {
```
nquinquenel: Probably Apache Commons has a method for this

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java
```diff
@@ -0,0 +1,255 @@
...
+      versionOption.fixLevel(),
+      versionOption.descriptionCode());
+  }
+
+  private static AnalyzeDependencyRiskProjectErrorDto toDto(AnalysisErrorResource error) {
+    return new AnalyzeDependencyRiskProjectErrorDto(error.id(), error.code().name(), error.path(), error.message());
+  }
+
+  private Path createWorkDir() throws IOException {
+    var scaWorkDir = userPaths.getWorkDir().resolve("sca-scanner/work");
```
nquinquenel: I'd declare this magic value once, I see it's used in multiple places

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java
```diff
@@ -0,0 +1,255 @@
...
+    }
+    var downloadBaseUrl = firstNonBlank(System.getProperty(SCA_DOWNLOAD_BASE_URL_PROPERTY), toScaDownloadBaseUrl(endpointParams));
+    if (downloadBaseUrl == null) {
+      throw invalidArgument("Missing SCA scanner download base URL. Provide it in the request or with system property '" + SCA_DOWNLOAD_BASE_URL_PROPERTY + "'");
+    }
+    var sonarQubeClient = sonarQubeClientManager.getValidClientOrThrow(binding.connectionId());
+    var credentials = sonarQubeClient.getCredentials();
+    if (!credentials.isLeft()) {
+      throw invalidArgument("SCA project analysis requires token credentials");
+    }
```
nquinquenel: I think we usually use .map

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java
```diff
@@ -0,0 +1,255 @@
...
+    var sonarToken = credentials.getLeft().getToken();
+    var sqServerVersion = endpointParams.isSonarCloud() ? null : storageService.connection(binding.connectionId()).serverInfo().read()
+      .map(info -> info.version().getName())
+      .orElse(null);
+    Path workDir = null;
+    var active = new AtomicBoolean(true);
+    try {
+      workDir = createWorkDir();
+      LOG.info("Running manual SCA project analysis for configuration scope '{}'", configurationScopeId);
```
nquinquenel: I don't think we should mention "SCA" in the info log

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaScannerFactory.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class ScaScannerFactory {
+
+  public Scanner create(ScaScannerOptions options) {
+    var scanner = new ScaScanner(options);
+    return scanner::analyzeProject;
+  }
+
+  @FunctionalInterface
+  public interface Scanner {
```
nquinquenel: Can we find a better name? Scanner is very similar to existing Scanner of the SonarQube ecosystem, what about ScaProjectScanner

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java
```diff
@@ -0,0 +1,255 @@
...
+import org.sonarsource.sonarlint.core.rpc.protocol.backend.sca.AnalyzeDependencyRiskProjectResponse.AnalyzeDependencyRiskProjectErrorDto;
+import org.sonarsource.sonarlint.core.rpc.protocol.backend.sca.AnalyzeDependencyRiskProjectResponse.AnalyzeDependencyRiskProjectIssueDto;
+import org.sonarsource.sonarlint.core.rpc.protocol.backend.sca.AnalyzeDependencyRiskProjectResponse.AnalyzeDependencyRiskProjectReleaseDto;
+import org.sonarsource.sonarlint.core.rpc.protocol.backend.sca.AnalyzeDependencyRiskProjectResponse.AnalyzeDependencyRiskProjectVersionOptionDto;
+import org.sonarsource.sonarlint.core.serverapi.EndpointParams;
+import org.sonarsource.sonarlint.core.serverapi.ServerApiHelper;
+import org.sonarsource.sonarlint.core.serverconnection.FileUtils;
+import org.sonarsource.sonarlint.core.storage.StorageService;
+
+public class ScaProjectAnalysisService {
```
nquinquenel, line 48: Not easy to check from the PR, but do all the folders contain a package-info file?

backend/core/src/test/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisServiceTests.java
```diff
@@ -0,0 +1,294 @@
...
+    assertThat(scaScannerFactory.analyzeProjectOptions.workDir()).hasParentRaw(workDir.resolve("sca-scanner/work"));
+    assertThat(scaScannerFactory.analyzeProjectOptions.excludedPaths()).containsExactly("excluded/path");
+    assertThat(scaScannerFactory.analyzeProjectOptions.scmExclusionEnabled()).isFalse();
+    assertThat(scaScannerFactory.analyzeProjectOptions.scannerProperties()).containsEntry("sonar.verbose", "true");
+    assertThat(scaScannerFactory.analyzeProjectOptions.insideSqc()).isFalse();
+    assertThat(scaScannerFactory.analyzeProjectOptions.sqServerVersion()).isEqualTo("2025.4");
+    assertThat(scaScannerFactory.analyzeProjectOptions.debug()).isTrue();
+  }
+
+  private static void assertAnalysisResponse(org.sonarsource.sonarlint.core.rpc.protocol.backend.sca.AnalyzeDependencyRiskProjectResponse response) {
```
nquinquenel: Static import?

medium-tests/src/test/java/mediumtest/sca/AnalyzeDependencyRiskProjectMediumTests.java
```diff
@@ -0,0 +1,111 @@
...
+import org.eclipse.lsp4j.jsonrpc.ResponseErrorException;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.io.TempDir;
+import org.sonarsource.sonarlint.core.rpc.protocol.SonarLintRpcErrorCode;
+import org.sonarsource.sonarlint.core.rpc.protocol.backend.sca.AnalyzeDependencyRiskProjectParams;
+import org.sonarsource.sonarlint.core.rpc.protocol.common.ClientFileDto;
+import org.sonarsource.sonarlint.core.test.utils.SonarLintTestRpcServer;
+import org.sonarsource.sonarlint.core.test.utils.junit5.SonarLintTest;
+import org.sonarsource.sonarlint.core.test.utils.junit5.SonarLintTestHarness;
+import org.sonarsource.sonarlint.core.serverapi.features.Feature;
```
nquinquenel: Wrong import order

medium-tests/src/test/java/mediumtest/sca/AnalyzeDependencyRiskProjectMediumTests.java
```diff
@@ -0,0 +1,111 @@
...
+import uk.org.webcompere.systemstubs.jupiter.SystemStub;
+import uk.org.webcompere.systemstubs.jupiter.SystemStubsExtension;
+import uk.org.webcompere.systemstubs.properties.SystemProperties;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.catchThrowable;
+
+@ExtendWith(SystemStubsExtension.class)
+class AnalyzeDependencyRiskProjectMediumTests {
```
nquinquenel, line 48: I think the medium tests are very thin, they do not test much

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/sca/AnalyzeDependencyRiskProjectResponse.java
```diff
@@ -0,0 +1,364 @@
...
+      private final String licenseExpression;
+
+      public PackageDto(String packageUrl, String packageManager, String packageName, String version, @Nullable String licenseExpression) {
+        this.packageUrl = packageUrl;
+        this.packageManager = packageManager;
+        this.packageName = packageName;
+        this.version = version;
+        this.licenseExpression = licenseExpression;
+      }
+    }
```
nquinquenel: We declare a bunch of DTO with private fields and no getter, I'm not sure this is correct

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/sca/DependencyRiskRpcService.java
eray-felek-sonarsource: `AnalyzeDependencyRiskProjectResponse` has a lot richer information and some of them we can't remove for merging such as: 

In our MMF we are expected to provide quick fix. In order to do that we would need `depdendencyChains` and `dependencyFilePaths`. How do you suggest we merge the responses while satisfying the MMF requirement?

eray-felek-sonarsource: Just to confirm, what we want to do in the end is when client calls `analyzeProject`, we want to internally also call `listAll` and merge the server ones with local ones using a **flag** like `isMatched`, `isLocalOnly` and return the result. Server issues will have less information but **DTO** we return will be able to have all the information making the extra fields like **chains** nullable

medium-tests/src/test/java/mediumtest/sca/AnalyzeDependencyRiskProjectMediumTests.java
eray-felek-sonarsource: Automatic order I have gives me this. I think this topic was discussed multiple times with you and also Julien. What do you think right order should be ?

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java
eray-felek-sonarsource: There is another JIRA [ticket](https://sonarsource.atlassian.net/jira/software/c/projects/SLCORE/boards/1527?quickFilter=4014&selectedIssue=SLCORE-2400) for cancellation. We can assume cancellation is not implemented in this PR.

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/sca/DependencyRiskRpcService.java
eray-felek-sonarsource: Not sure if I understand what needs to change here. Applying formatting to my recent changes doesn't change anything. Can you check if this is still the issue for **PR Review** commit?

backend/rpc-impl/src/main/java/org/sonarsource/sonarlint/core/rpc/impl/DependencyRiskRpcServiceDelegate.java
eray-felek-sonarsource: During the cancellation implementation we will introduce a tracking different than analysis tracking and I think this analysis is fundamentally different then regular analysis so I think we should keep it here. I don't see why we would want to have it inside analysis rpc service.

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java
eray-felek-sonarsource: This logic belongs to cancellation which has it's own JIRA ticket and not part of this PR.

Removed this part

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java
eray-felek-sonarsource: Since we are doing merging here instead, do you think this is still relevant?

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java
eray-felek-sonarsource: Yes, all the non-test classes do

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java
georgii-borovinskikh-sonarsource: Not sure, but probably more relevant (same code would be reused). However, I would need to see how it looks before I can tell. Is it ready yet?

backend/rpc-impl/src/main/java/org/sonarsource/sonarlint/core/rpc/impl/DependencyRiskRpcServiceDelegate.java
georgii-borovinskikh-sonarsource: >I don't see why we would want to have it inside analysis rpc service.

If we integrate this into the full project analysis, for example

medium-tests/src/test/java/mediumtest/sca/AnalyzeDependencyRiskProjectMediumTests.java
eray-felek-sonarsource: This was also an issue on CLI implementation I think. 

We would need real downloading of their **CLI** to add more meaningful tests which should be covered by [this](https://sonarsource.atlassian.net/browse/SLCORE-2402) ticket.

WDYT? Should we still add more tests in the scope of this PR? If so please specify which paths you'd like to see covered

backend/rpc-impl/src/main/java/org/sonarsource/sonarlint/core/rpc/impl/DependencyRiskRpcServiceDelegate.java
eray-felek-sonarsource: We do not plan to do such a thing

georgii-borovinskikh-sonarsource: Not in this iteration, but why not in the future?

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java
georgii-borovinskikh-sonarsource: You've already added `DependencyRiskDtoMapper` 👍

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java
```diff
@@ -0,0 +1,320 @@
...
+    for (var release : localReleases) {
+      for (var issue : release.issues()) {
+        var serverRisk = lookupServerRisk(serverRiskById, issue);
+        if (serverRisk != null) {
+          merged.add(dependencyRiskDtoMapper.enrichServerDto(serverRisk, release, issue));
+          matchedServerIds.add(serverRisk.getId());
+        } else {
+          dependencyRiskDtoMapper.toLocalOnlyDto(release, issue).ifPresent(merged::add);
+        }
+      }
```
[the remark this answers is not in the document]
eray-felek-sonarsource: It is not possible

backend/rpc-impl/src/main/java/org/sonarsource/sonarlint/core/rpc/impl/DependencyRiskRpcServiceDelegate.java
georgii-borovinskikh-sonarsource: Not a blocker, worst case it will be slightly inconvenient to call 2 different interfaces from the client - let's move on

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/sca/DependencyRiskRpcService.java
nquinquenel: The `    /**` has wrong indentation, you could take the time to fix it (it's not from your changes, but next to it)

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java
nquinquenel: I'd put a comment referring to this ticket

medium-tests/src/test/java/mediumtest/sca/AnalyzeDependencyRiskProjectMediumTests.java
nquinquenel: I don't think automatic order gave you this, here it's just alphabetical order from java packages. The topic that was discussed in the past is about kotlin/java packages order, and java packages when it's from java API modules or external packages, which is not the case here

medium-tests/src/test/java/mediumtest/sca/AnalyzeDependencyRiskProjectMediumTests.java
nquinquenel: Real downloading would likely be an integration test, here it's a medium test. We could mock interactions with CLI via the harness, similarly to what we do with `withSonarQubeConnection`

backend/rpc-impl/src/main/java/org/sonarsource/sonarlint/core/rpc/impl/DependencyRiskRpcServiceDelegate.java
eray-felek-sonarsource: As discussed

medium-tests/src/test/java/mediumtest/sca/AnalyzeDependencyRiskProjectMediumTests.java
eray-felek-sonarsource: Just to clarify the technical constraint here: the CLI interaction doesn't actually happen in our code. It is entirely encapsulated within the sca-scanner library. The library handles downloading the CLI, spawning it as a separate process, and reading its output. Because only the initial download is an HTTP call, the test harness doesn't have an interception point for the actual execution.

This is why we can't really compare this to `withSonarQubeConnection`, which works by returning canned responses to HTTP calls made directly by our backend. Stubbing the download step wouldn't solve this either—whatever we serve still gets executed as a real program. Serving anything other than a working executable just reproduces the failure case we already have covered.

To actually test the happy path in a medium test, we would have to choose between two approaches:

Option 1: **Provide a fake CLI executable**
We could provide a small fake CLI program that prints canned output and let it run normally (via the existing cliOverrideLocation option).

Trade-off: Because it’s a real executable, it is OS-specific (we'd need to handle Windows vs. Linux). This makes it feel much more like an integration test.

Option 2: **Inject a fake scanner**
We could change our backend dependency injection so tests can replace the real scanner with a fake one that simply returns canned results.

Trade-off: While this avoids the OS-specific execution problem, it requires adding test-only code to the production backend and entirely skips the download, run, and parse steps.

eray-felek-sonarsource: Do you think we should go with one of these options or should we continue with my initial suggestion?

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/tracking/DependencyRiskDto.java
```diff
@@ -37,21 +38,146 @@ public class DependencyRiskDto {
   @Nullable
   private final String cvssScore;
   private final List<Transition> transitions;
+  @Nullable
+  private final LocalAnalysisDetailsDto localAnalysisDetails;
+  private final boolean matched;
```
georgii-borovinskikh-sonarsource: 'matched' is a bit ambiguous - could we just have two booleans like `foundOnServer` and `foundLocally` and then derive the 3 different states from there: local only (false true), server only (true false) & both (true true)

An enum would also work here

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/sca/AnalyzeDependencyRiskProjectResponse.java
```diff
@@ -0,0 +1,96 @@
...
+ * flags and an optional {@link DependencyRiskDto.LocalAnalysisDetailsDto} block with local-analysis enrichment
+ * (dependency chains, dependency file paths, upgrade {@code versionOptions}, CWE ids, etc.).
+ * </p>
+ * <p>
+ * {@link #getParsedFiles()} and {@link #getErrors()} expose local-analysis diagnostics that have no equivalent in the
+ * server-tracked model.
+ * </p>
+ */
+public class AnalyzeDependencyRiskProjectResponse {
+  private final List<DependencyRiskDto> dependencyRisks;
```
georgii-borovinskikh-sonarsource: The current pattern for SLCore based analysis triggering is that explicit analyze methods DON'T return issues. If in the future we implement automatic analysis, then we'll have to change this response to fit that pattern. Should we already do that?

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/sca/DependencyRiskRpcService.java
```diff
@@ -33,6 +33,13 @@ public interface DependencyRiskRpcService {
   @JsonRequest
   CompletableFuture<ListAllDependencyRisksResponse> listAll(ListAllParams params);
```
georgii-borovinskikh-sonarsource, line 34: local scan results are not integrated into this method. What is the end vision for how the frontends would use/combine these methods?

georgii-borovinskikh-sonarsource: Same goes for the client notification `didChangeDependencyRisks`

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/tracking/DependencyRiskDto.java
```diff
@@ -94,6 +220,187 @@ public List<Transition> getTransitions() {
...
+    private final IssueDetailsDto issueDetails;
+    private final DependencyDetailsDto dependencyDetails;
+
+    public LocalAnalysisDetailsDto(ReleaseDetailsDto releaseDetails, IssueDetailsDto issueDetails, DependencyDetailsDto dependencyDetails) {
+      this.releaseDetails = releaseDetails;
+      this.issueDetails = issueDetails;
+      this.dependencyDetails = dependencyDetails;
+    }
+
+    public ReleaseDetailsDto getReleaseDetails() {
```
georgii-borovinskikh-sonarsource, line 231: What will this be used for on the frontend?

medium-tests/src/test/java/mediumtest/sca/AnalyzeDependencyRiskProjectMediumTests.java
eray-felek-sonarsource: Confirmed it's fixed

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/tracking/DependencyRiskDto.java
eray-felek-sonarsource: Currently we are not cutting any information that we return to frontend. Do you think this information would never be used? We can let the frontend decide if it's useful or not.

eray-felek-sonarsource: Not sure if we should investigate all the information we return to frontends are useful or not at this stage

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/sca/DependencyRiskRpcService.java
eray-felek-sonarsource: As discussed

> We first call analyze project locally
> Keep things in cache
> After analysis is finished we stream with `didChangeDependencyRisks` following the general patterns

eray-felek-sonarsource: @georgii-borovinskikh-sonarsource  following up on our discussion here, since the local analysis can yield different results depending on the branch we are on, I planning to store the analysis results in cache based on configurationScopeId + branchId, WDYT?

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/sca/AnalyzeDependencyRiskProjectResponse.java
eray-felek-sonarsource: Same as discussed below

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/tracking/DependencyRiskDto.java
eray-felek-sonarsource: We decided we can do this if we have extra time but not essential

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java
eray-felek-sonarsource: removed

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/tracking/DependencyRiskDto.java
nquinquenel: Does that mean we don't know what will be implemented on client side?

medium-tests/src/test/java/mediumtest/sca/AnalyzeDependencyRiskProjectMediumTests.java
nquinquenel: I think we could have a trade off solution with a very lightweight executable that fakes SCA CLI and creates a fake .tar.gz. It could be used with a system property to override the location of the CLI. This way, we can stub SCA analysis APIs and have a medium tests that can simulate the whole end to end chain.

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/DependencyRiskService.java
```diff
@@ -84,19 +90,32 @@ public List<DependencyRiskDto> listAll(String configurationScopeId, boolean shou
       .orElseGet(Collections::emptyList);
   }
 
+  public List<DependencyRiskDto> updateLocalAnalysisAndNotify(String configurationScopeId, AnalyzeProjectResponse localAnalysis, SonarLintCancelMonitor cancelMonitor) {
+    localDependencyRiskAnalysisCache.put(configurationScopeId, localAnalysis);
+    var newRisks = listAll(configurationScopeId, false, cancelMonitor);
+    var delta = dependencyRiskMerger.computeDelta(newRisks);
+    if (!delta.addedRisks().isEmpty() || !delta.updatedRisks().isEmpty()) {
+      client.didChangeDependencyRisks(new DidChangeDependencyRisksParams(configurationScopeId, Set.of(), delta.addedRisks(), delta.updatedRisks()));
+    }
```
nquinquenel: I'm not sure this work as expected, local analysis should compare previous merged snapshot vs new merged snapshot (by id where present, and a stable local key otherwise), not filter by Presence on the new list only

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java
```diff
@@ -0,0 +1,149 @@
...
+
+  private static ResponseErrorException requestFailed(String message) {
+    return new ResponseErrorException(new ResponseError(ResponseErrorCode.RequestFailed, message, null));
+  }
+}
+
+
+
+
+
```
nquinquenel: Empty lines to remove

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaProjectAnalysisService.java
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  public AnalyzeDependencyRiskProjectResponse analyzeProject(AnalyzeDependencyRiskProjectParams params, SonarLintCancelMonitor cancelMonitor) {
+    var configurationScopeId = params.getConfigurationScopeId();
+    var context = contextResolver.resolve(configurationScopeId);
+    var localResponse = runLocalAnalysis(configurationScopeId, context);
+    var mergedRisks = dependencyRiskService.updateLocalAnalysisAndNotify(configurationScopeId, localResponse, cancelMonitor);
+    var errors = localResponse.errors().stream().map(dependencyRiskDtoMapper::toErrorDto).toList();
+    return new AnalyzeDependencyRiskProjectResponse(mergedRisks, localResponse.parsedFiles(), errors);
+  }
```
nquinquenel, line 76: I think we may want to check if SCA is supported there

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/ScaAnalysisContextResolver.java
```diff
@@ -0,0 +1,162 @@
...
+  private static ScannerEndpoint resolveScannerEndpoint(AbstractConnectionConfiguration connection) {
+    var endpointParams = connection.getEndpointParams();
+    var apiBaseUrl = toScaApiBaseUrl(endpointParams);
+    if (apiBaseUrl == null) {
+      throw invalidArgument("Missing SCA API base URL");
+    }
+    var downloadBaseUrl = StringUtils.firstNonBlank(System.getProperty(SCA_DOWNLOAD_BASE_URL_PROPERTY), toScaDownloadBaseUrl(endpointParams));
+    if (downloadBaseUrl == null) {
+      throw invalidArgument("Missing SCA scanner download base URL. Provide it in the request or with system property '" + SCA_DOWNLOAD_BASE_URL_PROPERTY + "'");
+    }
```
nquinquenel, line 99: I think the message is misleading, you can't provide the download URL via the request

nquinquenel (commented):
You'll need to update API_CHANGES as well with all the new APIs

medium-tests/src/test/java/mediumtest/sca/AnalyzeDependencyRiskProjectMediumTests.java
eray-felek-sonarsource: Given the work left for this iteration, I created a separate ticket for [this](https://sonarsource.atlassian.net/jira/software/c/projects/SLCORE/boards/1527?quickFilter=4014&selectedIssue=SLCORE-2429). Let's prioritize it for a near future iteration

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/DependencyRiskService.java
eray-felek-sonarsource: `updateLocalAnalysisAndNotify` is called from `analyzeProject` which is the local analysis trigger. The purpose here is to let the **Client** update the view based on the local analysis results. After the local analysis, **Client** needs to know only the newly added items that are coming from local analysis and matched ones which are part of the **updatedItems**. We do not need to do any comparison for this update. There is no stable local key and no need for this comparison since we only interested in representing local issues, matched issues and server issues.

Design is, 

- User has server only issues
- User triggers local analysis
- The view gets updated

It is not

- User has Server and Local issues already
- User triggers local analysis
- We compare and report server and local analysis changes

Given the current limitations we have for not having a stable local key. 

And, I don't think doing comparison for local issues would add much value IMO since we are not going to be able to changing their state or do some operation like anticipated issues within this iteration

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/DependencyRiskService.java
```diff
@@ -84,19 +91,40 @@ public List<DependencyRiskDto> listAll(String configurationScopeId, boolean shou
...
+    return newRisks;
+  }
```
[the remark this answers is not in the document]
eray-felek-sonarsource: localDependencyRiskAnalysisCache.put(configurationScopeId, localAnalysis); This does not accumulate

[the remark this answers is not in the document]
eray-felek-sonarsource: That is intentional

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/tracking/DependencyRiskDto.java
eray-felek-sonarsource: We have [this](https://sonarsource.atlassian.net/browse/SLI-2713) ticket on client side for investigation and implementing the quick-fix. Within the ticket we are planning to investigate what more value we can get using the extra information the local analysis provides.

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/DependencyRiskService.java
```diff
@@ -84,19 +91,40 @@ public List<DependencyRiskDto> listAll(String configurationScopeId, boolean shou
...
+        var addedRisks = summary.addedItems().stream()
           .map(DependencyRiskService::toDto)
-          .toList(),
-        summary.updatedItems().stream()
+          .toList();
+        var updatedRisks = summary.updatedItems().stream()
           .map(DependencyRiskService::toDto)
-          .toList())));
+          .toList();
+        client.didChangeDependencyRisks(new DidChangeDependencyRisksParams(configurationScopeId, summary.deletedItemIds(),
```
georgii-borovinskikh-sonarsource: deleted here should affect status of local risks to change to found only locally

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/tracking/DependencyRiskDto.java
eray-felek-sonarsource: And this [epic](https://sonarsource.atlassian.net/browse/SLI-2650) includes tickets of what we are planning to do on client side

rpc-protocol/src/main/java/org/sonarsource/sonarlint/core/rpc/protocol/backend/sca/AnalyzeDependencyRiskProjectResponse.java
georgii-borovinskikh-sonarsource: Did you forget to remove this one?

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/DependencyRiskService.java
```diff
@@ -84,19 +91,40 @@ public List<DependencyRiskDto> listAll(String configurationScopeId, boolean shou
       .orElseGet(Collections::emptyList);
   }
 
+  public List<DependencyRiskDto> updateLocalAnalysisAndNotify(String configurationScopeId, AnalyzeProjectResponse localAnalysis, SonarLintCancelMonitor cancelMonitor) {
+    localDependencyRiskAnalysisCache.put(configurationScopeId, localAnalysis);
+    var newRisks = listAll(configurationScopeId, false, cancelMonitor);
+    var delta = dependencyRiskMerger.computeDelta(newRisks);
```
georgii-borovinskikh-sonarsource: it's not clear from the code how we compute the delta, and why it takes only one parameter as input

georgii-borovinskikh-sonarsource: I think I got confused thinking that listAll doesn't take local issues into account. It would be better if a separate class was managing the merging of local and server risks (combined storage). That way it would be clear that those are the merged issues and this class would be smaller

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/DependencyRiskService.java
georgii-borovinskikh-sonarsource: I suspect that currently local issues would not be removed from the client's own storage, becasue the deleted set here is empty. If we don't have internal logic to maintain the diff, we should at least force remove all local-only issues and re-add all of them. Of course, this won't be the true minimal diff, but currently the diff is not correct

georgii-borovinskikh-sonarsource (changes requested):

eray-felek-sonarsource:
image

For this, we decided, for now we will go with the simple solution of returning newly added local issues also returning previous cache value of the local analysis snapshot as deleted(for which we will add an unstable id to make it easy for clients).

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/DependencyRiskService.java
eray-felek-sonarsource: deleted items has ids in it, so we have local issue with an id in this and we can not match it so we address this situation in the end we want to see this as updated and the ones in server only in deleted. Already covered below

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/DependencyRiskService.java
eray-felek-sonarsource: We encapsulate the dependency on them

- private final StorageService storageService; //maybe not full dependency but the logic for newly introduced code
- private final LocalDependencyRiskAnalysisCache localDependencyRiskAnalysisCache;
- private final DependencyRiskMerger dependencyRiskMerger;

Into one class that would do the all the logical operations they are doing.

So in the end we only have 1-2 dependency instead of 3 of them.

backend/core/src/main/java/org/sonarsource/sonarlint/core/sca/DependencyRiskService.java
```diff
@@ -84,19 +94,60 @@ public List<DependencyRiskDto> listAll(String configurationScopeId, boolean shou
...
+    var closedLocalOnlyRiskIds = previousRisks.stream()
+      .filter(risk -> risk.getPresence() == DependencyRiskDto.Presence.LOCAL_ONLY)
+      .map(DependencyRiskDto::getId)
+      .filter(Objects::nonNull)
+      .collect(Collectors.toSet());
+    if (!closedLocalOnlyRiskIds.isEmpty() || !delta.addedRisks().isEmpty() || !delta.updatedRisks().isEmpty()) {
+      client.didChangeDependencyRisks(new DidChangeDependencyRisksParams(configurationScopeId, closedLocalOnlyRiskIds, delta.addedRisks(), delta.updatedRisks()));
+    }
+    return newRisks;
```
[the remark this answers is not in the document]
eray-felek-sonarsource: Yes, this is a workaround we decided to do so the clients don't accumulate

PR #2106, opened 2026-06-02
SLCORE-2426 Warm up file-sources Lambda before SonarCloud ITs

vnaskos-sonar:
----
## Summary by Gitar

- **Tests:**
  - Increase `await` timeouts to 2 minutes for taint vulnerability and source availability checks to handle slow staging.
  - Add `warmUpFileSourcesLambda` to `SonarCloudTests` to preemptively trigger Lambda cold starts before analysis execution.


nquinquenel (approved):
Sounds good, just worried about leaking such implementation details in the comments - I think they wanted to update the infra at some point, which means this could become outdated

PR #2214, opened 2026-08-04
SLCORE-2537 Send machine_id and ide_installation_id in IDE telemetry

kirill-knize-sonarsource:
----
## Summary by Gitar

- **Telemetry identity tracking:**
    - Added `MachineIdProvider` and `GessieIdentity` to share `machine_id` and `installation_id` across IDE telemetry and CLI.
    - Updated `TelemetryHttpClient`, `GessieHttpClient`, and measures payloads to include identity fields.

### DIFF
--- a/backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/gessie/GessieHttpClient.java
+++ b/backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/gessie/GessieHttpClient.java
@@ -23,34 +23,70 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.gson.FieldNamingPolicy;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import java.util.Map.Entry;
 import java.util.concurrent.CompletableFuture;
 import org.sonarsource.sonarlint.core.commons.log.SonarLintLogger;
 import org.sonarsource.sonarlint.core.http.HttpClient;
 import org.sonarsource.sonarlint.core.http.HttpClientProvider;
 import org.sonarsource.sonarlint.core.telemetry.InternalDebug;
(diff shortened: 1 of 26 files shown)


backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/gessie/GessieHttpClient.java
```diff
@@ -23,36 +23,72 @@
...
     logGessiePayload(json);
     var futureResponse = client.postAsync(endpoint + "/ide", HttpClient.JSON_CONTENT_TYPE, json);
     handleGessieResponse(futureResponse);
   }
 
+  private JsonObject mergeIdentity(GessieEvent event) {
+    var eventJson = gson.toJsonTree(event).getAsJsonObject();
+    var eventPayloadElement = eventJson.get(EVENT_PAYLOAD_KEY);
+    if (eventPayloadElement != null && eventPayloadElement.isJsonObject()) {
+      var identity = new GessieIdentity(machineIdProvider.getMachineId(), telemetryLocalStorageManager.ideInstallationId());
```
vnaskos-sonar, line 73: Possibly minor: `machineIdProvider.getMachineId()` self-gates on `isTelemetryEnabledByUser()`, but `telemetryLocalStorageManager.ideInstallationId()` doesn't. It'll cerate and persist the id even if telemetry is off. Right now it's fine because `GessieService.onStartup()` checks `isTelemetryEnabledByUser()` before calling `postEvent()`, so we never actually hit that path with telemetry disabled.

vnaskos-sonar (approved):
👍

backend/telemetry/src/main/java/org/sonarsource/sonarlint/core/telemetry/gessie/GessieHttpClient.java
kirill-knize-sonarsource: I think disabling telemetry shall only disable sending the ping. It's OK.

PR #2221, opened 2026-08-06
SLCORE-2418 Migrate analyzer download from SQC Web endpoint to CloudFront CDN

damien-urruty-sonarsource:
----
## Summary by Gitar

- **Plugin download migration:**
    - Updated `PluginsApi` to download SonarQube Cloud plugins from CloudFront CDN using `scanner.` host prefix and plugin hash
    - Added `buildSonarCloudPluginDownloadUrl` method to construct CDN download URLs and updated corresponding tests and fixtures

### DIFF
--- a/backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/plugins/PluginsApi.java
+++ b/backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/plugins/PluginsApi.java
@@ -80,8 +80,8 @@ public class PluginsApi {
       throw new IllegalArgumentException("SonarQube Cloud URL must contain a host: " + baseUrl);
     }
     try {
-      var scannerUri = new URI(baseUri.getScheme(), null, "scanner." + host, baseUri.getPort(), null, null, null);
-      return scannerUri + "/plugins/" + urlEncode(key) + "/versions/" + urlEncode(hash) + ".jar";
+      var path = "/plugins/" + key + "/versions/" + hash + ".jar";
+      return new URI(baseUri.getScheme(), null, "scanner." + host, baseUri.getPort(), path, null, null).toASCIIString();
     } catch (URISyntaxException e) {
       throw new IllegalArgumentException("Unable to build the SonarQube Cloud plugin download URL", e);
     }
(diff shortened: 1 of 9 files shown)


backend/server-api/src/main/java/org/sonarsource/sonarlint/core/serverapi/plugins/PluginsApi.java
```diff
@@ -63,8 +69,22 @@ public void getPlugin(String key, Consumer<InputStream> pluginFileConsumer, Sona
...
+    var host = baseUri.getHost();
+    if (host == null) {
+      throw new IllegalArgumentException("SonarQube Cloud URL must contain a host: " + baseUrl);
+    }
+    try {
+      var path = "/plugins/" + key + "/versions/" + hash + ".jar";
+      return new URI(baseUri.getScheme(), null, "scanner." + host, baseUri.getPort(), path, null, null).toASCIIString();
+    } catch (URISyntaxException e) {
+      throw new IllegalArgumentException("Unable to build the SonarQube Cloud plugin download URL", e);
+    }
```
Krosovok: I don't like building it here on-the-fly when for a similar case of `apiBaseUrl` we handle it on `EndpointParams`/`ServerApiHelper#buildApiEndpointUrl` level. It would make more sense to me to add an additional lazy getter to `EndpointParams` and use it in `ServerApiHelper` to make an explicit call to a separate url.

Because now it's different: for api url it is handled inside `ServerApiHelper` and the caller just decides on type of call they want, and for `plugins.` the caller dictates the entire url.

damien-urruty-sonarsource: I think this is a bit subjective. What is inside `ServerApiHelper` helps (no pun intended) for all kinds of APIs. Here this URL is more specific to plugins, so only one specific API. For the same reason, I wouldn't put it in `EndpointParams`, either, as I think it's too specific. In fact, thinking about it, `*Api` classes are supposed to mirror the SQ API endpoints. This new URL is not even part of the API, so maybe it shouldn't be located in PluginsApi but in the `ServerPluginDownloader` caller that would decide. WDYT?

Krosovok: Yeah, it makes more sense. `PluginsApi` for `/api/` call and something else for `plugins.` call and `ServerPluginDownloader` decides. Or maybe even moving the logic of choosing in this place as well.

damien-urruty-sonarsource: Done in 3feee4d60: `PluginsApi` is back to `/api/plugins/*` only, CDN downloads live in `SonarCloudCdnPlugins`, and `ServerPluginDownloader` chooses which one to use.

damien-urruty-sonarsource: Addressed: CDN URL building/download is now in `SonarCloudCdnPlugins`, and `ServerPluginDownloader` decides between that and `PluginsApi`.

Krosovok (approved):

PR #2225, opened 2026-08-07
SLCORE-2537 Fix race when concurrently creating shared machine id

damien-urruty-sonarsource:
## Summary
* Fix a race in `MachineIdProvider` where concurrent creators could return divergent UUIDs or `null` (CREATE_NEW left an empty file briefly; blank-file delete/rewrite raced; same-JVM file locks overlapped)
* Serialize in-process creation with a static lock, then take an exclusive file lock and write under it so callers converge on one on-disk value
* Harden the concurrent test with a start barrier and stronger assertions


damien-urruty-sonarsource:
Fixed in #2226

PR #2239, opened 2026-09-01
SLCORE-2443 Fix incorrect decoding of percent-encoded characters in file URIs

vnaskos-sonar:
## Summary

- `FileUtils.getFilePathFromUri` tried `Path.of(uri.toURL().getPath())` first, but `URL.getPath()` does not decode percent-encoded sequences. A file in a directory named `c++` (sent as `file:///.../c%2B%2B/main.cpp`) or containing `@` (sent as `%40`) resolved to a literal, non-existent path, so SonarQube for IDE silently skipped the file instead of analyzing it.
- Swap the try/catch order: try `Path.of(uri)` first (correctly decodes percent-encoded sequences, including `%2B`), and only fall back to the old `URL`-based approach for the JDK-8162518 case — a `file:///...` URI containing raw, non-percent-encoded non-ASCII characters, which `Path.of(URI)` rejects with `IllegalArgumentException: Bad escape`.
- Added a short Javadoc on the method documenting its two unchecked-exception exit paths (`IllegalArgumentException` for malformed/non-hierarchical URIs, `FileSystemNotFoundException` for unsupported schemes such as IntelliJ's `temp:` or Eclipse's `rse:`), since none of its four callers handle it the same way.

Fixes USER-2137, SLCORE-2443.

## Test plan

- [x] `FileUtilsTests` (new): covers `%2B%2B` decoding, `%20` decoding, and the JDK-8162518 raw-non-ASCII fallback path.
- [x] `BackendInputFileTests` (updated): added a case reproducing the exact `c%2B%2B` scenario from the linked support reproducer, through the real `ClientFile` → `BackendInputFile` path.
- [x] `mvn -pl backend/commons test` — 154/154 pass.
- [x] `mvn -pl backend/core -am test -Dtest=BackendInputFileTests` — 3/3 pass, including the pre-existing non-ASCII test (confirms the JDK-8162518 fallback still works).
- [x] Manually verified end-to-end in SonarQube for VS Code with a local build (sonarlint-core → sonarlint-language-server → sonarlint-vscode): a C/C++ reproducer with `compile_commands.json` covering `c++/main.cpp` and `user@example.com/main.cpp` directories confirmed the issue reproduces on the released build and is resolved by this fix.

After:

after

Before:

before

### DIFF
--- a/backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/FileUtils.java
+++ b/backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/FileUtils.java
@@ -45,7 +45,6 @@ public class FileUtils {
    */
   public static Path getFilePathFromUri(URI uri) {
     try {
-      // Path.of(URI) correctly decodes percent-encoded sequences (e.g. "%2B" -> "+"), unlike URI.getPath() below
       return Path.of(uri);
     } catch (IllegalArgumentException e) {
       // Works around JDK-8162518: Path.of(URI) throws "Bad escape" for raw non-ASCII characters in the URI.
--- a/backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/FileUtils.java
+++ b/backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/FileUtils.java
@@ -19,7 +19,6 @@
  */
 package org.sonarsource.sonarlint.core.commons.util;
 
-import java.net.MalformedURLException;
 import java.net.URI;
 import java.nio.file.Path;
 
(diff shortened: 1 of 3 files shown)


vnaskos-sonar:
Thanks for the review — both findings were correct, verified empirically and fixed in 8374d2f86:

1. **Bug**: the JDK-8162518 fallback used `uri.toURL().getPath()`, which doesn't decode percent-encoding, so a URI mixing raw non-ASCII with a percent-encoded segment (e.g. `file:///home/user/josé/c%2B%2B/main.cpp`) still resolved to a literal, undecoded path. Switched to `uri.getPath()`, which decodes percent-encoding while leaving raw non-ASCII untouched — confirmed via a direct JDK test and added a parameterized case covering exactly this combination.
2. **Javadoc**: corrected the `@throws IllegalArgumentException` contract — it doesn't actually escape for UNC-authority, query/fragment, or raw-non-ASCII URIs (all handled by the fallback); it only escapes for a URI with no hierarchical path at all (e.g. an opaque URI).

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/FileUtils.java
```diff
@@ -24,12 +24,37 @@
...
+   *                                   {@link Path#of(URI)}, silently drops the authority and any query/fragment.
+   *                                   This fallback works around JDK-8162518, where {@code Path.of(URI)} throws
+   *                                   {@code IllegalArgumentException: Bad escape} for raw non-ASCII characters
+   *                                   in a {@code file:///...} URI.
+   * @throws java.nio.file.FileSystemNotFoundException if {@code uri}'s scheme is not {@code file} and has no
+   *                                                     installed {@link java.nio.file.spi.FileSystemProvider}
+   *                                                     (e.g. the {@code temp} scheme used by IntelliJ or the
+   *                                                     {@code rse} scheme used by Eclipse Remote System Explorer)
+   */
   public static Path getFilePathFromUri(URI uri) {
```
damien-urruty-sonarsource, line 46: I already emitted doubts about this approach when it was implemented. I think the problem remains: we shouldn't try to convert a URI to a file path. The URI could represent some remote FS, and we shouldn't assume the file is present on the local FS. 
I checked all call sites of this method, and they all can access the file relative path that is already provided by the client together with the URI. This is what should be used. I understand it's a bigger change, so we could go with the proposed fix, but this is quite hacky

damien-urruty-sonarsource (approved):
I'm not a big fan of the fix, but if it works a bit better than what we previously had, I'm fine moving forward with it. A bigger change should be made to address the problem fully

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/FileUtils.java
```diff
@@ -24,12 +24,37 @@
...
     try {
-      // In case the path contains non-ASCII characters, Path.of() will fail, we should first try to encode the path via toURL()
-      return Path.of(uri.toURL().getPath());
-    } catch (Exception e) {
+      // Path.of(URI) correctly decodes percent-encoded sequences (e.g. "%2B" -> "+"), unlike URI.getPath() below
```
nquinquenel: I think this comment is stale, it contradicts with what's below

backend/commons/src/main/java/org/sonarsource/sonarlint/core/commons/util/FileUtils.java
nquinquenel: Just for clarification, the previous fix that led to this code fixed file path with non-ASCII characters. Here the fix is for another issue. But you're right that in the end it's the same hacky workaround.

PR #2268, opened 2026-09-14
SONARSEC-8851 Tolerate both legacy and reworked taint messages in SQDogfood ITs

vnaskos-sonar:
## Summary
- `master`'s `QA (SQDogfood)` job has been failing consistently since #2262 because `SONARSEC-8714` reworked taint issue messages in SonarSecurity from static canned text to a dynamic `<vulnerability> via <vector> in <sink>()` format. That rework landed first in the dogfood build, so `SonarQubeDeveloperEditionTests$TaintVulnerabilities`'s hardcoded message assertions broke only on `QA (SQDogfood)`, while `QA (SQLatest)`/`QA (SQLts99)` (still on released analyzers with the old wording) kept passing.
- This was previously attempted in #2263 (`SONARSEC-8851`), but that PR hardcoded the *new* message only, which would have broken `QA (SQLatest)`/`QA (SQLts99)` instead (confirmed by that PR's own CI run) — it was closed unmerged and the Jira ticket cancelled. This PR instead accepts **either** wording, so the assertions hold regardless of which SonarSecurity build the target server ships, including once the rework reaches a release.
- Also hardened the `@AfterEach` cleanup to assert the `api/projects/bulk_delete` call actually succeeded, matching the existing pattern in `SonarCloudTests`, instead of silently swallowing the response. A silently failed delete can leave a stale, differently-worded taint issue on the server that the next run's message-based matching won't recognize as the same issue — the likely cause of `shouldSyncTaintVulnerabilities` also seeing 2 vulnerabilities instead of 1 on `QA (SQDogfood)`.

## Test plan
- [x] `mvn -o -Pits -pl its/tests -am test-compile` passes locally
- [ ] `QA (SQDogfood)` CI job goes green

🤖 Generated with [Claude Code](https://claude.com/claude-code)


vnaskos-sonar:
Closing in favor of a new PR tracked under SLCORE-2578 — this fix belongs to the sonarlint-core/SLCORE backlog rather than SonarSecurity's SONARSEC project.

PR #2272, opened 2026-09-14
SLCORE-2079 Update sonar-plugin-api version to 14.0.0.4498

damien-urruty-sonarsource:
## Summary
- Bump `sonar-plugin-api` from 13.4.2.4284 to 14.0.0.4498
- Honor `SensorContext.newIssueResolution()` so issues matching a resolution (same file, start line in range, listed rule key) are never raised in the IDE
- Add a medium test that raises two issues and skips the one annotated with `sonar-resolve`

----
## Summary by Gitar

- **Plugin Updates:**
    - Updated `sonar-javascript-plugin` to `13.9.0.44793`
- **Issue Retraction:**
    - Added support for retracting streamed issues when issue resolutions arrive late from sensors


backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/container/analysis/sensor/SonarLintSensorStorage.java
```diff
@@ -78,7 +86,7 @@ public void store(Issue issue) {
     var inputComponent = sonarLintIssue.primaryLocation().inputComponent();
 
     var activeRule = activeRules.find(sonarLintIssue.ruleKey());
-    if ((activeRule == null) || noSonar(inputComponent, sonarLintIssue)) {
+    if ((activeRule == null) || noSonar(inputComponent, sonarLintIssue) || isResolved(sonarLintIssue)) {
```
kirill-knize-sonarsource, line 91: There's a question about this method call not always working correctly. So we sometimes will not respect `// sonar-resolve`. Leaving it here as well:

> In SonarLintSensorStorage L89 isResolved() call doesn't always work. It depends on issueResolutions processing that may or may not happen by the moment.
> And you seem to noticed it and introduced this buffering reportIfNotFiltered logic. But then it was removed in following commit.
> Because it breaks our issue streaming? We accumulate issues in a buffer and don't stream? That's why you have got rid of a buffer?
> So we still have this initial bug that sometimes we don't respect // sonar-resolve?

backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/container/analysis/sensor/SonarLintSensorStorage.java
```diff
@@ -120,6 +128,31 @@ private static boolean noSonar(InputComponent inputComponent, Issue issue) {
...
+    }
+    var inputFile = (InputFile) inputComponent;
+    var matched = issueResolutions.stream().anyMatch(resolution -> matches(resolution, inputFile, textRange.start().line(), issue.ruleKey()));
+    if (matched) {
+      LOG.debug("Issue {} on {}:{} ignored because it was resolved by a sensor", issue.ruleKey(), inputFile, textRange.start().line());
+    }
+    return matched;
+  }
+
+  private static boolean matches(IssueResolution resolution, InputFile inputFile, int issueLine, RuleKey ruleKey) {
```
kirill-knize-sonarsource: DefaultSonarLintIssueResolution contains status, but it's not being checked here. Doesn't seem like a problem but highlighting the fact it can be DEFAULT or FALSE_POSITIVE.

On SQS they don't hide FP issues, but mark them as such. We could be more coherent with server here. It's surely outside of the scope of API bump, but we could document this.

kirill-knize-sonarsource (commented):
There are some questions.

backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/container/analysis/sensor/SonarLintSensorStorage.java
damien-urruty-sonarsource: Yes, we don't explicitly check the status because we hide the issue the same way if it's DEFAULT or FALSE_POSITIVE. The change would be much bigger to support FP e.g. in standalone mode, so yes it's out of scope, and I don't think we will implement it in the near future. But I will at least leave a comment in the code

medium-tests/src/test/java/mediumtest/analysis/sensor/IssueResolutionSensor.java
kirill-knize-sonarsource, line 1: It duplicates some logic with `IssueThenResolutionSensor`. Maybe worth merging them in one or move common parts in some helper class. Just a nit.

backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/command/AnalyzeCommand.java
```diff
@@ -75,17 +76,34 @@ public AnalyzeCommand(String moduleKey, UUID analysisId, TriggerType triggerType
...
+  }
+
+  public AnalyzeCommand(String moduleKey, UUID analysisId, TriggerType triggerType, Supplier<AnalysisConfiguration> configurationSupplier, Consumer<Issue> issueListener,
+    @Nullable Trace trace, SonarLintCancelMonitor cancelMonitor, TaskManager taskManager, Consumer<List<ClientInputFile>> analysisStarted, Supplier<Boolean> isReadySupplier,
+    Set<URI> files, Map<String, String> extraProperties, Consumer<Issue> issueRetractListener) {
+    this(moduleKey, analysisId, triggerType, configurationSupplier, issueListener, trace, cancelMonitor, taskManager, analysisStarted, isReadySupplier, files, extraProperties,
+      issueRetractListener, new CompletableFuture<>());
   }
 
   public AnalyzeCommand(String moduleKey, UUID analysisId, TriggerType triggerType, Supplier<AnalysisConfiguration> configurationSupplier, Consumer<Issue> issueListener,
```
kirill-knize-sonarsource: This one became unused.

kirill-knize-sonarsource (changes requested):
Some changes needed.

medium-tests/src/test/java/mediumtest/analysis/sensor/IssueResolutionSensor.java
damien-urruty-sonarsource: The sensor is compiled individually via bytebuddy and assembled into a JAR to be used in the medium test. I am not sure I can properly support resolution of a class hierarchy. I would prefer to leave it as is for now

backend/analysis-engine/src/main/java/org/sonarsource/sonarlint/core/analysis/container/analysis/sensor/SonarLintSensorStorage.java
damien-urruty-sonarsource: Fixed.

I implemented a mechanism for 'retracting' an issue, it means if we streamed an issue that we later learn should be consider resolved, we remove it from the list. In theory there is a possibility that the issue list will flicker, in practice I did not notice it. I could only test with sonar-cpp, other analyzers have a condition to not contribute issue resolution in the IDE. I will open PRs for them to allow it

kirill-knize-sonarsource (approved):

PR #2275, opened 2026-09-16
SLCORE-2584 Remove leftover Xodus data store after H2 migration

damien-urruty-sonarsource:
## Summary
- Remove leftover Xodus stores, bindings, and H2 migration code now that findings and local-only issues persist in H2
- Drop the Xodus Maven dependency and the Kotlin BOM that only existed to pin Xodus’s stdlib
- Replace JetBrains annotations that came in transitively from Xodus with jsr305, and simplify test fixtures accordingly


backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/storage/EntityMapper.java
```diff
@@ -375,6 +376,10 @@ public ServerTaintIssue adaptTaint(ServerFindingsRecord rec) {
       severity, type, textRangeWithHash, ruleDescCtx, cleanCodeAttr, impacts, flows);
   }
 
+  private static UUID idOrRandom(@Nullable UUID id) {
```
kirill-knize-sonarsource: SERVER_FINDINGS.id is still nullable with no backfill migration — idOrRandom mints a new UUID per read for legacy rows instead of persisting it.

kirill-knize-sonarsource (approved):
One comment, else LGTM

backend/server-connection/src/main/java/org/sonarsource/sonarlint/core/serverconnection/storage/EntityMapper.java
damien-urruty-sonarsource: Good point. I fixed it and made a database migration, to remove the null UUIDs, add a not-null constraint on the column in the schema and adjust the code around.

PR #2276, opened 2026-09-16
SLCORE-2583 Add shared MCP configuration inspection and update

nquinquenel:
## Summary
Adds shared MCP configuration inspection and full-document update planning. Clients still own file I/O, backups, and writes.

- `inspectMcpConfiguration` classifies missing, malformed, standalone, CLI-managed (`sonar run mcp`), and unknown documents.
- `planMcpConfigurationUpdate` adds or replaces only the `sonarqube` entry and returns a complete replacement document when that is safe.
- CLI-managed and unknown configs are left unchanged. JSON/JSONC is parsed with Gson; comments and formatting are not preserved.
- Requests fail if the agent does not support standalone JSON MCP; clients should check `getIntegrationState` first.

Depends on #2281.


backend/core/src/main/java/org/sonarsource/sonarlint/core/ai/ide/McpConfigurationService.java
```diff
@@ -0,0 +1,199 @@
...
+  private static McpConfigurationState classify(@Nullable ObjectNode section) {
+    if (section == null || !section.has(SONARQUBE_ENTRY)) {
+      return McpConfigurationState.NOT_CONFIGURED;
+    }
+    var entry = section.get(SONARQUBE_ENTRY);
+    if (isCliManagedEntry(entry)) {
+      return McpConfigurationState.CLI_MANAGED;
+    }
+    return isStandaloneEntry(entry) ? McpConfigurationState.STANDALONE : McpConfigurationState.UNKNOWN;
+  }
```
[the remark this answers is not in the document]
nquinquenel: Leaving `"sonarqube": null` / non-object values as UNKNOWN.

UNKNOWN is the "present but unrecognized — do not overwrite" gate. A string or array under `sonarqube` is still user data; treating every non-object as NOT_CONFIGURED would write over it.

JSON `null` is the only borderline case (`has("sonarqube")` is true, the node is not an object). I'd only treat `isNull()` as missing if we explicitly want that as product behavior — not as a silent follow-up to this finding. Empty `{}` would still be UNKNOWN under the proposed wording anyway.

backend/core/src/main/java/org/sonarsource/sonarlint/core/ai/ide/McpConfigurationService.java
```diff
@@ -0,0 +1,186 @@
...
+    var parsed = parse(content, sectionName);
+    if (parsed.error != null) {
+      return new ClassifiedDocument(McpConfigurationState.MALFORMED, null, parsed.error);
+    }
+    return new ClassifiedDocument(classify(parsed.section), parsed.root, null);
+  }
+
+  private static String updatedDocument(ObjectNode root) {
+    return root.toPrettyString() + "\n";
+  }
```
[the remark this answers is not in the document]
nquinquenel: It's acceptable, that does not break JSON

backend/core/src/main/java/org/sonarsource/sonarlint/core/ai/ide/AiAgentCapabilities.java
```diff
@@ -19,18 +19,25 @@
...
   static AiIntegrationAgentCapability of(AiAgent agent) {
-    return new AiIntegrationAgentCapability(agent, supportsCliIntegration(agent), STANDALONE_MCP_SUPPORTED);
+    return new AiIntegrationAgentCapability(agent, supportsCliIntegration(agent), jsonSectionName(agent).isPresent());
+  }
+
+  static Optional<String> jsonSectionName(AiAgent agent) {
+    return switch (agent) {
+      case GITHUB_COPILOT -> Optional.of("servers");
+      case CURSOR, WINDSURF, KIRO, CLAUDE_CODE -> Optional.of("mcpServers");
+      case CODEX -> Optional.empty();
```
damien-urruty-sonarsource, line 40: We don't support Codex? Why? Could we leave a comment to briefly explain why?

backend/core/src/main/java/org/sonarsource/sonarlint/core/ai/ide/AiAgentCapabilities.java
```diff
@@ -19,18 +19,25 @@
...
-
   private AiAgentCapabilities() {
   }
 
   static AiIntegrationAgentCapability of(AiAgent agent) {
-    return new AiIntegrationAgentCapability(agent, supportsCliIntegration(agent), STANDALONE_MCP_SUPPORTED);
+    return new AiIntegrationAgentCapability(agent, supportsCliIntegration(agent), jsonSectionName(agent).isPresent());
+  }
+
+  static Optional<String> jsonSectionName(AiAgent agent) {
```
damien-urruty-sonarsource, line 35: What if the agent is not relying on Json? Shouldn't we design an API that can work regardless of the format?

backend/core/src/main/java/org/sonarsource/sonarlint/core/ai/ide/McpConfigurationService.java
```diff
@@ -0,0 +1,186 @@
...
+
+  private static String sectionName(AiAgent agent) {
+    return AiAgentCapabilities.jsonSectionName(agent)
+      .orElseThrow(() -> new IllegalArgumentException(agent + " is not supported by standalone MCP configuration"));
+  }
+
+  private static ClassifiedDocument parseAndClassify(String content, String sectionName) {
+    var parsed = parse(content, sectionName);
+    if (parsed.error != null) {
+      return new ClassifiedDocument(McpConfigurationState.MALFORMED, null, parsed.error);
```
damien-urruty-sonarsource: "Classified Document" sounds a bit like FBI or CIA 😂 
Maybe `ParsedMcpDocument` or `ParsedMcpSettingsFile`

backend/core/src/main/java/org/sonarsource/sonarlint/core/ai/ide/McpConfigurationService.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+    var classified = parseAndClassify(content, sectionName);
+    if (classified.error != null) {
+      return new McpConfigurationInspectionResponse(McpConfigurationState.MALFORMED,
+        List.of(classified.error));
+    }
+    return new McpConfigurationInspectionResponse(classified.state, diagnosticsFor(classified.state));
+  }
+
+  public McpConfigurationUpdatePlanResponse planUpdate(McpConfigurationUpdateParams params) {
```
damien-urruty-sonarsource, line 59: Instead of planning, couldn't we do the update directly from the backend?

damien-urruty-sonarsource (approved):
Looks mostly good, I left some feedback. One thing that comes to my mind when reading this PR is that it looks very similar to what we do from the CLI via `sonar integrate`. I am a bit afraid we might end up implementing the same thing in different places. But `sonar integrate` does a lot of things, not only setting up the MCP. One idea, at some point we might want to introduce new specialized commands, e.g. `sonar agent claude mcp install` (quite verbose but explicit), and we could invoke the CLI from the IDE. Just an idea, maybe not a good one

backend/core/src/main/java/org/sonarsource/sonarlint/core/ai/ide/AiAgentCapabilities.java
nquinquenel: Codex has a TOML configuration. I wanted to keep this simple and re-use the Json mapper only. MCP can still come with CLI integration, or could be a follow-up improvement. I'll add a comment, I also created a follow-up ticket: https://sonarsource.atlassian.net/browse/SLCORE-2593

backend/core/src/main/java/org/sonarsource/sonarlint/core/ai/ide/AiAgentCapabilities.java
nquinquenel: As mentioned above, I wanted to keep it simple with JSON/JSONC format only for now.

backend/core/src/main/java/org/sonarsource/sonarlint/core/ai/ide/McpConfigurationService.java
nquinquenel: 😅 
I think `McpDocument` is enough (there's already a `Parsed` record), I'll rename it

backend/core/src/main/java/org/sonarsource/sonarlint/core/ai/ide/McpConfigurationService.java
nquinquenel: I think it's better for client to handle this, client has to show the diff and get confirmation. I wanted to keep IDE filesystem access out of SLCORE 🤷🏻

nquinquenel:
> sonar agent claude mcp install (quite verbose but explicit), and we could invoke the CLI from the IDE. Just an idea, maybe not a good one

Yes that's a good idea! I had in mind similar improvement for the CLI that would simplify the implementation in SLCORE, however, one drawback I see for MCP is that it would require to have the CLI installed - here the user still has the choice only to setup MCP

PR #2281, opened 2026-09-17
SLCORE-2588 Prepare shared CLI setup commands

nquinquenel:
## Summary
Prepares SonarQube CLI setup commands for IDE clients to run in a native terminal, and locks compatibility with the production CLI.

- `prepareInstallCommand` returns the OS-specific installer (`curl | bash` / `irm | iex`).
- `prepareAuthenticateCommand` prepares `auth login` with optional `--server` / `--org` and never includes credentials.
- `prepareIntegrateCommand` prepares `sonar integrate <agent> --global` for `cursor`, `claude`, and `codex`, and rejects unsupported agents.
- Install, authenticate, and integrate fail closed when no usable CLI is installed; clients should call `getIntegrationState` first.

Depends on #2280.


.github/workflows/cli-compatibility.yml
```diff
@@ -0,0 +1,39 @@
...
+    name: Latest production CLI
+    runs-on: sonar-xs
+    timeout-minutes: 15
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      - uses: jdx/mise-action@e6a8b3978addb5a52f2b4cd9d91eafa7f0ab959d # v4.2.0
+        with:
```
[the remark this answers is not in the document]
nquinquenel: A master push can cancel an in-flight scheduled run (same static group + `cancel-in-progress: true`), but that push also triggers this workflow, so the compatibility job still runs on the new SHA. Cancelled runs correctly skip Slack (`conclusion == 'failure'` only).

Splitting the group by `github.event_name` would let schedule and push overlap, which is extra load without covering a skipped night. I’d keep the current group unless we want parallel schedule+push on purpose.

backend/core/src/test/java/org/sonarsource/sonarlint/core/ai/ide/AiIntegrationServiceTests.java
```diff
@@ -236,6 +239,89 @@ void should_find_cli_using_mac_os_path_helper() throws IOException {
...
+    var command = service.prepareCliCommand(new PrepareCliCommandParams(
+      CliCommandAction.AUTHENTICATE, null, " https://sonar.example ", " acme "));
+
+    assertThat(command.getExecutable()).isEqualTo(executable.toString());
+    assertThat(command.getArguments()).containsExactly("auth", "login", "--server", "https://sonar.example", "--org", "acme");
+    assertThat(command.getArguments()).noneMatch(argument -> argument.toLowerCase().contains("token"));
+    assertThat(command.isInteractive()).isTrue();
+  }
+
+  @Test
```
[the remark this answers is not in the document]
nquinquenel: The null-agent path is already covered by `should_reject_integrate_when_agent_is_missing` after the RPC split (`prepareIntegrateCommand` / `PrepareIntegrateCliCommandParams`). The suggested snippet still uses the removed `prepareCliCommand` API.

What was still missing is `auth login` with blank/absent `--server`/`--org` (the first-run case). That’s now `should_prepare_login_without_connection_details` on `prepareAuthenticateCommand`.

.github/workflows/cli-compatibility.yml
```diff
@@ -0,0 +1,39 @@
+name: SonarQube CLI compatibility
```
damien-urruty-sonarsource, line 1: That's a good thing! But I wonder if we should extend this to all tests. There is not a lot of activity on SLCORE (and in fact many of our other projects) and it could be good to run tests nightly. But I understand it's maybe out of scope of this effort

backend/core/src/main/java/org/sonarsource/sonarlint/core/ai/ide/AiIntegrationService.java
```diff
@@ -44,6 +49,11 @@ public AiIntegrationService() {
       Paths.get(System.getProperty("user.home")), System.getenv(), OsSearchPath.MAC_OS_PATH_HELPER));
   }
 
+  @VisibleForTesting
+  public AiIntegrationService(System2 system2, CommandExecutor commandExecutor, Path userHome, Map<String, String> environment) {
```
damien-urruty-sonarsource, line 53: Related to a change already in place, but I wouldn't rely on `System2` here. It comes from the plugin API, and usually we try to confine its usage only to certain modules. I think it's used only to check the OS, which we do in some places already using `SystemUtils` from Apache. No big deal

its/tests/src/test/java/its/LatestSonarQubeCliCompatibilityTests.java
```diff
@@ -0,0 +1,159 @@
...
+import org.sonarsource.sonarlint.core.rpc.protocol.backend.ai.AiAgent;
+import org.sonarsource.sonarlint.core.rpc.protocol.backend.ai.CliAuthenticationStatus;
+import org.sonarsource.sonarlint.core.rpc.protocol.backend.ai.CliInstallationStatus;
+import org.sonarsource.sonarlint.core.rpc.protocol.backend.ai.GetAiIntegrationStateParams;
+import org.sonarsource.sonarlint.core.rpc.protocol.backend.ai.PrepareIntegrateCliCommandParams;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
+
+@DisabledOnOs(OS.WINDOWS)
```
damien-urruty-sonarsource, line 50: Why? Shouldn't we run tests on all platforms ideally?

its/tests/src/test/java/its/LatestSonarQubeCliCompatibilityTests.java
```diff
@@ -0,0 +1,159 @@
...
+@Tag("CliCompatibility")
+class LatestSonarQubeCliCompatibilityTests {
+  private static final long INSTALL_TIMEOUT_MINUTES = 2L;
+
+  @TempDir
+  private Path tempDir;
+
+  @Test
+  void should_be_compatible_with_the_latest_published_cli() throws Exception {
```
damien-urruty-sonarsource, line 59: This is not really an IT as we do it usually. Our ITs are badly named, they are more E2E tests I guess, because we exercise SLCORE from the outside: the RPC API. Here we are using the class directly, and mocking some things. It's maybe OK, just wanted to point the different strategy

backend/core/src/main/java/org/sonarsource/sonarlint/core/ai/ide/AiIntegrationService.java
```diff
@@ -58,6 +68,27 @@ public GetAiIntegrationStateResponse getIntegrationState(GetAiIntegrationStatePa
     return new GetAiIntegrationStateResponse(cliState, agentCapabilities);
   }
 
+  public PrepareCliCommandResponse prepareInstallCommand() {
+    return CliCommandFactory.prepareInstallCommand(locator.isWindows());
```
damien-urruty-sonarsource, line 72: Nitpick: I am not a big fan of factories like these that have many static methods. I think it would be clearer if:
* the locator returns a lookup result, as today
* the lookup result contains an object that represents the CLI (when found), e.g. `SonarQubeCli`
* this `SonarQubeCli` class has the methods that are today in the factory, like `prepareXXXCommand`

But I reckon this is more of a personal taste

damien-urruty-sonarsource (approved):
LGTM, left some suggestions

its/tests/src/test/java/its/LatestSonarQubeCliCompatibilityTests.java
nquinquenel: This test is the Unix installer so it wouldn't work on Windows, and the runner is Linux as well. We'd need a second job to run Windows tests here

PR #2285, opened 2026-09-18
SLCORE-2587 Discover supported AI agent CLIs for IDE integrations

nquinquenel:
## Summary
Discovers supported AI agent CLIs on the local machine and merges them with IDE-detected agents.

- Opt-in via `GetAiIntegrationStateParams.discoverLocalAgentClis` (default `false`, so existing clients are unchanged).
- Finds Claude Code, Codex, GitHub Copilot CLI, Cursor Agent (`cursor-agent`), and Antigravity (`agy`) by executable, not leftover config folders. PATH, Windows suffixes, `~/.local/bin` (Unix), and macOS `path_helper` are reused.
- Merges IDE and CLI detections with stable order. Each capability reports `detectionSources` (`IDE`, `CLI`, or both). GitHub Copilot in the IDE and GitHub Copilot CLI stay separate targets.
- CLI-detected agents get the matching integrate/MCP/skill capabilities even off their native host. PATH resolution and agent probes are memoized for the backend lifetime.

Depends on #2278.

Stack: #2286 → #2287 → #2280 → #2281 → #2276 → #2278 → #2285

PR #2286, opened 2026-09-18
SLCORE-2581 Extract shared OS PATH search helpers

nquinquenel:
## Summary
Extracts shared OS PATH resolution so later CLI discovery can reuse it.

- Adds `OsSearchPath` for PATH lookup, malformed entries, case-insensitive Windows env vars, and macOS `path_helper`.
- Switches `NodeJsHelper` and `ExecutableLocator` to the shared helper. Behavior is unchanged.

## Test plan
- [ ] `OsSearchPathTests` pass
- [ ] Existing Node.js / hook executable detection still works on Linux, macOS, and Windows


backend/core/src/main/java/org/sonarsource/sonarlint/core/nodejs/OsSearchPath.java
```diff
@@ -0,0 +1,110 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.sonarlint.core.nodejs;
```
Krosovok: Shouldn't it go to a different package since it's a shared helper?

Krosovok (approved):

backend/core/src/main/java/org/sonarsource/sonarlint/core/nodejs/OsSearchPath.java
nquinquenel: You're right, probably better on a dedicated OS folder

